PR #2, opened 2019-06-05
Add buildkite configuration yaml file.

ifoox:
For now just a simple configuration that builds and tests everything on ubuntu1804.


ifoox:
See https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md, basically this sets up the config, I'll also file an issue with the bazel team to setup a pipeline that uses this config (this is a prerequisite to that). Will later also add tests for go build/test (see email thread with philwo).

PR #4, opened 2019-06-06
Add a Windows CI pipeline

ifoox:
We don't have a huge need for it at the moment, but this is to test if it actually currently works, and if so help us keep it working. If it turns out to not currently work I'll abandon this for the moment since it's not critical.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -6,3 +6,8 @@ tasks:
           - "//..."
       test_targets:
           - "//..."
+  windows:
+      build_targets:
+          - "//..."
+      test_targets:
+          - "//..."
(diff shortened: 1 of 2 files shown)


ola-rozenfeld:
Igor, what's the state of this?

ifoox:
This is broken (does not compile) due to dependencies we have that don't support windows (you can see that in the log). I'm fine with just killing this PR for the moment, or we can keep it open and use it later when we fix windows support. Your call.

ola-rozenfeld:
I suggest to delete the PR, and keep the code in your branch to reuse whenever we want to get back to this. It will make it more convenient to see the number of open PRs as the current TODO list.

PR #5, opened 2019-06-06
Setup go build/test of the go test as a CI pipeline.

ifoox:
Runs both build and test separately in case the build is broken
somewhere we don't have test coverage.


ifoox:
Passed on CI after Philipp added Go to the docker container used on buildkite:
https://buildkite.com/bazel/remote-apis-sdks/builds/16#b7156b14-09bc-4258-8e5b-6898a09af633

ola-rozenfeld:
I'm wondering why did you need to manually tell me that CI passed. We should do something so that it's automatically reported on every PR.

ifoox:
It does actually show up, I'm probably just not very familiar with the github workflow so wasn't sure exactly how it shows up.

ola-rozenfeld:
I don't see it -- at least not in the same place Bazel's CI is shown (part of the conversation). Where does it show up for you?

ola-rozenfeld:
Ah, I see! Yes, this is different, in Bazel they add a separate message to the conversation that shows the build status. But this is good enough, I think!

PR #6, opened 2019-06-10
Add a starlark deps function. 

ifoox:
This can be used by repos depending on this to load the necessary dependencies instead of having to copy them over manually.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_gazelle//:def.bzl", "gazelle")
 
 # gazelle:prefix github.com/bazelbuild/remote-apis-sdks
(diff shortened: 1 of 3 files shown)


ifoox:
Good idea. This ended up requiring updating our dependency on Gazelle to a particular commit which adds the -to_macro flag (see instructions in the WORKSPACE file). There's no current release after this commit, so I also pinged Jay to ask if he plans one.

ola-rozenfeld:
Oh, I see build is failing.

ifoox:
As per our offline discussion:
 - in order to use the new deps function in WORKSPACE I needed to update gazelle to a new version
 - this caused an unexpected build failure (related to the semver proto checked in to remote-apis)
I've reverted that for now, and will make a separate repro case so we can fix either gazelle or the remote-apis repo.

PR #11, opened 2019-06-19
Refactoring the digest library to use Go structs instead of protos

ola-rozenfeld:


go/digest/digest_test.go
```diff
@@ -65,25 +61,74 @@ func TestValidateDigests_Errors(t *testing.T) {
...
+	}
+	for _, tc := range testcases {
+		dPb := tc.digest.ToProto()
+		gotDigest, err := NewFromProto(dPb)
+		if err == nil && tc.wantErr {
+			t.Errorf("%s: NewFromProto(%v) = (_, nil), want (_, error)", tc.label, dPb)
+			continue
+		}
+		if err != nil {
+			if !tc.wantErr {
```
ramymedhat: I think its generally advised to have a separate test for error cases to avoid this code here.

go/digest/BUILD.bazel
```diff
@@ -16,7 +16,6 @@ go_test(
     srcs = ["digest_test.go"],
     embed = [":go_default_library"],
     deps = [
-        "@com_github_bazelbuild_remote_apis//build/bazel/remote/execution/v2:go_default_library",
```
ramymedhat: Why did this go away? Doesn't it still depend on RE (repb)?

ola-rozenfeld: Not directly, no.

go/digest/digest_test.go
ola-rozenfeld: Good idea. Done.

PR #13, opened 2019-06-24
Adding some general packages for command executions.

ola-rozenfeld:


go/pkg/command/command.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Only directories match.
+	DirectoryInputKind
+
+	// Only files match.
+	FileInputKind
+)
+
+func (s InputKind) String() string {
+	return [...]string{"UnspecifiedInputKind", "DirectoryInputKind", "FileInputKind"}[s]
```
ramymedhat: Wouldn't a switch be less brittle? I'm not sure what the idiomatic way to do this is here.

go/pkg/command/command.go
```diff
@@ -0,0 +1,273 @@
+// Package command defines common types to be used with remote execution.
```
ramymedhat: Can we not refer to remote execution here? This is for command execution, local or remote, correct?

go/pkg/command/command.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Only files match.
+	FileInputKind
+)
+
+func (s InputKind) String() string {
+	return [...]string{"UnspecifiedInputKind", "DirectoryInputKind", "FileInputKind"}[s]
+}
+
+// InputExclusion represents inputs to be excluded from being uploaded for remote execution.
```
ramymedhat: Same here. Not necessary uploaded, but mounted into docker for local execution for instance.

go/pkg/command/command.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// The input kind to match for exclusion.
+	Kind InputKind
+}
+
+// InputSpec represents all the required inputs to a remote command.
+type InputSpec struct {
+	// Input paths (files or directories) that need to be uploaded to the remote
+	// server for the command execution.
+	// Note: a known limitation of the first version of the proxy is that for any
```
ramymedhat: Same comment about uploading. Also the proxy is irrelevant here right?

go/pkg/command/command.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+	}
+	sha256Arr := sha256.Sum256(buf)
+	return hex.EncodeToString(sha256Arr[:])[:8]
+}
+
+// FillDefaultFieldValues initializes valid default values to inner Command fields.
+// This function should be called on every new Command object before use.
+func (c *Command) FillDefaultFieldValues() {
+	if c == nil {
```
ramymedhat: You don't check for this in other functions?

go/pkg/command/command.go
```diff
@@ -0,0 +1,273 @@
...
+
+// DefaultExecutionOptions returns the recommended ExecutionOptions.
+func DefaultExecutionOptions() *ExecutionOptions {
+	return &ExecutionOptions{
+		AcceptCached:    true,
+		DoNotCache:      false,
+		DownloadOutputs: true,
+	}
+}
+
```
ramymedhat: Shouldn't this be part of the sdk not here?

go/pkg/command/command.go
ola-rozenfeld: I looked it up and apparently this is a common pattern (see e.g. time.go), with some error checking which I inserted.

go/pkg/command/command.go
ola-rozenfeld: Right.

go/pkg/command/command.go
ola-rozenfeld: Ah, derp, that's a copy-paste bug!

go/pkg/command/command.go
ola-rozenfeld: I do now :-)

go/pkg/command/command.go
ola-rozenfeld: Not sure, because these options are also relevant to local execution. Same question about Metadata and Result. I suggest to start by exporting these here, and see if we later want to split them out.

go/pkg/outerr/outerr.go
```diff
@@ -0,0 +1,63 @@
...
+// NewStreamOutErr creates an OutErr from two Writers.
+func NewStreamOutErr(out, err io.Writer) *StreamOutErr {
+	return &StreamOutErr{out, err}
+}
+
+// A constant wrapping the standard stdout/stderr streams.
+var SystemOutErr = NewStreamOutErr(os.Stdout, os.Stderr)
+
+func (s *StreamOutErr) WriteOut(buf []byte) {
+	_, e := s.OutWriter.Write(buf)
```
gkousik: nit: inline in the if block

go/pkg/outerr/outerr.go
```diff
@@ -0,0 +1,63 @@
...
+	if e != nil {
+		log.Errorf("error writing to stdout stream: %v", e)
+	}
+}
+
+func (s *StreamOutErr) WriteErr(buf []byte) {
+	_, e := s.ErrWriter.Write(buf)
```
gkousik: nit: inline into the if block

go/pkg/outerr/outerr_test.go
```diff
@@ -0,0 +1,62 @@
...
+		t.Errorf("expected o.GetStdout() to return hello, got %v", gotOut)
+	}
+	gotErr := o.GetStderr()
+	if !bytes.Equal(gotErr, []byte("world")) {
+		t.Errorf("expected o.GetStderr() to return hello, got %v", gotErr)
+	}
+}
+
+func TestSystemOutErr(t *testing.T) {
+	t.Parallel()
```
gkousik: Is it safe to run this test in parallel though since its redirecting stdout and stderr of this process?

go/pkg/command/command_test.go
```diff
@@ -0,0 +1,303 @@
...
+
+func TestFillDefaultFieldValues_PreserveExisting(t *testing.T) {
+	t.Parallel()
+	ids := &Identifiers {
+		CommandId: "bla",
+		ToolName: "foo",
+		InvocationId: "bar",
+	}
+	inputSpec := &InputSpec{}
+	c := &Command{ InputSpec: inputSpec, Identifiers: ids}
```
gkousik: nit: remove space

go/pkg/command/command.go
```diff
@@ -0,0 +1,292 @@
...
+	// An optional id to use to identify a build spanning multiple invocations.
+	CorrelatedInvocationId string
+
+	// An optional tool name to pass to the remote server for logging.
+	ToolName string
+
+	// An optional tool version to pass to the remote server for logging.
+	ToolVersion string
+
+	// A GUID generated for a particular execution of this command.
```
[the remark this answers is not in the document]
ola-rozenfeld: Done.

go/pkg/command/command.go
```diff
@@ -0,0 +1,292 @@
...
+	// Required: command line elements to execute.
+	Args []string
+
+	// An absolute path to the execution root of the command. All the other paths are
+	// specified relatively to this path.
+	ExecRoot string
+
+	// The working directory, relative to the exec root, for the command to run
+	// in. It must be a directory which exists in the input tree. If it is left
+	// empty, then the action is run in the exec root.
```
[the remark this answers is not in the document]
ola-rozenfeld: Right.

go/pkg/command/command_test.go
ola-rozenfeld: Done.

go/pkg/outerr/outerr.go
ola-rozenfeld: Right!

go/pkg/outerr/outerr.go
ola-rozenfeld: Yep.

go/pkg/outerr/outerr_test.go
ola-rozenfeld: Good point! It is now, because there's only one test case that does it. But if we ever add another test case here, that may no longer be true.

go/pkg/command/command.go
```diff
@@ -0,0 +1,292 @@
...
+	return fmt.Sprintf("%!InputKind(%d)", s)
+}
+
+// InputExclusion represents inputs to be excluded from being considered for command execution.
+type InputExclusion struct {
+	// Required: the path regular expression to match for exclusion.
+	Regex string
+
+	// The input kind to match for exclusion.
+	Kind InputKind
```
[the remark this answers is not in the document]
ola-rozenfeld: Sure, done.

go/pkg/outerr/outerr_test.go
gkousik: I'd suggest removing it since its likely we'll forget this exists when adding a new test case and wonder why test output is flaky :)

ola-rozenfeld: Sure, done.

go/pkg/command/command.go
ramymedhat: sgtm

PR #16, opened 2019-07-02
Parsing inputs from disk into in-memory tree

ola-rozenfeld:


ola-rozenfeld:
Commit with everything: https://github.com/bazelbuild/remote-apis-sdks/commit/50b27cf9bbb44777bfffc679e076b7a8a0300b37

go/client/tree.go
```diff
@@ -21,6 +25,77 @@ type FileTree struct {
...
+			continue
+		}
+		if m, _ := regexp.MatchString(r.Regex, inp); m {
+			return true
+		}
+	}
+	return false
+}
+
+func collectFiles(execRoot string, is *command.InputSpec, path string, fs map[string][]byte) error {
```
ramymedhat: Can you add a comment for this function and shouldIgnore? It'll be useful to quickly understand what they're doing.

go/client/tree.go
```diff
@@ -21,6 +25,77 @@ type FileTree struct {
...
+		return err
+	}
+	var t command.InputType
+	switch mode := fi.Mode(); {
+	case mode.IsDir():
+		t = command.DirectoryInputType
+	case mode.IsRegular():
+		t = command.FileInputType
+	default:
+		return fmt.Errorf("Unsupported input type: %s, %v", absPath, mode)
```
ramymedhat: errors should start with lowercase

go/client/tree.go
ramymedhat: We should check that fs is not null

ramymedhat: Also mention that collectFiles mutates fs

ramymedhat: Perhaps s/collect/load ? Ideally something more descriptive of what the function does.

ramymedhat: You only need exclusions here not the full inputspec, correct?

go/client/tree_test.go
```diff
@@ -112,6 +116,242 @@ func TestBuildTree(t *testing.T) {
...
+	symlinkTarget string
+}
+
+func construct(dir string, ips []*inputPath) error {
+	for _, ip := range ips {
+		path := filepath.Join(dir, ip.path)
+		if ip.isSymlink {
+			target := ip.symlinkTarget
+			if ip.isAbsolute {
+				target = filepath.Join(dir, target)
```
ramymedhat: Where does `target` get used?

go/client/tree_test.go
```diff
@@ -112,6 +116,242 @@ func TestBuildTree(t *testing.T) {
...
+			want: &client.FileTree{
+				Dirs: map[string]*client.FileTree{
+					"a":   &client.FileTree{Files: map[string][]byte{"foo": []byte("foo")}},
+					"b":   &client.FileTree{Files: map[string][]byte{"bla": []byte("bla")}},
+					"txt": &client.FileTree{Files: map[string][]byte{"a": []byte("a")}},
+				},
+			},
+		},
+		{
+			desc: "dir_exclusions",
```
ramymedhat: Can we test directory exclusions for subdirectories?

go/client/tree_test.go
```diff
@@ -112,6 +116,242 @@ func TestBuildTree(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("failed to make temp dir: %v", err)
+			}
+			defer os.RemoveAll(root)
+			if err := construct(root, tc.input); err != nil {
+				t.Fatalf("failed to construct input dir structure: %v", err)
+			}
+			got, err := client.BuildTreeFromInputs(root, tc.spec)
+			if err != nil {
+				t.Fatalf("BuildTreeFromInputs(%s, %v) = _, %v want _, nil", root, tc.spec, err)
```
ramymedhat: Should be Errorf since this is not test setup

go/client/tree.go
ola-rozenfeld: In a private recursive helper function I'd rather not, too many checks. If it were public, sure.

ola-rozenfeld: Right, changed.

ola-rozenfeld: Done.

ola-rozenfeld: Done.

go/client/tree.go
ola-rozenfeld: Right! I'm surprised go lint did not tell me this. The internal one does.

go/client/tree_test.go
ola-rozenfeld: Oh! Lol, it didn't, thank you! It was a bug! Now my test case of absolute symlinks actually tests this path.

go/client/tree_test.go
ola-rozenfeld: Done.

go/client/tree_test.go
ola-rozenfeld: But there's no point testing the next condition if this one is not true, right? I thought Fatalf vs Errorf express this difference, like expectTrue vs assertTrue in Py. http://go/unittestgo#how-do-test-functions-work suggests that, too.

ramymedhat: I'm basing this off of this guideline: http://go/go-test-comments#keep-going
My understanding is that since you're testing BuildTreeFromInputs, you should check each of its return values and not abort when one is not expected. "When comparing several different properties of a function's output, use t.Error for each of those comparisions".

ola-rozenfeld: Ah, okay, I interpreted it as the last instruction -- I don't see result and error as orthogonal properties of the output. But I think most people do it the way you suggest. Changed.

PR #18, opened 2019-07-03
Adding DownloadActionResult function to intermediate layer.

ola-rozenfeld:
Also downgrading some log verbosity.

### DIFF
--- a/go/client/cas.go
+++ b/go/client/cas.go
@@ -46,9 +46,9 @@ func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte)
 	if c.useBatchOps {
 		batches = makeBatches(missing)
 	} else {
-		log.V(2).Info("uploading them individually")
+		log.V(2).Info("Uploading them individually")
 		for i := range missing {
-			log.V(3).Infof("creating single batch of blob %s", missing[i])
+			log.V(3).Infof("Creating single batch of blob %s", missing[i])
 			batches = append(batches, missing[i:i+1])
 		}
 	}
@@ -59,7 +59,7 @@ func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte)
 		eg.Go(func() error {
 			for batch := range todo {
 				if len(batch) > 1 {
-					log.V(3).Infof("uploading batch of %d blobs", len(batch))
+					log.V(3).Infof("Uploading batch of %d blobs", len(batch))
 					bchMap := make(map[digest.Digest][]byte)
 					for _, dg := range batch {
 						bchMap[dg] = blobs[dg]
@@ -68,7 +68,7 @@ func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte)
 						return err
 					}
 				} else {
-					log.V(3).Infof("uploading single blob with digest %s", batch[0])
+					log.V(3).Infof("Uploading single blob with digest %s", batch[0])
 					if _, err := c.WriteBlob(eCtx, blobs[batch[0]]); err != nil {
 						return err
 					}
@@ -222,7 +222,7 @@ func makeBatches(dgs []digest.Digest) [][]digest.Digest {
 			batch = append(batch, dgs[0])
 			dgs = dgs[1:]
 		}
-		log.V(3).Infof("created batch of %d blobs with total size %d", len(batch), sz)
+		log.V(3).Infof("Created batch of %d blobs with total size %d", len(batch), sz)
 		batches = append(batches, batch)
 	}
 	log.V(2).Infof("%d batches created", len(batches))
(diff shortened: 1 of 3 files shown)


go/client/cas.go
```diff
@@ -39,14 +41,14 @@ func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte)
 	if err != nil {
 		return err
 	}
-	log.V(1).Infof("%d blobs to store", len(missing))
+	log.V(2).Infof("%d blobs to store", len(missing))
 	var batches [][]digest.Digest
 	if c.useBatchOps {
 		batches = makeBatches(missing)
 	} else {
-		log.V(1).Info("uploading them individually")
```
ramymedhat: Logs should start with a capital letter here and elsewhere

go/client/cas_test.go
```diff
@@ -625,3 +628,158 @@ func TestFlattenActionOutputs(t *testing.T) {
...
+		NoSecurity: true,
+	}, client.UseBatchOps(true))
+	if err != nil {
+		t.Fatalf("Error connecting to server: %v", err)
+	}
+	defer c.Close()
+
+	fooDigest := fake.Put([]byte("foo"))
+	barDigest := fake.Put([]byte("bar"))
+	dirB := &repb.Directory{
```
ramymedhat: nit: dirA first?

go/client/cas_test.go
```diff
@@ -625,3 +628,158 @@ func TestFlattenActionOutputs(t *testing.T) {
...
+			{Name: "b", Digest: bDigest.ToProto()},
+		},
+	}
+	tree := &repb.Tree{
+		Root:     root,
+		Children: []*repb.Directory{dirA, dirB},
+	}
+	treeBlob, err := proto.Marshal(tree)
+	if err != nil {
+		t.Errorf("failed marshalling Tree: %s", err)
```
ramymedhat: This is test setup right? So Fatalf?

go/client/cas_test.go
```diff
@@ -625,3 +628,158 @@ func TestFlattenActionOutputs(t *testing.T) {
...
+	}
+	treeA := &repb.Tree{
+		Root:     dirA,
+		Children: []*repb.Directory{dirB},
+	}
+	treeABlob, err := proto.Marshal(treeA)
+	if err != nil {
+		t.Errorf("failed marshalling Tree: %s", err)
```
ramymedhat: Same?

go/client/cas_test.go
```diff
@@ -625,3 +628,158 @@ func TestFlattenActionOutputs(t *testing.T) {
...
+		},
+	}
+	execRoot, err := ioutil.TempDir("", "DownloadOuts")
+	if err != nil {
+		t.Fatalf("failed to make temp dir: %v", err)
+	}
+	defer os.RemoveAll(execRoot)
+	err = c.DownloadActionOutputs(ctx, ar, execRoot)
+	if err != nil {
+		t.Fatalf("error in DownloadActionOutputs: %s", err)
```
ramymedhat: Errorf since this is what we test?

go/client/cas_test.go
```diff
@@ -625,3 +628,158 @@ func TestFlattenActionOutputs(t *testing.T) {
...
+		},
+	}
+	for _, out := range wantOutputs {
+		path := filepath.Join(execRoot, out.path)
+		fi, err := os.Lstat(path)
+		if err != nil {
+			t.Errorf("expected output %s is missing", path)
+		}
+		if out.symlinkTarget != "" {
+			if fi.Mode()&os.ModeSymlink == 0 {
```
ramymedhat: spaces? Or whatever gofmt is ok with :)

go/client/cas_test.go
ola-rozenfeld: Hmm, I wonder how that even happened...

go/client/cas.go
ola-rozenfeld: Done.

go/client/cas_test.go
ola-rozenfeld: Definitely.

go/client/cas_test.go
ola-rozenfeld: Yep.

go/client/cas_test.go
ola-rozenfeld: Hmm, yes, to be consistent with other things, but this is a good example of why I disagree with the pattern. If you have a 

res1, res2, ..., err := ComplexFunction()

and you do a Errorf instead of a Fatalf on a err != nil check, you will get a huge amount of test errors that are all caused by an actual single error. So only the first one will actually matter, you'd need to find the first one and ignore the rest. I agree that you can test res1...resK individually with Errorf, but the err != nil check seems gating to everything else.

go/client/cas_test.go
ola-rozenfeld: Apparently, go fmt did this. I think it's horrendous, I wrote it with spaces, but that's how it is :-)

go/client/cas_test.go
ola-rozenfeld: Oh, that's how -- note that dirA includes dirB. So it needs to be defined second.

go/client/cas_test.go
ramymedhat: I think the "keep going" motivation is to print all the errors related to this call at once, so that you fix it at once. Note that if you Fatal on err != nil, you wouldn't notice if you have a bug where you return an error and non-nil outputs. With Errorf you verify the entire output of the function (including the error).

ola-rozenfeld:
Ping?

ramymedhat:
> Ping?

Sorry! Approved.

PR #20, opened 2019-07-03
 Adding capability to handle virtual inputs.

ola-rozenfeld:
This doesn't handle the executable bit correctly, but then nothing else in the current client does. It's an upstream bug that I will fix separately, for now the unit tests do not include the executable bit case.

### DIFF
--- a/go/pkg/command/command.go
+++ b/go/pkg/command/command.go
@@ -45,6 +45,13 @@ type InputExclusion struct {
 	Type InputType
 }
 
+// VirtualInput represents an input that does not actually exist as a file on disk, but we want
+// to stage it as a file on disk for the command execution.
+type VirtualInput struct {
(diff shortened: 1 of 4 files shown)


go/pkg/command/command.go
```diff
@@ -45,11 +45,27 @@ type InputExclusion struct {
...
+// to stage it as a file on disk for the command execution.
+type VirtualInput struct {
+	// The path for the input file to be staged at, relative to the ExecRoot.
+	Path string
+
+	// The byte contents of the file to be staged.
+	Contents []byte
+
+	// Whether the file should be staged as executable.
+  IsExecutable bool
```
ramymedhat: indentation seems broken here

ola-rozenfeld: Indeed! Forgot to run lint and go fmt. I should really add it to presubmit...
Done.

PR #22, opened 2019-07-04
Setting CASConcurrency to global accross parallel requests.

ola-rozenfeld:
Also limiting the default to 50 uploaders. This is the first step in the direction of limiting the overall memory (and network throughput) used by simultaneous uploads.

### DIFF
--- a/go/client/client.go
+++ b/go/client/client.go
@@ -61,12 +61,12 @@ type Client struct {
 	// Retrier is the Retrier that is used for RPCs made by this client.
 	//
 	// This field is logically "protected" and is intended for use by extensions of Client.
-	Retrier        *Retrier
-	chunkMaxSize   ChunkMaxSize
-	useBatchOps    UseBatchOps
-	casConcurrency CASConcurrency
-	rpcTimeout     time.Duration
-	creds          credentials.PerRPCCredentials
+	Retrier      *Retrier
+	chunkMaxSize ChunkMaxSize
+	useBatchOps  UseBatchOps
+	casUploaders chan bool
+	rpcTimeout   time.Duration
+	creds        credentials.PerRPCCredentials
 	// Used to close the underlying connection.
 	io.Closer
 }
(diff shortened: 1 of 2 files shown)


ola-rozenfeld:
This way to do it is sub-optimal for a few reasons:
1. Retries are executed on the same uploader thread. This means, if 50 uploads are executing, 100 are waiting, and one failed, that failed one will use the thread to wait for an exponentially increasing period of time, while the 100 continue to wait. However, doing this thing optimally is not easy!
2. It doesn't remember which digests are concurrently being uploaded by other threads.
3. It doesn't remember which digests have already been uploaded.

But all these things have a lower priority than getting the memory limited in the first place.

go/client/client.go
```diff
@@ -61,12 +61,12 @@ type Client struct {
...
-	Retrier        *Retrier
-	chunkMaxSize   ChunkMaxSize
-	useBatchOps    UseBatchOps
-	casConcurrency CASConcurrency
-	rpcTimeout     time.Duration
-	creds          credentials.PerRPCCredentials
+	Retrier      *Retrier
+	chunkMaxSize ChunkMaxSize
+	useBatchOps  UseBatchOps
+	casUploaders chan bool
```
ramymedhat: I think semaphores are a more natural fit here.

ola-rozenfeld: Not according to this: http://jmoiron.net/blog/limiting-concurrency-in-go/ which was the second result on controlling concurrency in Go. The first result (https://blog.golang.org/pipelines) agrees with this pattern as well. In general, the guideline is "you should rarely need to use sync package primitives, use channels instead".

go/client/client.go
```diff
@@ -93,14 +93,13 @@ func (u UseBatchOps) Apply(c *Client) {
...
-// methods can result in N * CASConcurrency requests in flight at once.
+// CASConcurrency is the number of simultaneous requests that will be issued for CAS upload and
+// download operations.
 type CASConcurrency int
 
 // Apply sets the CASConcurrency flag on a client.
 func (cy CASConcurrency) Apply(c *Client) {
-	c.casConcurrency = cy
+	c.casUploaders = make(chan bool, cy)
 }
```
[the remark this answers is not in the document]
ola-rozenfeld: It's a design pattern. Look at the Opt interface which the CasConcurrency type implements, as well as all the other Client options. This way, we can pass variadic Opt to Dial (will be renamed to New at some point).

go/client/client.go
ramymedhat: Discussed offline. sgtm.

PR #26, opened 2019-07-08
A Chunker to support memory-bounded buffered uploads with lazy file

ola-rozenfeld:
reads.

### DIFF
--- a/go/pkg/chunker/chunker.go
+++ b/go/pkg/chunker/chunker.go
@@ -123,15 +123,11 @@ func (c *Chunker) HasNext() bool {
 	return !c.initialized || c.bytesLeft() > 0
 }
 
-// Chunk is a piece of a byte[] blob suitable for being uploaded. The first Chunk returned by a
-// Chunker will contain the Digest of the whole data. In all other cases, the Digest will be nil.
+// Chunk is a piece of a byte[] blob suitable for being uploaded.
 type Chunk struct {
-	Digest *digest.Digest
 	Offset int64
 	Data   []byte
 }
-
-var emptyChunk = &Chunk{Digest: &digest.Empty}
 
 // Next returns the next chunk of data or error. ErrEOF is returned if and only if HasNext is false.
 // Chunk.Data will be empty if and only if the full underlying data is empty (in which case it will
@@ -142,7 +138,7 @@ func (c *Chunker) Next() (*Chunk, error) {
 	}
 	c.initialized = true
 	if c.digest.Size == 0 {
-		return emptyChunk, nil
+		return &Chunk{}, nil
 	}
 	if c.contents == nil && c.reader == nil { // First read from a file.
 		f, err := os.Open(c.path)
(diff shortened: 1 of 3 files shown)


ola-rozenfeld:
Yes! Ramy, you were totally right, using the bufio.Reader is so much better! I managed to do all the same optimizations with much less code! Thank you, PTAL!

go/pkg/chunker/chunker.go
```diff
@@ -139,65 +137,47 @@ func (c *Chunker) Next() (*Chunk, error) {
...
 		return emptyChunk, nil
 	}
-	toRead := IOBufferSize
-	if int64(toRead) > bytesLeft {
-		toRead = int(bytesLeft)
-	}
-	if c.contents != nil {
-		c.buffer = c.contents
-	} else if c.reader == nil {
+	if c.contents == nil && c.reader == nil {
```
ramymedhat: Comment that this handles the case where this is the first call to Next() and not after a reset?

go/pkg/chunker/chunker.go
```diff
@@ -23,13 +24,13 @@ var ErrEOF = errors.New("ErrEOF")
...
-	chunkSize        int
-	reader           io.ReadCloser
-	contents, buffer []byte
-	digest           digest.Digest
-	offset           int64
-	initialized      bool
-	path             string
+	chunkSize   int
+	reader      *bufio.Reader
+	contents    []byte
```
ramymedhat: nit: comment that contents caches small files that are read from buffered io in one chunk?

go/pkg/chunker/chunker.go
```diff
@@ -139,65 +137,47 @@ func (c *Chunker) Next() (*Chunk, error) {
...
 	}
-	start := int(c.offset % int64(len(c.buffer)))
+	bytesLeft := c.bytesLeft()
 	bytesToSend := c.chunkSize
 	if bytesLeft < int64(bytesToSend) {
 		bytesToSend = int(bytesLeft)
 	}
 	var data []byte
-	leftover := bytesToSend
 	if c.reader != nil {
```
ramymedhat: I would structure this as follows to avoid the nested branching logic:
buffOffset := c.offset
if c.reader != nil {
  buffOffset = 0
  newBuffSize := bytesToSend
  if c.offset == 0 && c.digest.Size <= int64(IOBufferSize) {
   newBuffSize = c.digest.Size
  }
  c.contents = make([]byte, newBuffSize)
  n, err := io.ReadFull(c.reader, c.contents)
  ... // error handling
}
data = c.contents[buffOffset:buffOffset + bytesToSend]
...

ola-rozenfeld: That doesn't quite work -- I can't set c.contents unless it's the fully cached data. And you set it unconditionally here whenever reader is nil. So the next read would fail here.

ola-rozenfeld:
Works now, PTAL!

PR #30, opened 2019-07-11
Fix the fake to compare the received action digest.

ola-rozenfeld:


go/pkg/fakes/server.go
```diff
@@ -74,6 +74,7 @@ func (s *Server) NewTestClient(ctx context.Context) (*rc.Client, error) {
 // TestEnv is a wrapper for convenient integration tests of remote execution.
 type TestEnv struct {
 	Client     *rexec.Client
+	GrpcClient *rc.Client
```
ramymedhat: Why do we need the grpc client here? Can't we access it from the rexec.Client?

ola-rozenfeld: Oh, right! I forgot it was public there! Sure, I can use it from there. I thought to expose it for the intermediate layer tests, which don't care about rexec.

PR #31, opened 2019-07-12
Use

ola-rozenfeld:
Also adding max concurrency computation in the CAS fake and checking it is as expected.

### DIFF
--- a/go/client/cas.go
+++ b/go/client/cas.go
@@ -11,7 +11,7 @@ import (
 	"sync"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/digest"
-	log "github.com/golang/glog"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/chunker"
 	"github.com/golang/protobuf/proto"
 	"github.com/pborman/uuid"
 	"golang.org/x/sync/errgroup"
@@ -19,12 +19,12 @@ import (
 	"google.golang.org/grpc/status"
 
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	log "github.com/golang/glog"
 )
 
-// WriteBlobs stores a large number of blobs from a digest-to-blob map. It's intended for use on the
-// result of PackageTree. Unlike with the single-item functions, it first queries the CAS to
-// see which blobs are missing and only uploads those that are.
-func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte) error {
+// Upload stores a number of uploadable items.
+// It first queries the CAS to see which items are missing and only uploads those that are.
+func (c *Client) Upload(ctx context.Context, data ...*chunker.Chunker) error {
 	if cap(c.casUploaders) <= 0 {
 		return fmt.Errorf("CASConcurrency should be at least 1")
 	}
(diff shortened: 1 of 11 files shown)


ola-rozenfeld:
Hmm. I had a commit description but git eated it :-)
This uses Chunkers in uploads. I still keep the original API of using byte[] and map[Digest][]byte for now, mostly because I don't want to break existing users.

go/client/cas.go
```diff
@@ -11,20 +11,20 @@ import (
...
+	log "github.com/golang/glog"
 )
 
-// WriteBlobs stores a large number of blobs from a digest-to-blob map. It's intended for use on the
-// result of PackageTree. Unlike with the single-item functions, it first queries the CAS to
-// see which blobs are missing and only uploads those that are.
-func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte) error {
+// Upload stores a number of uploadable items.
+// It first queries the CAS to see which items are missing and only uploads those that are.
+func (c *Client) Upload(ctx context.Context, data ...*chunker.Chunker) error {
```
ramymedhat: suggestion: UploadIfMissing?

go/client/cas_test.go
```diff
@@ -390,34 +376,42 @@ func TestWriteBlobs(t *testing.T) {
...
 					for _, blob := range tc.input {
-						input[digest.NewFromBlob(blob)] = blob
+						input = append(input, chunker.NewFromBlob(blob, 20))
 					}
 
-					err := c.WriteBlobs(ctx, input)
+					err := c.Upload(ctx, input...)
 					if err != nil {
-						t.Fatalf("c.WriteBlobs(ctx, inputs) gave error %s, expected nil", err)
+						t.Fatalf("c.Upload(ctx, input) gave error %v, expected nil", err)
```
ramymedhat: Errorf

go/client/cas_test.go
```diff
@@ -390,34 +376,42 @@ func TestWriteBlobs(t *testing.T) {
...
 
-					for _, blob := range tc.present {
-						dg := digest.NewFromBlob(blob)
-						if fake.BlobWrites(dg) > 0 {
-							t.Errorf("blob %v with digest %s was uploaded even though it was already present in the CAS", blob, dg)
+					for _, ch := range input {
+						dg := ch.Digest()
+						blob, err := ch.FullData()
+						if err != nil {
+							t.Fatalf("ch.FullData() returned an error: %v", err)
```
ramymedhat: Errorf

go/client/cas_test.go
```diff
@@ -390,34 +376,42 @@ func TestWriteBlobs(t *testing.T) {
...
+						}
+						if present[dg] {
+							if fake.BlobWrites(dg) > 0 {
+								t.Errorf("blob %v with digest %s was uploaded even though it was already present in the CAS", blob, dg)
+							}
+							continue
```
ramymedhat: How do you make sure the blobs where actually uploaded and not only that the chunkers pulled all the data?

go/client/cas.go
ola-rozenfeld: Sure.

go/client/cas_test.go
ola-rozenfeld: Sorry, I keep forgetting :-)

go/client/cas_test.go
ola-rozenfeld: fake.BlobWrites returns how many times Write or Batch upload was called on the digest. I don't actually know that the chunkers pulled all the data, but I verify in the next lines that the data is correct, so the conclusion is that it was uploaded.

ramymedhat: sgtm

PR #46, opened 2019-07-15
Adding file metadata to be used by rexec and caches

ola-rozenfeld:


go/pkg/filemetadata/filemetadata.go
```diff
@@ -0,0 +1,41 @@
...
+		IsExecutable: (file.Mode() & 0100) != 0,
+		MTime:        file.ModTime(),
+	}, nil
+}
+
+// NoopFileMetadataCache is a non-caching metadata cache (always computes).
+type NoopFileMetadataCache struct{}
+
+// Get computes the metadata from the file contents.
+func (c *NoopFileMetadataCache) Get(path string) (*Metadata, error) {
```
ramymedhat: The cache currently returns a digest, not a pointer to the digest. So make this return Metadata?

ola-rozenfeld: That's actually intentional -- remember we said that we always pass all structs by pointer, with the sole exception of Digest? I want to follow that, just for simplicity.

PR #47, opened 2019-07-15
Using file metadata instead of file digest cache everywhere

ola-rozenfeld:


go/pkg/rexec/rexec.go
```diff
@@ -19,23 +20,15 @@ import (
...
-
-// NoopFileDigestCache is a non-caching cache (always returns a cache miss).
-type NoopFileDigestCache struct{}
-
-// Get computes the digest from the file contents.
-func (c *NoopFileDigestCache) Get(path string) (digest.Digest, error) {
-	return digest.NewFromFile(path)
+// FileMetadataCache is a cache for file contents->Metadata.
+type FileMetadataCache interface {
+	Get(string) (*filemetadata.Metadata, error)
```
ramymedhat: Does golint not complain about Get not being documented?

ola-rozenfeld: Apparently not! :-/

Issue #32, opened 2019-07-15
Lazy file reads in uploader


ola-rozenfeld:
Only read from the file when the chunk needs to be uploaded. Read by chunks to upload, too -- it's okay to have a bigger in-memory buffer, if that proves more efficient than repeated IO, but what's not okay is reading the entire file contents into memory before the upload even starts (unless it's for a small file). Current situation is even worse: ALL the inputs are read into memory first thing, even before we start querying GetMissingBlobs.

ola-rozenfeld:
PRs #26, #29, and #31 are for this.

PR #48, opened 2019-07-15
Building Merkle tree into Chunkers

ola-rozenfeld:
Forking client/tree into a separate package. For the time being, keeping the old code as well, will remove it once all existing users are migrated.
Coverting rexec to use the new package along with the new UploadIfMissing, completing the fix of #32.
Tested manually on a command with 400000 input files.

### DIFF
--- a/go/pkg/tree/tree_test.go
+++ b/go/pkg/tree/tree_test.go
@@ -382,7 +382,7 @@ func TestComputeMerkleTree(t *testing.T) {
 						t.Errorf("  Diff between unpacked roots (want -> got):\n%s", diff)
 					}
 				} else {
-					t.Logf("  Root digest gotten not present in blobs map")
+					t.Errorf("  Root digest gotten not present in blobs map")
 				}
 			}
 			if diff := cmp.Diff(wantBlobs, gotBlobs); diff != "" {
@@ -434,8 +434,7 @@ func TestComputeMerkleTreeErrors(t *testing.T) {
 			t.Fatalf("failed to construct input dir structure: %v", err)
 		}
 		t.Run(tc.desc, func(t *testing.T) {
-			_, _, err := ComputeMerkleTree(root, tc.spec, chunker.DefaultChunkSize)
-			if err == nil {
+			if _, _, err := ComputeMerkleTree(root, tc.spec, chunker.DefaultChunkSize); err == nil {
 				t.Errorf("ComputeMerkleTree(%v) succeeded, want error", tc.spec)
 			}
 		})
--- a/go/pkg/tree/tree_test.go
+++ b/go/pkg/tree/tree_test.go
@@ -1,0 +1,522 @@
+package tree
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/bazelbuild/remote-apis-sdks/go/digest"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/chunker"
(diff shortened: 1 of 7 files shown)


ola-rozenfeld:
A sec, I'll try now to tell git that the tree*.go files were copied, not modified...

ramymedhat:
> A sec, I'll try now to tell git that the tree*.go files were copied, not modified...
> Edit: I give up -- I don't know how to do that, sorry!

I think you need to do a git cp in one commit, then edit files in the next commit

go/pkg/tree/tree_test.go
```diff
@@ -0,0 +1,522 @@
...
+			t.Fatalf("failed to make temp dir: %v", err)
+		}
+		defer os.RemoveAll(root)
+		if err := construct(root, tc.input); err != nil {
+			t.Fatalf("failed to construct input dir structure: %v", err)
+		}
+		t.Run(tc.desc, func(t *testing.T) {
+			_, _, err := ComputeMerkleTree(root, tc.spec, chunker.DefaultChunkSize)
+			if err == nil {
+				t.Errorf("ComputeMerkleTree(%v) succeeded, want error", tc.spec)
```
ramymedhat: nit: if _,_,err:= ...; err == nil {

go/pkg/tree/tree_test.go
```diff
@@ -0,0 +1,522 @@
...
+
+			gotBlobs := make(map[digest.Digest][]byte)
+			gotRootDg, inputs, err := ComputeMerkleTree(root, tc.spec, chunker.DefaultChunkSize)
+			if err != nil {
+				t.Errorf("ComputeMerkleTree(...) = gave error %v, want success", err)
+			}
+			for _, ch := range inputs {
+				blob, err := ch.FullData()
+				if err != nil {
+					t.Errorf("chunker %v FullData() returned error %v", ch, err)
```
ramymedhat: What does this print?

go/pkg/tree/tree_test.go
```diff
@@ -0,0 +1,522 @@
...
+				if gotRootBlob, ok := gotBlobs[gotRootDg]; ok {
+					gotRoot := new(repb.Directory)
+					if err := proto.Unmarshal(gotRootBlob, gotRoot); err != nil {
+						t.Errorf("  When unpacking root blob, got error: %s", err)
+					} else {
+						diff := cmp.Diff(tc.rootDir, gotRoot)
+						t.Errorf("  Diff between unpacked roots (want -> got):\n%s", diff)
+					}
+				} else {
+					t.Logf("  Root digest gotten not present in blobs map")
```
ramymedhat: Why is this not an error?

go/pkg/tree/tree_test.go
ola-rozenfeld: The chunker has a String() method. This prints something like:
chunker <3 bytes>: /tmp/Directory exclusions884289053/barDir/bar FullData() returned error <...>

go/pkg/tree/tree_test.go
ola-rozenfeld: Huh. You're right, it totally should be an error!

go/pkg/tree/tree_test.go
ola-rozenfeld: Done.

go/pkg/tree/tree_test.go
ramymedhat: Thanks for clarifying!

PR #56, opened 2019-07-17
Handling non-normalized paths in tree construction. Fixes #55.

### LINKED ISSUE #55: Tree construction should be able to handle non-normalized paths

ola-rozenfeld:


go/pkg/tree/tree.go
```diff
@@ -108,11 +108,9 @@ func ComputeMerkleTree(execRoot string, is *command.InputSpec, chunkSize int, ca
 }
 
 func buildTree(files map[string]*chunker.Chunker) *treeNode {
-	// This is not the fastest way to build a Merkle tree, but it should do for the intended uses of
-	// this library.
 	root := &treeNode{}
 	for name, ch := range files {
-		segs := strings.Split(name, "/")
+		segs := strings.Split(filepath.Clean(name), string(filepath.Separator))
```
ramymedhat: nit: use filepath.Split

go/pkg/tree/tree.go
```diff
@@ -138,9 +136,6 @@ func buildTree(files map[string]*chunker.Chunker) *treeNode {
 }
 
 func packageTree(t *treeNode, chunkSize int) (root digest.Digest, blobs map[digest.Digest]*chunker.Chunker, err error) {
-	if t == nil {
-		return digest.Empty, nil, errors.New("nil treeNode while packaging tree")
-	}
```
ramymedhat: Why remove this?

go/pkg/tree/tree.go
ola-rozenfeld: I didn't want to, actually, because it splits into only parts and even then it leaves the separator in, which I don't want.

go/pkg/tree/tree.go
ola-rozenfeld: Because it can't happen. Actually, I want to remove a lot of other checks as well. Public functions should error-check for conditions that are parameter or system triggered, and these error checks should be covered  by tests. Checking for invariants that your own private code creates is, imo redundant and should be avoided. Returning an error is part of the public contract. If we violate our own private invariants, we don't need to return an error, we need to crash.
So I want to remove all these checks.

go/pkg/tree/tree.go
ramymedhat: Ah right, you're using all the segs.

go/pkg/tree/tree.go
ramymedhat: I'm not sure what that gets us. This is either covered by tests or not. Panicing doesn't add any extra level of visibility. And the function already returns an error, so I feel like with this rule we have to assess whether the error condition is an internal invariant or not (for instance, is proto.Marshal failing expected or not? In general, "Don't panic" is preferable unless there is absolutely no way of recovery.

go/pkg/tree/tree.go
```diff
@@ -48,7 +48,7 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 // loadFiles reads all files specified by the given InputSpec (descending into subdirectories
 // recursively), and loads their contents into the provided map.
 func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*chunker.Chunker, chunkSize int, cache FileMetadataCache) error {
-	absPath := filepath.Join(execRoot, path)
+	absPath := filepath.Clean(filepath.Join(execRoot, path))
 	meta, err := cache.Get(absPath)
 	t := command.FileInputType
 	if err != nil {
```
ramymedhat: The check below seems brittle. Depending on the FileMetadataCache implementation, the error message could be wrapped/modified. I suggest using IsDir to check if the input is a directory.

ola-rozenfeld: Discussed offline, added the FileError type.

go/pkg/tree/tree.go
ola-rozenfeld: Discussed offline, removed more checks.

ola-rozenfeld:
Added the new error type, PTAL!

Issue #45, opened 2019-07-15
Report execution metadata for performance analysis


ola-rozenfeld:
The things we need to add are, to begin with:

- Start and end times for various interesting events, such as computing the Merkle tree, checking action cache, remote and local execution stages, etc.
- Number of inputs, outputs, total bytes of inputs and outputs
- CAS cache hits for inputs

Whether the action was a Action Cache hit is already obvious from the `ResultStatus`.

ola-rozenfeld:
We can keep adding stuff, but let's file issues on a per-needed-metric basis.

PR #57, opened 2019-07-18
Adding inputs metadata calculation (#45).

ola-rozenfeld:


go/pkg/rexec/rexec_test.go
```diff
@@ -36,8 +36,10 @@ func TestExecCacheHit(t *testing.T) {
 	res, meta := e.Client.Run(context.Background(), cmd, opt, oe)
 
 	wantMeta := &command.Metadata{
-		CommandDigest: cmdDg,
-		ActionDigest:  acDg,
+		CommandDigest:    cmdDg,
```
ramymedhat: Can we also have a test where the input file bytes are not zero?

ola-rozenfeld: Sure, added an input file to one of the tests.

PR #62, opened 2019-07-22
Adding UpdateCachedResult to rexec. Completes #59.

ola-rozenfeld:


ola-rozenfeld:
I wanted to provide an example, but then thought it would be either pretty trivial or not any different from the unit test, so I wasn't sure about it.
I can do it in a separate CL anyway.

PR #64, opened 2019-07-25
Feature/gofmt

### LINKED ISSUE #53: Set up go fmt and lint to run as part of presubmit

ifoox:
Add got pre-commit hook and buildkite CI pre-commit check for go fmt.

Note the CI presubmit hook currently checks all go files under go/ since it didn't seem trivial to figure out the merge base branch that the pull request is against so that we can compute the files changed between the current branch and that branch. We can always add this later.

Part way towards #53

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -4,7 +4,7 @@ tasks:
   ubuntu1804:
       shell_commands:
           - "echo +++ Check go format"
-          - "bash check-gofmt.sh"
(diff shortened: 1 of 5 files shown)


ifoox:
You can see an example of this failing on 7b691f7 where I modified a go file (and made this check actually work) and then succeeding on 6064431 where I reverted the go file change.

ifoox:
Thanks for the review, missed the lint part on first reading, will add that in a later pass.

PR #73, opened 2019-08-14
avoid mac build error because repository_rules.bzl needs a BUILD file

mostynb:
In another project (https://github.com/buchgr/bazel-remote/pull/97)
I noticed a mac build error in buildkite with the previous configuration
claiming that repository_rules.bzl needed a matching BUILD file.
Updating to the tip of master helped- I suspect due to
https://github.com/bazelbuild/remote-apis/pull/83


ifoox:
Thanks for your contribution!

PR #74, opened 2019-08-22
fix some go vet 'unkeyed fields' warnings

mostynb:


ola-rozenfeld:
Not sure if you intended to send this for review, I didn't get an email. Thank you!

ola-rozenfeld:
Just so I understand: does go vet not like it when any structs are initialized without naming the parameters?

mostynb:
> Not sure if you intended to send this for review, I didn't get an email. Thank you!

I just opened a pull-request, not sure if I was supposed to something else to actually request a review?

> Just so I understand: does go vet not like it when any structs are initialized without naming the parameters?

Correct. This is go vet's "composites" check. If the fields in the struct are changed in such a way that the order of the types still match, then the compiler can't warn about unkeyed struct literals that still assume the old version of the struct.

ola-rozenfeld:
> I just opened a pull-request, not sure if I was supposed to something else to actually request a review?
Yes, there is "Reviewers" box on the top right -- when you open a PR, click my name to add me to reviewers. That's the only way it would send me an email, otherwise it doesn't.

Thank you, LGTM! I will follow this convention from now on!

ola-rozenfeld:
Actually, maybe we should add go vet to presubmit?

mostynb:
> Actually, maybe we should add go vet to presubmit?

Sounds good to me.

PR #76, opened 2019-09-16
rexec: Do not try downloading empty streams

edbaunton:
In the case where no stderr or stdout is emitted rexec will attempt to download
a stream that has a zero size.

Add a check to see that the stream is not empty before trying to download it.


ola-rozenfeld:
Not sure if you intended to send it for review, I didn't get the email. Thank you! This looks good -- could you please also fix cas.go, readBlobStreamed and readBlobToFile?

ola-rozenfeld:
Oh, and add some tests. This should be easy to test in rexec_test using the fakes package (not adding the empty outputs/stdout to the fake, but it all should still work).

edbaunton:
Thank you @ola-rozenfeld, I totally forgot about this during the meetup!

PR #77, opened 2019-09-24
Adding event time metrics to remote metadata.

ola-rozenfeld:
Addresses most of #45.


go/pkg/rexec/rexec.go
```diff
@@ -273,6 +289,35 @@ func (ec *Context) ExecuteRemotely() {
...
+		From: timeFromProto(start),
+		To:   timeFromProto(end),
+	}
+}
+
+func setTimingMetadata(cm *command.Metadata, em *repb.ExecutedActionMetadata) {
+	if em == nil {
+		return
+	}
+	setEventTimes(cm, "RBEQueued", em.QueuedTimestamp, em.WorkerStartTimestamp)
```
ramymedhat: Can these be constants?

ola-rozenfeld: Done.

Issue #41, opened 2019-07-15
Implement parallel downloads


ola-rozenfeld:
Performance improvement similar to #38, but much more important.

Issue #38, opened 2019-07-15
Implement batch downloads


ola-rozenfeld:
Batch uploads are already there; this feature will greatly enhance performance for workers/bots using this library to download action inputs. It matters a lot less for build clients, because actions usually do not produce lots of outputs.

PR #79, opened 2019-09-29
Implementing batch/parallel downloads

ola-rozenfeld:
Fixing #38, #41.

Making sure to not read too much into memory at the same time.

### DIFF
--- a/go/client/cas_test.go
+++ b/go/client/cas_test.go
@@ -716,9 +716,8 @@ func TestDownloadActionOutputs(t *testing.T) {
 			if !bytes.Equal(contents, out.contents) {
 				t.Errorf("expected %s to contain %v, got %v", path, out.contents, contents)
 			}
-			if out.isExecutable && fi.Mode()&0100 == 0 {
-				t.Errorf("expected %s to be a executable, got %v", path, fi.Mode())
-			}
+			// TODO(olaola): verify the file is executable, if required.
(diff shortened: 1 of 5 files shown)


go/client/cas_test.go
```diff
@@ -774,6 +714,143 @@ func TestDownloadActionOutputs(t *testing.T) {
...
+			batchReqs: 0,
+		},
+		{
+			name:      "large and small blobs hitting max exactly",
+			sizes:     []int{client.MaxBatchSz - 1, client.MaxBatchSz - 1, client.MaxBatchSz - 1, 1, 1, 1},
+			batchReqs: 3,
+		},
+		{
+			name:      "small batches of big blobs",
+			sizes:     []int{1*mb + 1, 1*mb + 1, 1*mb + 1, 1*mb + 1, 1*mb + 1, 1*mb + 1, 1*mb + 1},
```
ramymedhat: Can you also use MaxBatchSz here?

ola-rozenfeld: Done, PTAL.

ola-rozenfeld:
Wow, this is very interesting -- I was wondering why I don't have tests that my executable files are actually executable. So I added them. The tests pass via Bazel, but fail via go test :-/
I suspect I ran into this before, and couldn't figure it out, that's why I removed the lines. Do you remember anything about that?

PR #88, opened 2019-10-22
Picking an unused port instead of hardcoding it.

### LINKED ISSUE #87: Add library for picking an unused port
Hard-coding the port number in test code is flaky.

ola-rozenfeld:
Fixes #87


ola-rozenfeld:
I shamelessly copied the code from https://github.com/google/devx-tools/blob/master/waterfall/golang/testutils/net.go

PR #93, opened 2019-11-06
Fix generic comparisons on protobuf messages

dsnet:
Generated protobuf messages contain internal data structures
that general purpose comparison functions (e.g., reflect.DeepEqual,
pretty.Compare, etc) do not properly compare. It is already the case
today that these functions may report a difference when two messages
are actually semantically equivalent.

Fix all usages by either calling proto.Equal directly if
the top-level types are themselves proto.Message, or by calling
cmp.Equal with the cmp.Comparer(proto.Equal) option specified.
This option teaches cmp to use proto.Equal anytime it encounters
proto.Message types.

### DIFF
--- a/go/client/batch_retries_test.go
+++ b/go/client/batch_retries_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/client"
 	"github.com/bazelbuild/remote-apis-sdks/go/digest"
 	"github.com/bazelbuild/remote-apis-sdks/go/retry"
+	"github.com/golang/protobuf/proto"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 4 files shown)


ola-rozenfeld:
I think you need to run gazelle to fix your dependencies, then this will work. Thank you!

ola-rozenfeld:
Do a `bazel run //:gazelle` from the repo root directory. Then, `bazel test ...` to make sure everything works. If you don't have bazel installed, you can get it here: https://github.com/bazelbuild/bazel/releases/tag/1.1.0

Thank you for cleaning this up!

dsnet:
Would you be willing to make this change for me? I need to make these trivial changes to a number of projects and I don't have the bandwidth to setup the right developer environment for each one.

ola-rozenfeld:
Sure, will do! Thank you!

PR #94, opened 2019-11-07
Adding output metadata metrics. #45.

ola-rozenfeld:


go/pkg/rexec/rexec.go
```diff
@@ -83,7 +83,32 @@ func (ec *Context) downloadStream(raw []byte, dgPb *repb.Digest, write func([]by
...
+func (ec *Context) setOutputMetadata() {
+	if ec.resPb == nil {
+		return
+	}
+	ec.Metadata.OutputFiles = len(ec.resPb.OutputFiles) + len(ec.resPb.OutputFileSymlinks)
+	ec.Metadata.OutputDirectories = len(ec.resPb.OutputDirectories) + len(ec.resPb.OutputDirectorySymlinks)
+	ec.Metadata.OutputDigests = make(map[string]digest.Digest)
+	for _, file := range ec.resPb.OutputFiles {
+		dg := digest.NewFromProtoUnvalidated(file.Digest)
+		ec.Metadata.OutputDigests[file.Path] = dg
```
ramymedhat: Would it be possible to have output digests of files under directories?

ola-rozenfeld: Yes, but a few technical problems with that, which is why I prefer not to do it at least right away (for the same reason, note I explicitly excluded output directories from TotalOutputBytes).

The issue is: how and when to compute this. The files, I can compute based on `ActionResult` proto alone. If I want to go deeper into the directories, I need to either do it extremely inefficiently, or do it during the actual upload/download. The latter is problematic, because -- what if we don't want to download intermediates? Then we have no stats.
So, TL;DR: it's non-trivial, let's talk offline if it's important to support that case. I didn't think it was important, at least for now, which is why I avoided it.

ramymedhat: SGTM

PR #97, opened 2019-11-08
Don't download empty blobs. Closes #76.

ola-rozenfeld:


ola-rozenfeld:
@edbaunton: decided to do it myself here instead of https://github.com/bazelbuild/remote-apis-sdks/pull/76, will apply to more cases that way.

PR #100, opened 2019-11-14
Fix generic comparisons on protobuf messages

ola-rozenfeld:
Closes #93. Generated protobuf messages contain internal data structures
that general purpose comparison functions (e.g., reflect.DeepEqual,
pretty.Compare, etc) do not properly compare. It is already the case
today that these functions may report a difference when two messages
are actually semantically equivalent.

Fix all usages by either calling proto.Equal directly if
the top-level types are themselves proto.Message, or by calling
cmp.Equal with the cmp.Comparer(proto.Equal) option specified.
This option teaches cmp to use proto.Equal anytime it encounters
proto.Message types.

### DIFF
--- a/go/client/BUILD.bazel
+++ b/go/client/BUILD.bazel
@@ -22,7 +22,6 @@ go_library(
         "@com_github_golang_glog//:go_default_library",
         "@com_github_golang_protobuf//proto:go_default_library",
         "@com_github_golang_protobuf//ptypes:go_default_library_gen",
-        "@com_github_google_go_cmp//cmp:go_default_library",
         "@com_github_pborman_uuid//:go_default_library",
         "@com_github_pkg_errors//:go_default_library",
         "@go_googleapis//google/bytestream:bytestream_go_proto",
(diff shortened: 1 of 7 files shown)


ola-rozenfeld:
@dsnet for CLA. Thank you!

dsnet:
Can you rebase the commit with "joetsai@google.com" as the author e-mail?

ola-rozenfeld:
Yes, that worked, thank you!

ola-rozenfeld:
@dsnet : this flakes our tests (#101), because apparently cmp.SortSlices is ignored when proto.Equal is used. Is this a known issue? Is there an OSS workaround?

dsnet:
Unfortunately not. The package you want is protocmp
<https://godoc.org/google.golang.org/protobuf/testing/protocmp> in the v2
module, but it's not released yet in open source.

We're in a bit of a catch-22 where release of v2 protobufs is contingent on
Google using it universally. However, usage of v2 universally is contingent
on all targets (including third-party projects) to be compatible with v2
generated messages.

The best you can do for the time being is to manually sort the relevant
fields before calling cmp.Equal.

JT

ola-rozenfeld:
I see, yep, that's what I was writing now. Thanks for the prompt response!

Issue #40, opened 2019-07-15
restructure the Go SDK repo directories according to Golang standard


ola-rozenfeld:
The structure will https://github.com/golang-standards/project-layout. I'm adding new packages based on this, but the old ones need refactoring.

Issue #101, opened 2019-11-14
flaky test of batch updates


ola-rozenfeld:
TestBatchUpdateBlobsIndividualRequestRetries flaked out with:

```
--- FAIL: TestBatchUpdateBlobsIndividualRequestRetries (0.44s)
    batch_retries_test.go:189: client.BatchWriteBlobs(ctx, blobs) diff on request at index 0 (want -> got):
          (*remoteexecution.BatchUpdateBlobsRequest)(
        -       s`instance_name:"instance" requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000a" size_bytes:1 > data:"\001" > requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000b" size_bytes:1 > data:"\002" > requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000c" size_bytes:1 > data:"\003" > requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000d" size_bytes:1 > data:"\004" > `,
        +       s`instance_name:"instance" requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000d" size_bytes:1 > data:"\004" > requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000a" size_bytes:1 > data:"\001" > requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000b" size_bytes:1 > data:"\002" > requests:<digest:<hash:"000000000000000000000000000000000000000000000000000000000000000c" size_bytes:1 > data:"\003" > `,
          )
FAIL
```

ola-rozenfeld:
Very likely broken by https://github.com/bazelbuild/remote-apis-sdks/commit/6f589a6c79afded70dae048db472022ed6374198 due to not supporting ignoring order in slice comparisons.

PR #102, opened 2019-11-14
Removing old copy of tree package. #40.

ola-rozenfeld:


ola-rozenfeld:
Failing tests are due to a different bug, #101.

PR #104, opened 2019-11-14
Cleanup: forking top-level directories under pkg, to later delete the…

ola-rozenfeld:
… original ones. #40


ola-rozenfeld:
This doesn't actually change any code, I just did `cp -R` a few times.

ola-rozenfeld:
Btw, if you're wondering whether any of these should be under internal/pkg instead: the only candidate currently is `actas`, but I'm not sure even about that. I could move it to internal, but it seems pretty reusable.

PR #109, opened 2019-11-29
Add a ReadProto function to Client 

peterebden:
Pairs with WriteProto; seems pretty generally useful. It's not hard to roll oneself but nice to provide it here I think.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -538,18 +538,16 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
 		}
 	}
 	for _, dir := range ar.OutputDirectories {
-		if blob, err := c.ReadBlob(ctx, digest.NewFromProtoUnvalidated(dir.TreeDigest)); err == nil {
-			t := &repb.Tree{}
-			if err := proto.Unmarshal(blob, t); err != nil {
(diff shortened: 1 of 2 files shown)


ola-rozenfeld:
Sorry, my Github settings were wrong, I missed all PR notifications :-(
Will review shortly.

peterebden:
No worries, thank you!

go/pkg/client/cas.go
```diff
@@ -395,6 +395,15 @@ func (c *Client) readBlobStreamed(ctx context.Context, hash string, sizeBytes, o
 	return n, nil
 }
 
+// ReadProto reads a blob from the CAS and unmarshals it into the given message.
+func (c *Client) ReadProto(ctx context.Context, d digest.Digest, msg proto.Message) error {
```
ola-rozenfeld: Thanks for adding this! Could you please change line 534 here in FlattenActionOutputs to use this function instead? It will save a few lines.

peterebden: Sure, have done.

I've changed it slightly; looks like before it was silently ignoring errors on reading that blob, which seemed wrong? Let me know if I've misunderstood though...

ola-rozenfeld: No, you're right, that was weird! Thank you!

PR #111, opened 2019-12-02
Multiplex remote capabilities between execution / CAS

peterebden:
If you configure the Client with a different CAS endpoint to the execution endpoint, it only returns capabilities from the latter; if that is execution-only then the caller has no direct means of getting the CacheCapabilities without calling themselves.

I'm unsure if this is the "right" way to handle this issue in this SDK; maybe it should be a different function, and there's an awkward fringe case where the execution server returns CacheCapabilities as well (I think this is still the right thing though since the CAS server is the one that actually gets used).

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -52,7 +52,6 @@ type Client struct {
 	byteStream   bsgrpc.ByteStreamClient
 	cas          regrpc.ContentAddressableStorageClient
 	execution    regrpc.ExecutionClient
-	capabilities regrpc.CapabilitiesClient
 	operations   opgrpc.OperationsClient
 	// Retrier is the Retrier that is used for RPCs made by this client.
 	//
@@ -257,7 +256,6 @@ func NewClient(ctx context.Context, instanceName string, params DialParams, opts
 		byteStream:     bsgrpc.NewByteStreamClient(casConn),
 		cas:            regrpc.NewContentAddressableStorageClient(casConn),
 		execution:      regrpc.NewExecutionClient(conn),
-		capabilities:   regrpc.NewCapabilitiesClient(conn),
 		operations:     opgrpc.NewOperationsClient(conn),
 		rpcTimeout:     time.Minute,
 		Connection:     conn,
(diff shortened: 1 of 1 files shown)


go/pkg/client/client.go
```diff
@@ -522,6 +522,20 @@ func (c *Client) GetCapabilities(ctx context.Context, req *repb.GetCapabilitiesR
 	if err != nil {
 		return nil, err
 	}
+	// If the CAS server is different to the execution server, we need to get its capabilities separately.
```
ola-rozenfeld: Since this function is not yet used anywhere, we still have freedom to decide what would be best. I think the behavior is a bit unintuitive, so it's worth splitting out explicitly:
- one wrapper function, GetBackendCapabilities that accepts a connection as parameter (it's not really that important to persist the regrpc.CapabilitiesClient as instance variable, this will only be called once-per-client anyway)
- another wrapper function, GetCapabilities, that only accepts the context and the instance_name and returns the merged ServerCapabilities.
WDYT?

peterebden: Sorry for the slow response - been a busy week here.
That makes sense to me, I think it would be clearer for a user what's going on. Will update this PR shortly.

go/pkg/client/client.go
```diff
@@ -510,33 +510,42 @@ func (c *Client) WaitExecution(ctx context.Context, req *repb.WaitExecutionReque
...
+	return res, nil
+}
+
+// GetCapabilities returns the capabilities for the targeted servers.
+// If the CAS URL was set differently to the execution server then the CacheCapabilities will
+// be determined from that; ExecutionCapabilities will always come from the main URL.
+func (c *Client) GetCapabilities(ctx context.Context) (res *repb.ServerCapabilities, err error) {
+	caps, err := c.GetBackendCapabilities(ctx, c.Connection, &repb.GetCapabilitiesRequest{
+		InstanceName: c.InstanceName,
+	})
```
ola-rozenfeld: nit: the request is the same, you can save lines by putting it into a variable.

go/pkg/client/client.go
```diff
@@ -510,33 +510,42 @@ func (c *Client) WaitExecution(ctx context.Context, req *repb.WaitExecutionReque
...
-// GetCapabilities wraps the underlying call with specific client options.
-func (c *Client) GetCapabilities(ctx context.Context, req *repb.GetCapabilitiesRequest) (res *repb.ServerCapabilities, err error) {
+// GetBackendCapabilities returns the capabilities for a specific server connection
+// (either the main connection or the CAS connection).
+func (c *Client) GetBackendCapabilities(ctx context.Context, conn *grpc.ClientConn, req *repb.GetCapabilitiesRequest) (res *repb.ServerCapabilities, err error) {
 	opts := c.RPCOpts()
 	err = c.Retrier.Do(ctx, func() (e error) {
 		return c.CallWithTimeout(ctx, func(ctx context.Context) (e error) {
-			res, e = c.capabilities.GetCapabilities(ctx, req, opts...)
+			res, e = regrpc.NewCapabilitiesClient(conn).GetCapabilities(ctx, req, opts...)
```
ola-rozenfeld: Now we no longer need to save it as an instance variable.

peterebden: Ah yes, you mentioned before but I forgot to actually do it 🤦

go/pkg/client/client.go
peterebden: Thanks, done!

PR #115, opened 2020-01-08
Warn on empty instance name instead of failing.

arberx:
### Description
As stated in the [design doc](https://docs.google.com/document/d/1AaGk7fOPByEvpAbqeXIyE8HX_A3_axxNnvroblTZ_6s/edit#heading=h.7ayoyyemcljy) `instance name` should be allowed to be empty.

### Changes
Log a warning instead of failing, when the instance name is empty.

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -268,7 +268,7 @@ func DialRaw(ctx context.Context, params DialParams) (*grpc.ClientConn, error) {
 // functionality.
 func NewClient(ctx context.Context, instanceName string, params DialParams, opts ...Opt) (*Client, error) {
 	if instanceName == "" {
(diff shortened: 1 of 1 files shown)


mostynb:
Maybe this shouldn't even be a warning?

arberx:
> Maybe this shouldn't even be a warning?

Agreed. I wonder the original motivation though...

ola-rozenfeld:
Sorry, my Github settings were wrong, I missed all PR notifications :-(
Will review shortly.

ola-rozenfeld:
Oh, right, sorry about that -- since instance is actually mandatory in RBE, we completely forgot that it can be optional elsewhere :-) You're right, this should be a warning.

Issue #42, opened 2019-07-15
Handle empty directory structure


ola-rozenfeld:
On both inputs and outputs, currently only directories that have a file somewhere underneath will be uploaded/downloaded.

PR #116, opened 2020-01-08
Handling empty directory input structure (#42)

ola-rozenfeld:


ola-rozenfeld:
This also updates VirtualInputs to be able to encode empty directories.

go/pkg/tree/tree.go
```diff
@@ -146,6 +161,13 @@ func buildTree(files map[string]*fileNode) *treeNode {
 			node = child
 		}
 
+		if fn.EmptyDirectoryMarker {
+			if node.Dirs == nil {
```
gkousik: I think you would create every directory in the path in the for-loop above, so this check is redundant?

ola-rozenfeld: > I think you would create every directory in the path in the for-loop above, so this check is redundant?

I was a bit paranoid what if segs is empty. It's not supposed to happen, but still...

PR #119, opened 2020-01-15
Output directories empty structure support. Fixes #42.

### LINKED ISSUE #42: Handle empty directory structure

ola-rozenfeld:


go/pkg/tree/tree.go
```diff
@@ -263,6 +262,14 @@ func flattenTree(root digest.Digest, rootPath string, dirs map[digest.Digest]*re
 			return nil, fmt.Errorf("couldn't find directory %s with digest %s", flatDir.p, flatDir.d)
 		}
 
+		// Check whether this is an empty directory.
+		if len(dir.Files) == 0 && len(dir.Directories) == 0 && len(dir.Symlinks) == 0 {
```
gkousik: You can add the three and do a single comparison check ;)

ola-rozenfeld: Oh, right :-)

PR #127, opened 2020-03-05
presubmit: shell_commands should exit if any command fails

mostynb:
This may not be needed if/when this bazelci PR lands:
https://github.com/bazelbuild/continuous-integration/pull/957

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -2,16 +2,16 @@
 validate_config: 1
 tasks:
   ubuntu1804:
-      shell_commands:
-          - "set -e"
-          - "echo +++ Check go format"
-          - "./check-gofmt.sh `find go -name '*.go'`"
-          - "echo +++ Check go vet"
-          - "go vet ./..."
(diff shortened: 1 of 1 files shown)


mostynb:
> So this is just so that it fails fast and doesn't run the build if format fails, etc? Because the lint check certainly works, I remember failing it a few times.

If by lint you mean "go vet", I think that can fail the build because it's the last line in the shell_commands. The problem is that bazelci joins all the shell_commands lines with a newline in between, then executes them together, and the shell only gives the exit code of the last command.

This PR in the bazel-remote repository demonstrates the issue: https://github.com/buchgr/bazel-remote/pull/191

Note that the first commit should obviously fail, but it didn't: https://github.com/buchgr/bazel-remote/pull/191/commits/1d0a0c524e404206b78b994951ffcf28a3c75045
Then when I added `set -e`, it did fail: https://github.com/buchgr/bazel-remote/pull/191/commits/8955f6045c85f3e3fc1ec8e491e06b5128ee04eb

PR #131, opened 2020-03-19
Remove explicit "set -e" from presubmit.yml

philwo:
This was made the default in https://github.com/bazelbuild/continuous-integration/pull/957.


mostynb:
Thanks (and sorry for the noise in this repo).

ifoox:
Thanks Philipp!

PR #134, opened 2020-03-31
Ignore invalid symlinks when computing Merkle tree

gkousik:
When given an output directory containing an invalid symlink (i.e., a
symlink pointing to a non-existent file), the SDK fails to compute
Merkle tree because of Stat failure. This change makes the SDK ignore
it.

Tested by both using this version of re-client and added unit-tests.

### DIFF
--- a/go/pkg/filemetadata/filemetadata.go
+++ b/go/pkg/filemetadata/filemetadata.go
@@ -19,14 +19,23 @@ type Metadata struct {
 
 // FileError is the error returned by the Compute function.
 type FileError struct {
-	IsDirectory bool
-	IsNotFound  bool
-	Err         error
+	IsDirectory      bool
+	IsNotFound       bool
+	IsInvalidSymlink bool
+	Err              error
 }
 
 // Error returns the error message.
 func (e *FileError) Error() string {
 	return e.Err.Error()
+}
+
+func isSymlink(filename string) (bool, error) {
+	file, err := os.Lstat(filename)
+	if err != nil {
+		return false, err
+	}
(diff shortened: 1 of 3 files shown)


go/pkg/filemetadata/filemetadata.go
```diff
@@ -19,23 +19,38 @@ type Metadata struct {
...
 
+func isSymlink(filename string) (bool, error) {
+	file, err := os.Lstat(filename)
+	if err != nil {
+		return false, err
+	}
+	if file.Mode()&os.ModeSymlink != 0 {
+		return true, nil
+	}
+	return false, nil
```
ola-rozenfeld: nit:
return file.Mode()&os.ModeSymlink != 0, nil

go/pkg/tree/tree.go
```diff
@@ -66,7 +66,11 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
 	meta := cache.Get(absPath)
 	t := command.FileInputType
 	if meta.Err != nil {
-		if e, ok := meta.Err.(*filemetadata.FileError); !ok || !e.IsDirectory {
+		e, ok := meta.Err.(*filemetadata.FileError)
```
ola-rozenfeld: if !ok {
  return meta.Err
}
and from there on ignore ok

gkousik: Done

go/pkg/filemetadata/filemetadata.go
gkousik: Done.

PR #135, opened 2020-04-10
Dont retry Internal error codes in RetryTransient

peterebden:
gRPC describes this as used for serious errors; I don't think it should be retried as transient (FWIW we would not normally retry it internally).


peterebden:
Oops, this has a mess of other commits on it too. Will redo without, sorry!

PR #136, opened 2020-04-10
Don't retry Internal errors

peterebden:
gRPC describes this as used for serious errors; I don't think it should be retried as transient (FWIW we would not normally retry that code).


ola-rozenfeld:
For the RE API, this is a retriable error, and we use this code for any server-side problem in RBE. We need to retry this.

peterebden:
OK - what code do we expect a server to send for an error it's received that it does not think should be retried?

ola-rozenfeld:
Any code not in the current retriable list is non-retriable. So, for
example, the server could send a FAILED_PRECONDITION. Or INVALID_ARGUMENT
(if that's relevant).

peterebden:
OK, thanks; I guess what put me off there is those seem to be for more specific cases. Will look into that instead.

PR #139, opened 2020-06-24
Splitting cache_test into two for Windows

ola-rozenfeld:
This is a patch of Fumitoshi's change.


go/pkg/filemetadata/cache_posix_test.go
```diff
@@ -0,0 +1,34 @@
+// +build !windows
```
gkousik, line 1: Out of curiosity - what does this line do?

gkousik: Ah I think https://golang.org/pkg/go/build/#hdr-Build_Constraints. I wonder if its necessary given the selects, but lets leave it for now since you're probably reapplying the patch from Fumitoshi.

ola-rozenfeld: I think it tells go build to not build this file if the platform is windows. For bazel, the same is achieved via the BUILD file.

PR #140, opened 2020-06-24
Fixing thread sanitizer test failure.

ola-rozenfeld:
The test is fine as is -- the problem is that thread sanitizer thinks that every time a single function reads and writes to the same variable it is a data race. In this case, it is not a data race; but it's better to add critical sections than disable tsan for this test entirely (because in other tests this warning might actually be indicative of a problem).

### DIFF
--- a/go/pkg/client/batch_retries_test.go
+++ b/go/pkg/client/batch_retries_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


go/pkg/client/batch_retries_test.go
```diff
@@ -289,24 +293,32 @@ func (s *sleepyBatchServer) GetTree(req *repb.GetTreeRequest, stream regrpc.Cont
 }
 
 func (s *sleepyBatchServer) BatchReadBlobs(ctx context.Context, req *repb.BatchReadBlobsRequest) (*repb.BatchReadBlobsResponse, error) {
+	defer s.wg.Done()
+	s.mu.Lock()
 	s.readRequests++
 	s.numErrors++
 	if s.numErrors < 4 {
+		s.mu.Unlock()
```
ramymedhat, line 301: Why not defer the Unlock()?

ola-rozenfeld: I want the sleep to be outside of the critical section.

PR #141, opened 2020-06-24
Remove Validating cache

gkousik:
Since we are not using the mtime based validating cache anywhere and
dont have immediate plans of using it due to performance reasons, I'm
removing this code. This would make it a little bit simpler to update
the cache entry of the output files.

Test: Unit tests


ola-rozenfeld:
Please resolve conflicts. Also: why the command_proto change?

go/api/command/BUILD.bazel
```diff
@@ -12,7 +12,7 @@ proto_library(
 go_proto_library(
     name = "command_go_proto",
     importpath = "github.com/bazelbuild/remote-apis-sdks/go/api/command",
-    proto = ":command_proto",
+    proto = ":cmd_proto",
```
ola-rozenfeld, line 15: Why this change?

gkousik: This is something that gazelle did automatically. I'll revert it from this change, but perhaps that's a naming convention that gazelle thinks is correct.

gkousik:
Resolved merge conflicts.

PR #142, opened 2020-06-25
Update cache entries of output files

gkousik:
When we download files as part of action result, we also want to update
the digest of the downlaoded file into the file metadata cache so that:
1. Actions that have inputs as outputs reflect the updated digest of the
output
2. We don't have to recompute the digest of outputs if its already
present in the cache and the outputs are used as inputs to another
action

Test:
1. Updated existing unit-test to verify that digest is set in cache

2. Tested the change by integrating it with reproxy:
Ran this command with/ without this change:
FLAG_platform="container-image=docker://gcr.io/androidbuild-re-dockerimage/android-build-remoteexec-image@sha256:582efb38f0c229ea39952fff9e132ccbe183e14869b39888010dacf56b360d62,jdk-version=10" FLAG_exec_strategy=remote FLAG_server_address="unix:///tmp/reproxy.sock" FLAG_exec_root=`pwd` prebuilts/remoteexecution-client/latest//rewrapper --labels=type=tool --platform="Pool=default,container-image=docker://gcr.io/androidbuild-re-dockerimage/android-build-remoteexec-image@sha256:582efb38f0c229ea39952fff9e132ccbe183e14869b39888010dacf56b360d62" --exec_strategy=remote --inputs=test.txt --output_files=test.txt -- /bin/bash -c "x=\`cat test.txt\`; echo \$x,\$PWD > test.txt"

Without the change, it gets cache-hits when I run the above command more than once and with the change it
re-executes the action everytime.


gkousik:
I addressed the comments about the typos (not sure why they aren't showing up for me) and added unit-test to Store() method.

PR #155, opened 2020-07-17
Upgrading to grpc-go v1.30.0

ola-rozenfeld:


ola-rozenfeld:
Note: this should be merged internally today, otherwise grpc-go Monday's release will break the internal code (because it will remove the old symbols).
+ @dfawley, FYI.

PR #156, opened 2020-07-30
Add UploadIfMissingStats to report the CAS misses

k-ye:
Adding this stats so that we don't have to repeat ourselves on the Swarming side. Is there any concerns that changing the method signature could break the clients?

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -27,8 +27,7 @@ import (
 
 // UploadIfMissingStats reports the CAS hits/misses.
 type UploadIfMissingStats struct {
-	Hits   map[digest.Digest]struct{}
-	Misses map[digest.Digest]struct{}
+	Misses []digest.Digest
 }
 
 // UploadIfMissing stores a number of uploadable items.
@@ -109,16 +108,7 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker) 
 	log.V(2).Info("Done")
 
 	result := &UploadIfMissingStats{
-		Hits:   make(map[digest.Digest]struct{}),
-		Misses: make(map[digest.Digest]struct{}),
-	}
-	for _, dg := range missing {
-		result.Misses[dg] = struct{}{}
-	}
-	for _, dg := range dgs {
-		if _, ok := result.Misses[dg]; !ok {
-			result.Hits[dg] = struct{}{}
-		}
+		Misses: missing,
 	}
 	return result, err
 }
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -25,11 +25,17 @@ import (
 	log "github.com/golang/glog"
 )
 
+// UploadIfMissingStats reports the CAS hits/misses.
+type UploadIfMissingStats struct {
+	Hits   map[digest.Digest]struct{}
+	Misses map[digest.Digest]struct{}
+}
+
 // UploadIfMissing stores a number of uploadable items.
 // It first queries the CAS to see which items are missing and only uploads those that are.
-func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker) error {
+func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker) (*UploadIfMissingStats, error) {
 	if cap(c.casUploaders) <= 0 {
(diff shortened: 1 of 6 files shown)


go/pkg/client/cas.go
```diff
@@ -25,11 +25,17 @@ import (
 	log "github.com/golang/glog"
 )
 
+// UploadIfMissingStats reports the CAS hits/misses.
+type UploadIfMissingStats struct {
+	Hits   map[digest.Digest]struct{}
```
atetubou, line 30: I think it is fine to return only Hits or Misses and let user calculate opposite

go/pkg/client/cas.go
```diff
@@ -101,7 +107,20 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker)
 	log.V(2).Info("Waiting for remaining jobs")
 	err = eg.Wait()
 	log.V(2).Info("Done")
-	return err
+
+	result := &UploadIfMissingStats{
+		Hits:   make(map[digest.Digest]struct{}),
+		Misses: make(map[digest.Digest]struct{}),
+	}
+	for _, dg := range missing {
```
atetubou, line 115: maybe just returning `missing` is fine?

go/pkg/client/cas.go
k-ye: Done.

go/pkg/client/cas.go
k-ye: Ack, a slice will be consistent with https://github.com/bazelbuild/remote-apis-sdks/blob/6f048255943e502b14630709f22412f23a7395c3/go/pkg/client/cas.go#L433. But I still feel like a `map` would be more convenient for the caller side. Anyway, if `map` makes more sense, maybe we should do that all together in another PR.

go/pkg/client/cas.go
```diff
@@ -27,8 +27,7 @@ import (
 
 // UploadIfMissingStats reports the CAS hits/misses.
 type UploadIfMissingStats struct {
-	Hits   map[digest.Digest]struct{}
-	Misses map[digest.Digest]struct{}
+	Misses []digest.Digest
```
atetubou, line 30: no need to use struct now?

k-ye: Hmm, actually I'd like to keep this struct, for the ease of extending it (We've had a few times where the args kept piling up, so eventually we made a struct for them). Maybe rename it to `UploadIfMissingResult`?

atetubou: I see. But I prefer to be lazy for using struct here, because I'm not so sure whether we want to have other field sometime soon.

k-ye: OK, maybe we can wait for foundry-x's opinion?

atetubou: I see.

ramymedhat: I would prefer we not add a struct until we need to. This makes the function signature simpler, similar to MissingBlobs and basic Go functions like Write returning number of bytes written.

go/pkg/rexec/rexec.go
```diff
@@ -261,7 +261,7 @@ func (ec *Context) ExecuteRemotely() {
 	log.V(1).Infof("%s> Checking inputs to upload...", cmdID)
 	// TODO(olaola): compute input cache hit stats.
 	ec.Metadata.EventTimes[command.EventUploadInputs] = &command.TimeInterval{From: time.Now()}
-	err := ec.client.GrpcClient.UploadIfMissing(ec.ctx, ec.inputBlobs...)
+	_, err := ec.client.GrpcClient.UploadIfMissing(ec.ctx, ec.inputBlobs...)
```
ramymedhat, line 264: It would be really nice if you add this stat to the command Metadata here as well :-) We can also file a bug and do that later.

go/pkg/client/cas.go
k-ye: OK, done.

go/pkg/rexec/rexec.go
k-ye: Done. Note that I was a bit lazy and just ignored this field in the test comparison... Otherwise it pulled in the digests of `Action` and the command, which is a bit cumbersome to set up...

k-ye:
Hmm, could you help merge? I don't seem to be authorized to do so...

PR #158, opened 2020-08-04
client: make DownloadFiles public

atetubou:
I'd like to use this function from our client implementation tracked in https://crbug.com/1110569

We will use local disk cache for blobs in our client, but existing functions in this SDK don't allow to integrate our cache implementation easily.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -615,7 +615,7 @@ func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionRe
 			downloads[out.Digest] = out
 		}
 	}
-	if err := c.downloadFiles(ctx, execRoot, downloads); err != nil {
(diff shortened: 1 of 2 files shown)


atetubou:
I added simple test.

go/pkg/client/cas.go
```diff
@@ -668,7 +668,7 @@ func copyFile(execRoot, from, to string) error {
 	return err
 }
 
-func (c *Client) downloadFiles(ctx context.Context, execRoot string, outputs map[digest.Digest]*tree.Output) error {
+func (c *Client) DownloadFiles(ctx context.Context, execRoot string, outputs map[digest.Digest]*tree.Output) error {
```
ramymedhat, line 671: Can you add a comment for this function? golint would've caught this. I'll file a bug to set it up on this repo.

atetubou: done.

PR #160, opened 2020-08-12
Tool to download results of a given action digest

gkousik:
More often we find ourselves wanting to download the results of an
action digest to see what they were. This tool lets us do that. Pass it
an action digest and it would download the results to the given
directory.

Test:
1. Ran against a test RBE stack and it downloaded the RBE action output
2. Unit tests

### DIFF
--- a/go/cmd/remotetool/main.go
+++ b/go/cmd/remotetool/main.go
@@ -1,0 +1,78 @@
+// Main package for the remotetool binary.
+//
+// Here's a sample invocation of this tool to download an action result:
+// bazelisk run //go/cmd/remotetool -- \
+// 	--instance=projects/rbe-android-ci/instances/default_instance \
+// 	--service remotebuildexecution.googleapis.com:443 \
+// 	--alsologtostderr --v 1 \
+// 	--credential_file ~/.config/foundry/keys/dev-foundry.json \
+// 	--digest=52a54724e6b3dff3bc44ef5dceb3aab5892f2fc7e37fce5aa6e16a7a266fbed6/147 \
+// 	--path=`pwd`/tmp
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/tool"
+
+	rflags "github.com/bazelbuild/remote-apis-sdks/go/pkg/flags"
+	log "github.com/golang/glog"
+)
+
+// OpType denotes the type of operation to perform.
+type OpType string
+
+const (
+	downloadActionResult OpType = "download_action_result"
+)
+
+var (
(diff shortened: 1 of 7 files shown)


go/cmd/remotetool/main.go
```diff
@@ -0,0 +1,72 @@
...
+	pathPrefix   = flag.String("path", "/tmp", "Directory to which outputs should be downloaded to. Defaults to /tmp/.")
+)
+
+func validate() {
+	if *actionDigest == "" {
+		log.Exitf("action_digest must be specified.")
+	}
+	if _, err := os.Stat(*pathPrefix); err != nil && os.IsNotExist(err) {
+		log.Exitf("path %v does not exist.", *pathPrefix)
+	}
```
ramymedhat, line 36: shouldn't this validation happen depending on the operation?

go/cmd/remotetool/main.go
```diff
@@ -0,0 +1,72 @@
+// Main package for the remotetool binary.
```
ramymedhat, line 1: Can you add documentation for an example use of the tool?

go/pkg/tool/tool_test.go
```diff
@@ -0,0 +1,43 @@
...
+	cmd := &command.Command{
+		Args:        []string{"tool"},
+		ExecRoot:    e.ExecRoot,
+		InputSpec:   &command.InputSpec{},
+		OutputFiles: []string{"a/b/out"},
+	}
+	opt := command.DefaultExecutionOptions()
+	want := "output"
+	_, acDg := e.Set(cmd, opt, &command.Result{Status: command.CacheHitResultStatus}, &fakes.OutputFile{Path: "a/b/out", Contents: want},
+		fakes.StdOut("stdout"), fakes.StdErrRaw("stderr"))
```
ramymedhat, line 26: should the tool also download the StdOut/StdErr of the command and print them?

go/cmd/remotetool/main.go
gkousik: Done.

go/cmd/remotetool/main.go
gkousik: Actually, I realized path doesn't have to be validated. The SDK creates it if it doesn't exist, so removed it. But generally also agree that this validation should happen depending on the type of operation.

go/pkg/tool/tool_test.go
gkousik: Done.

go/cmd/remotetool/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+// OpType denotes the type of operation to perform.
+type OpType string
+
+const (
+	downloadActionResult OpType = "download_action_result"
+)
+
+var (
+	operation    = flag.String("operation", "download_action_result", "Specifies the operation to perform. Currently only get_action_result is supported.")
```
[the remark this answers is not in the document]
gkousik: Done.

PR #161, opened 2020-08-25
Tool to show inputs/outputs & command of an action

gkousik:
The tool traverses both input and output directories recursively and
shows:
1. Empty directories
2. Files
3. Symlinks

The previous tool based on tools_remote only used to show the top level
output directories and didn't support symlinks.

As part of this change, I also:
1. Added fake GetTree() function (this was also a feature reequest from
the Goma team).
2. Added support for fake inputs in SDK.
3. Rearranged some of the fake server setup code so that options are
applied first before Merkle tree is computed. I'm doing this since fake
inputs need to be created from the options before Merkle tree is
computed.

Test: Unit-tests, variety of inteegration tests against actual action
digests in test stack.

### DIFF
--- a/go/pkg/fakes/server.go
+++ b/go/pkg/fakes/server.go
@@ -2,6 +2,7 @@ package fakes
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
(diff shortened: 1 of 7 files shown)


go/pkg/fakes/server.go
```diff
@@ -200,56 +247,62 @@ type OutputFile struct {
 }
 
 // Apply puts the file in the fake CAS and the given ActionResult.
-func (f *OutputFile) Apply(ac *repb.ActionResult, s *Server) {
+func (f *OutputFile) Apply(ac *repb.ActionResult, s *Server, execRoot string) error {
```
ramymedhat, line 250: Do we need to add execRoot here? I'd rather we don't change the Apply() interface just for InputFile. How about putting the full path in InputFile? Or creating a separate function in the TestEnv for creating input files (similar to the execroot package in reclient)?

gkousik: I considered creating something like what we have in re-client. But then this interface is only used in tests, so I decided its ok to change the interface. execRoot is also a common property that I thought it maybe useful for future things (like input directory) that we might add.

ramymedhat: SGTM

PR #167, opened 2020-09-09
Make max concurrent requests and streams as flags

gkousik:
Also changed the default values of these parameters to reflect our
experiments with them from asia region.
1. We need to be able to do more than 50 concurrent CAS RPCs if we are
running a build at 500 parallelism for example.
2. In long-haul links like from asia-east <-> us-central, fewer stream
RPCs per connection is better than more stream RPCs per stream. Even at
50 stream RPCs per conn, we see a 3x latecy increase.

### DIFF
--- a/go/pkg/balancer/gcp_balancer.go
+++ b/go/pkg/balancer/gcp_balancer.go
@@ -19,6 +19,11 @@ const (
 	healthCheckEnabled = true
 )
 
+// MinConnections is the minimum number of gRPC sub-connections the gRPC balancer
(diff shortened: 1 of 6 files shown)


gkousik:
`TestWriteBlobsBatching` is flaky even without this change.

```
 bazelisk test --runs_per_test=10 //go/pkg/client:go_default_test --test_filter="TestWriteBlobsBatching" --test_output=streamed
```

fails 4/10 times.

ifoox (approved):

PR #173, opened 2020-09-11
Add DownloadDirectory function

atetubou:
I'd like to re-use most of sdks' [DownloadActionOutputs](https://github.com/bazelbuild/remote-apis-sdks/blob/daf8fa09cc33/go/pkg/client/cas.go#L585) function in our client.
But we use digest of Directory instead of Tree, so let me add function to download Directory from digest.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -581,6 +581,29 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
 	return outs, nil
 }
 
+// DownloadDirectory downloads the entire directory of given digest.
+func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoot string, cache filemetadata.Cache) error {
+	dir := &repb.Directory{}
+	if err := c.ReadProto(ctx, d, dir); err != nil {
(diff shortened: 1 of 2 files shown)


go/pkg/client/cas.go
```diff
@@ -581,6 +581,29 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
 	return outs, nil
 }
 
+// DownloadDirectory downloads the entire directory of given digest.
+func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoot string, cache filemetadata.Cache) error {
+	dir := &repb.Directory{}
+	if err := c.ReadProto(ctx, d, dir); err != nil {
+		return err
```
gkousik, line 588: It will useful to have more informative error description here:
```
return fmt.Errorf("digest %v cannot be mapped to a directory proto: %v", d, err)
```

gkousik (approved):

go/pkg/client/cas.go
atetubou: changed.

atetubou:
I changed to not using go1.15's function.

PR #176, opened 2020-09-14
make file mode configuable

atetubou:
In our client, we need to specify more restrict permissions for created directories/files. So let me make file mode created by this sdk configuable.

### DIFF
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -83,7 +83,7 @@ func (c *Client) ReadResourceToFile(ctx context.Context, name, fpath string) (in
 }
 
 func (c *Client) readToFile(ctx context.Context, name string, fpath string) (int64, error) {
-	f, err := os.Create(fpath)
+	f, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, c.RegularMode)
 	if err != nil {
 		return 0, err
 	}
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -638,7 +638,7 @@ func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionRe
 		}
 		src := downloads[out.Digest]
 		if src.IsEmptyDirectory {
-			return fmt.Errorf("unexpected empty directory :%s", src.Path)
+			return fmt.Errorf("unexpected empty directory: %s", src.Path)
 		}
 		if err := copyFile(execRoot, src.Path, out.Path, perm); err != nil {
 			return err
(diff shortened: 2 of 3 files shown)


go/pkg/client/client.go
```diff
@@ -82,6 +89,15 @@ const (
 	// DefaultMaxBatchDigests is a suggested approximate limit based on current RBE implementation.
 	// Above that BatchUpdateBlobs calls start to exceed a typical minute timeout.
 	DefaultMaxBatchDigests = 4000
+
+	// DefaultDirMode is mode used to create directories.
+	DefaultDirMode = 0700
```
gkousik, line 94: It seems like this will change current behaviour from creating dirs / files with 0777 to 0700. This may / may not be problematic for other clients - but at minimum we need to test it before we can sign off on the change.

Wdyt about keeping the existing defaults and rather overriding via params for your use case? That way you can be unblocked and then once we test our other clients, we can change the defaults as well.

atetubou: sounds, good. I changed default.

go/pkg/client/bytestream.go
```diff
@@ -83,7 +83,7 @@ func (c *Client) ReadResourceToFile(ctx context.Context, name, fpath string) (in
 }
 
 func (c *Client) readToFile(ctx context.Context, name string, fpath string) (int64, error) {
-	f, err := os.Create(fpath)
+	f, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, c.RegularMode)
```
atetubou, line 86: Note that this is still changing permission from 0666 to 0644
https://cs.opensource.google/go/go/+/master:src/os/file.go;l=319;drc=6b420169d798c7ebe733487b56ea5c3fa4aab5ce

go/pkg/client/cas.go
```diff
@@ -632,7 +632,11 @@ func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionRe
 		}
 	}
 	for _, out := range copies {
-		if err := copyFile(execRoot, downloads[out.Digest].Path, out.Path); err != nil {
+		perm := c.RegularMode
+		if out.IsExecutable {
+			perm = c.ExecutableMode
+		}
+		if err := copyFile(execRoot, downloads[out.Digest].Path, out.Path, perm); err != nil {
```
gkousik, line 639: Given that we are pass RegularMode now as param to `copyFile()`, can you check here and make sure that `downloads[out.Digest]` is NOT a directory?

go/pkg/client/bytestream.go
gkousik, line 86: Can you also change this to `0666`? That was the previous behaviour with Create() - https://golang.org/pkg/os/#Create

gkousik (changes requested):

go/pkg/client/cas.go
atetubou: done

go/pkg/client/bytestream.go
atetubou: done

gkousik (approved):

atetubou:
Thank you for the review

Issue #89, opened 2019-10-23
Standalone tool for CAS Interaction


edbaunton:
The `rexec` tool is a great one stop shop for submitting and executing jobs synchronously.

We have a use case where we would like to get results directly from the ActionCache/CAS given only an ActionId (without the original input root, so we cannot submit the job again).

Would you be open to an additional command line tool dedicated to CAS interaction? Specifically, it would support upload/download of blobs and trees. Additionally, it would be able to interact with the ActionCache and check/download results.

We would be happy to contribute it!

ola-rozenfeld:
Sorry, just saw this -- note that a command line tool for all of that already exists in https://github.com/bazelbuild/tools_remote/! It is in our plans to refactor that repo to use this one; while I'm sure it is helpful to have the SDK in multiple languages, I'm not so sure it is similarly helpful to have all the command line tools in all languages.
And even if it is helpful to have a separate Go tool mirroring tools_remote, then it's still a question of whether it should live here or in the tools_remote repository.

But yes, if tools_remote is not enough for your current use-case, and you send me a patch to add the tool, I'll be happy to review, and we can discuss then where it should live. WDYT?

edbaunton:
Thanks @ola-rozenfeld! I missed your update and was not aware of `tools_remote`. We have a suite of tools available in C++ for CAS interaction now at [buildbox-tools](https://gitlab.com/BuildGrid/buildbox/buildbox-tools).

ola-rozenfeld:
Note we now have a tool for this here as well -- check out go/cmd/remotetool (allows showing/downloading action results, blobs and directories).

PR #179, opened 2020-09-17
Adding download file and Merkle dir to remote tool.

ola-rozenfeld:


ola-rozenfeld:
This also has a few minor fixes and cleanups: sorting the printed paths, always printing the full digests (including size), and not exposing any internal details in example command.

go/cmd/remotetool/main.go
```diff
@@ -29,35 +36,28 @@ type OpType string
 const (
 	downloadActionResult OpType = "download_action_result"
 	showAction           OpType = "show_action"
+	downloadBlob         OpType = "download_blob"
+	downloadMerkleTree   OpType = "download_merkle_tree"
 )
 
 var (
-	operation  = flag.String("operation", "download_action_result", "Specifies the operation to perform. Currently only download_action_result is supported.")
+	operation  = flag.String("operation", "", "Specifies the operation to perform. Currently only download_action_result is supported.")
```
gkousik, line 44: I think this should be removed and the appropriate options should be added.

go/cmd/remotetool/main.go
```diff
@@ -29,35 +36,28 @@ type OpType string
 const (
 	downloadActionResult OpType = "download_action_result"
 	showAction           OpType = "show_action"
+	downloadBlob         OpType = "download_blob"
+	downloadMerkleTree   OpType = "download_merkle_tree"
```
gkousik, line 40: probably name this download_tree / download_dir ?

go/pkg/tool/tool.go
```diff
@@ -71,6 +77,55 @@ func (c *Client) DownloadActionResult(ctx context.Context, actionDigest, pathPre
...
+	}
+	dg, err := digest.NewFromString(blobDigest)
+	if err != nil {
+		return "", err
+	}
+	log.Infof("Downloading blob of %v to %v.", dg, path)
+	if _, err := c.GrpcClient.ReadBlobToFile(ctx, dg, path); err != nil {
+		return "", err
+	}
+	if outputToStdout {
```
gkousik, line 105: nit:
```
if !outputToStdout {
    return "", nil
}
contents, err := ioutil.ReadFile(path)
if err != nil {
    return "", err
}
return string(contents), nil
```

gkousik (approved):

go/cmd/remotetool/main.go
ola-rozenfeld: Done.

go/cmd/remotetool/main.go
ola-rozenfeld: Done.

go/pkg/tool/tool.go

ola-rozenfeld:
This addresses #89 as well.

PR #181, opened 2020-09-18
Read lock before looping through scStates in regeneratePicker()

gkousik:
Not read locking this results in a race condition in writing / reading
to balancer.scStates map variable.

Test: Ran
bazelisk test --features race //go/pkg/balancer/... --test_output=streamed --test_timeout=10
and it fails without the fix to regeneratePicker().


ola-rozenfeld (approved):

gkousik:
@philwo the build failure seems to be unrelated to this change. Is it due to the buildkite image update?

philwo:
> build failure
We figured this out over Hangouts - it was actually a hard to spot regression in this repo and luckily this time at least not our infra's fault. 😁

gkousik:
Yep, thanks @philwo!

PR #183, opened 2020-09-21
Produce chunks for individual dir protos as well

peterebden:
This is useful for some things - for example we've found Buildbarn's browser to be very helpful, but it requires the individual directory protos in an output to be uploaded in order to browse them, and the client doesn't have access to the Tree any more to construct this themselves.


ola-rozenfeld (commented):
Can you please rebase? Thanks!
Also: do I understand correctly that the purpose is so that someone can use GetTree to fetch any of the output directories by its digest? 
Also, if I understand correctly, the tree_test should now fail, and needs to be fixed to require the additional blobs.

peterebden:
Thanks, you're correct - have updated the test as well.

ola-rozenfeld (approved):

PR #184, opened 2020-09-21
Optimise out uploading the empty blob in a couple of places

peterebden:
Per https://github.com/bazelbuild/remote-apis/pull/131 there is no longer any requirement to upload the empty blob since the server must provide it. This optimises it out in a couple of places to make callers' life a bit easier.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -80,10 +80,11 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker) 
 	var dgs []digest.Digest
 	chunkers := make(map[digest.Digest]*chunker.Chunker)
 	for _, c := range data {
-		dg := c.Digest()
-		if _, ok := chunkers[dg]; !ok {
-			dgs = append(dgs, dg)
-			chunkers[dg] = c
+		if dg := c.Digest(); dg != digest.Empty {
+			if _, ok := chunkers[dg]; !ok {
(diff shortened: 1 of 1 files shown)


ola-rozenfeld (approved):

ola-rozenfeld:
Can you please rebase? Thank you!

peterebden:
Apologies for the delay - have rebased.

go/pkg/client/cas.go
```diff
@@ -597,6 +601,9 @@ func (c *Client) ResourceNameWrite(hash string, sizeBytes int64) string {
 // GetDirectoryTree returns the entire directory tree rooted at the given digest (which must target
 // a Directory stored in the CAS).
 func (c *Client) GetDirectoryTree(ctx context.Context, d *repb.Digest) (result []*repb.Directory, err error) {
+	if d.SizeBytes == 0 && d.Hash == digest.Empty.Hash {
+		return nil, nil
```
ola-rozenfeld, line 605: Wait, is that correct? Shouldn't this return an empty Directory proto?

peterebden: It's returning a slice of directory protos, so nil is pretty equivalent to an empty slice here.

ola-rozenfeld: Yes, but I mean: shouldn't it return a slice containing one item, `[]*repb.Directory{&repb.Directory{}}`, instead? An empty digest maps to an empty directory.

PR #185, opened 2020-09-23
Adding action reexecution. Also fixing a bug in input tree computation.

ola-rozenfeld:


go/pkg/tool/tool.go
```diff
@@ -31,6 +36,102 @@ type Client struct {
...
+		FileMetadataCache: fmc,
+		GrpcClient:        c.GrpcClient,
+	}
+	_, inputPaths, err := c.getInputTree(ctx, actionProto.GetInputRootDigest())
+	if err != nil {
+		return err
+	}
+	if inputRoot == "" {
+		curTime := time.Now().Format(time.RFC3339)
+		inputRoot = filepath.Join(os.TempDir(), acDg.Hash+"_"+curTime)
```
gkousik, line 73: Why not `ioutil.TempDir(os.TempDir(), acDg.Hash + "_" + curTime)` ?

gkousik (approved):

go/pkg/tool/tool.go
ola-rozenfeld: Only because it's one line longer (ioutil.TempDir returns an error), and I don't need to worry about races here (since it's a manual tool).

Issue #146, opened 2020-07-03
Support `SymlinkNode` in `ComputeMerkleTree()`?


k-ye:
IIUC, currently the file tree produced by `ComputeMerkeTree()` https://github.com/bazelbuild/remote-apis-sdks/blob/c00c035a74d44f1dc7d7ce8fe5dbca3b8956ed1a/go/pkg/tree/tree.go#L110 could only contain `FileNode` or `DirectoryNode`, but not `SymlinkNode`. 

This could be very useful for Swarming's isolate support, since the `.isolate` file could contain symlinks. It could even contain invalid symlinks for testing purpose (@atetubou correct me if I'm wrong). As an example, I wrote this small case:

```go
		{
			desc: "File invalid symlink included",
			input: []*inputPath{
				{path: "foo", isSymlink: true, symlinkTarget: "fooDir/foo"},  // invalid
			},
			spec: &command.InputSpec{
				Inputs: []string{"foo"},
			},
			rootDir: &repb.Directory{
				Symlinks: []*repb.SymlinkNode{{Name: "foo", Target: "fooDir/foo"}},
			},
			additionalBlobs: [][]byte{},
                         // ignore the cache stats
		},
```

But I got an empty `repb.Directory{}` instead.

From what I can see, the symlink support needs to be added at several places:

1. `filemeta.Metadata` needs a symlink flag: https://github.com/bazelbuild/remote-apis-sdks/blob/c00c035a74d44f1dc7d7ce8fe5dbca3b8956ed1a/go/pkg/filemetadata/filemetadata.go#L12-L16
2. `tree` implementation needs to distinguish symlinks from actual files: https://github.com/bazelbuild/remote-apis-sdks/blob/c00c035a74d44f1dc7d7ce8fe5dbca3b8956ed1a/go/pkg/tree/tree.go#L24-L33.
3. `VirtualInput` might also need this flag: https://github.com/bazelbuild/remote-apis-sdks/blob/c00c035a74d44f1dc7d7ce8fe5dbca3b8956ed1a/go/pkg/command/command.go#L56-L70

[...]

PR #188, opened 2020-09-29
Handle symlinks in go/pkg/tree

peterebden:
Currently they are silently turned into file nodes of the contents of whatever they point to.

### DIFF
--- a/go/pkg/tree/tree.go
+++ b/go/pkg/tree/tree.go
@@ -22,12 +22,14 @@ import (
 // tree later. It corresponds roughly to a *repb.Directory, but with pointers, not digests, used to
 // refer to other nodes.
 type treeNode struct {
-	Files map[string]*fileNode
-	Dirs  map[string]*treeNode
+	Files    map[string]*fileNode
+	Dirs     map[string]*treeNode
+	Symlinks map[string]string
 }
 
 type fileNode struct {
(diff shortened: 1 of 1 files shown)


ola-rozenfeld:
Relevant discussion in https://github.com/bazelbuild/remote-apis-sdks/issues/146.

ola-rozenfeld:
This will need a bit more work before going in:
- fix tests
- add new test cases
- add configuration to Client (Opt) to configure the symlink handling, with current behaviour as the default. Maybe something like:

```
type DanglingSymlinkResolution int

const (
  DanglingSymlinkResolutionPassThrough DanglingSymlinkResolution = iota
  DanglingSymlinkResolutionIgnore
  DanglingSymlinkResolutionError
)

def SymlinkOpt struct {
  ConvertToFiles bool
  DanglingSymlink DanglingSymlinkResolution
}
```

We could add more details later, like absolute symlink resolition. This assumes absolute paths will always be converted to files, btw, which I think we should start with (unless BuildGrid supports absolute symlinks?), so the new option will only trigger for relative symlinks.

peterebden:
As discussed in #146 I don't have capacity to do more here at present, @k-ye is kindly looking into the issue.
Will abandon this, happy for others to reuse parts or ignore as they see fit :)

Issue #44, opened 2019-07-15
Don't upload the same input concurrently


ola-rozenfeld:
Similar to #43, but the implementation is not super-trivial: in order to do this, we need a map of digest->channel (in Go, or Future in Java), so that a function that needs to write a bunch of blobs knows which futures it should wait on. This way, if WriteBlobs(b1, b2) is called followed by WriteBlobs(b1, b3) , both functions will wait on the same future that will write the b1 blob.
It will, however, result in significant upload cost savings (I believe) on an uncached build for clients using a hermetic toolchain.

PR #189, opened 2020-09-30
Deduping concurrent write requests.

### LINKED ISSUE #43: Don't upload previously uploaded blobs
This should be configurable with a parameter, but the default behavior should be to avoid querying GetMissingBlobs for previously uploaded blobs, drastically reducing its time. However, GetMissingBlobs doesn't take all that long in itself, so this is lower priority. It is, however, very easy to implement.

ola-rozenfeld:
Fixes #43 and #44, because this also doesn't query GetMissingBlobs for
previously uploaded blobs.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -25,6 +25,47 @@ import (
 	log "github.com/golang/glog"
 )
 
+func (c *Client) afterUpload(batch []digest.Digest, err error) {
+	for _, dg := range batch {
+		l, loaded := c.casUploadLocks.Load(dg)
+		lock, ok := l.(*sync.RWMutex)
+		if !ok {
+			log.Errorf("Unexpected type found in locks map")
+		}
+		if !loaded {
+			log.Errorf("Precondition failed: upload lock not found on %v.", dg)
+		}
+		if err != nil {
+			c.casUploadErrors.Store(dg, err)
+		}
+		lock.Unlock()
+	}
+}
+
+// Reset clears the Client state (e.g. makes it forget which blobs were
+// previously uploaded, or in the process of being uploaded).
+// This function is not thread safe, and should only be used outside of
+// ongoing client operations!
+func (c *Client) Reset() {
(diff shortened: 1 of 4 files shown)


go/pkg/client/cas_test.go
```diff
@@ -375,6 +376,88 @@ func TestMissingBlobs(t *testing.T) {
...
+				go func() {
+					var input []*chunker.Chunker
+					for _, blob := range blobs {
+						input = append(input, chunker.NewFromBlob(blob, 10))
+					}
+					_, err := c.UploadIfMissing(ctx, input...)
+					if err != nil {
+						t.Errorf("c.UploadIfMissing(ctx, input) gave error %v, expected nil", err)
+					}
+					defer wg.Done()
```
ramymedhat, line 443: defer at the top?

go/pkg/client/cas.go
```diff
@@ -99,9 +168,27 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker)
...
+		}
+		if v, _ := c.casUploadErrors.Load(dg); v != nil {
+			e, ok := v.(error)
+			if !ok {
+				err = fmt.Errorf("unexpected type found in error map")
+			}
+			err = e
+		}
+	}
+	log.V(2).Info("Done")
```
ramymedhat, line 191: Can you make this more descriptive about what was finished?

ramymedhat (commented):
Thanks! Did you run the tests with feature=race?

go/pkg/client/cas.go
ola-rozenfeld: Yep!

go/pkg/client/cas_test.go
ola-rozenfeld: Oh, right :-)

ola-rozenfeld:
Yes, features=race and --runs_per_test=50.

ramymedhat (approved):

PR #190, opened 2020-10-01
Reading server capabilities at client startup.

### LINKED ISSUE #99: Query the cache server Capabilities to determine max batch size
When trying to use `rexec` against [BuildGrid](https://gitlab.com/buildgrid/buildgrid), I got an error trying to batch upload/download blobs. Specifically because the blobs in the message was larger than what was advertised in the configured max batch size.  It looks like the max batch size (MaxBatchSz) is hardcoded to a value just under 4MB, whereas BuildGrid advertised 2MB as part of it's CacheCapabilities: https://github.com/bazelbuild/remote-apis-sdks/blob/3e40a08e09f4c1de8e36f4b493954448dc506f2c/go/client/cas.go#L139

Ideally, the cas client would check the CacheCapabilities endpoint for this value, and if set use that instead of it's default. It looks like no capabilities logic is currently in `remote-apis-sdk`, so maybe this would be a good starting point?

ola-rozenfeld:
For now, only check digest function compatibility and set MaxBatchSize
accordingly. Fixes #99.
Still TODO: check API version compatibility, cache and execution
priorities once we implement those.

### DIFF
--- a/go/pkg/client/BUILD.bazel
+++ b/go/pkg/client/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "bytestream.go",
+        "capabilities.go",
         "cas.go",
         "client.go",
         "client_context.go",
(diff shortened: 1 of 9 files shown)


ola-rozenfeld:
I generalize the Digest function used in the client, although not to the point of allowing it to be set with a flag (although this now should be easy to do as well, if there is demand).

ramymedhat (approved):

PR #191, opened 2020-10-01
Add more TLS config flags, and support remote caching mode

asuffield:
These changes are sufficient to persuade remotetool to interact with
my remote caching setup. Notably:

* TLS wiring added
* In cache-only mode, bazel does not upload input roots, so tolerate errors
* Show actions without results, to handle actions which failed
* Render a few interesting fields that were previously skipped

### DIFF
--- a/go/pkg/flags/flags.go
+++ b/go/pkg/flags/flags.go
@@ -3,7 +3,10 @@ package flags
 
 import (
 	"context"
+	"crypto/x509"
 	"flag"
+	"fmt"
+	"io/ioutil"
 	"time"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/balancer"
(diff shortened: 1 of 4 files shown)


go/pkg/tool/tool.go
```diff
@@ -245,42 +248,86 @@ func (c *Client) ShowAction(ctx context.Context, actionDigest string) (string, e
...
 	return showActionRes.String(), nil
 }
 
 func (c *Client) getOutputs(ctx context.Context, actionRes *repb.ActionResult) (string, error) {
 	var res bytes.Buffer
-	res.WriteString("Output Files:\n=============\n")
+
+	res.WriteString("------------------------------------------------------------------------\n")
+	res.WriteString("Action Result\n\n")
+	res.WriteString(fmt.Sprintf("Exit code: %d\n", actionRes.ExitCode))
```
ola-rozenfeld, line 312: Note: tt's always zero, because failed actions are not cached. But it won't hurt displaying it.

ola-rozenfeld (commented):
Thank you for doing this!

go/pkg/tool/tool.go
asuffield: Hah, and I even just fixed the other bit of the code to not call this function in that scenario (because the ActionResult doesn't exist). Why do we even have that field?

go/pkg/flags/flags.go
```diff
@@ -76,12 +85,25 @@ func NewClientFromFlags(ctx context.Context, opts ...client.Opt) (*client.Client
 		}
 		opts = append(opts, client.RPCTimeouts(timeouts))
 	}
+	certPool := x509.NewCertPool()
```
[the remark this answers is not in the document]
asuffield: Done.

go/pkg/tool/tool.go
ola-rozenfeld: Because ActionResult doesn't always come from the remote cache, it is also returned by remote execution, where it can fail.

ola-rozenfeld (approved):

ola-rozenfeld:
Ooops, this broke the client. I am looking into why now, will either roll back or fix forward.

ola-rozenfeld:
Fixed in #195

PR #192, opened 2020-10-01
Fixing streamed RPC timeouts.

ola-rozenfeld:
This will effectively set the timeouts at 1 minute for
Write/Read/GetTree.


go/pkg/client/bytestream.go
```diff
@@ -21,14 +21,12 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
...
-	opts := c.RPCOpts()
 	defer cancel()
-	closure := func() error {
+	closure := func(ctx context.Context) error {
 		ch.Reset() // Retry by starting the stream from the beginning.
 		// TODO(olaola): implement resumable uploads.
 
-		// Use lower-level Write in order to not retry twice.
-		stream, err := c.byteStream.Write(cancelCtx, opts...)
+		stream, err := c.Write(cancelCtx)
```
gkousik, line 29: Should this be ctx from the closure instead of cancelCtx (also maybe rename the "ctx" here to something else to disambiguate from the parent "ctx")?

gkousik (approved):

PR #195, opened 2020-10-02
Bugfix: keeping previous codepath when TLS is enabled.

ola-rozenfeld:


ola-rozenfeld:
This is a forward fix to #191

PR #196, opened 2020-10-03
Fixing low-level streaming wrappers to make sense

ola-rozenfeld:


ola-rozenfeld:
This should be submitted after #192 . Actually, let me patch this commit there instead.

PR #198, opened 2020-10-04
Add --service_no_security to connect to the RBE service without TLS

Yannic:
This is useful, e.g., for connecting to local servers without TLS (e.g.,
for debugging).


ola-rozenfeld:
Wait, `--use_rpc_credentials=false` doesn't already do an equivalent of that? Maybe it should?

Yannic:
> Wait, `--use_rpc_credentials=false` doesn't already do an equivalent of that? Maybe it should?

Not it's not. From the docs of `TransportCredsOnly` (which is controlled by `--use_rpc_credentials`):

> TransportCredsOnly is true if it's the caller's responsibility to set per-RPC credentials
> on individual calls. This overrides ActAsAccount, UseApplicationDefault, and UseComputeEngine.
> This is not the same as NoSecurity, as transport credentials will still be set.

ola-rozenfeld:
I see, thank you!

ola-rozenfeld (approved):

PR #164, opened 2020-09-01
cas: utilize locality when writing files to disk in DownloadFiles

atetubou:
This improves time of DownloadFiles in n1-standard-8 Windows GCE with standard disk when downloading 14k+ files from 125s to 75s (14s to 11s in one time measure on Linux).

PR #200, opened 2020-10-05
Revert "Revert "cas: utilize locality when writing files to disk in D…

atetubou:
…ownloadFiles" (#170)"

This reverts commit a7fc931d67e13f358ae344b40c9217fc1455338d.
And this is also re-land of https://github.com/bazelbuild/remote-apis-sdks/pull/164

This moves digest sort only to UploadIfMissing to utilize filepath
locality only in download.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -53,7 +53,7 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker) 
 	log.V(2).Infof("%d items to store", len(missing))
 	var batches [][]digest.Digest
 	if c.useBatchOps {
-		batches = c.makeBatches(missing)
+		batches = c.makeBatches(missing, true)
 	} else {
 		log.V(2).Info("Uploading them individually")
 		for i := range missing {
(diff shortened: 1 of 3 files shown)


go/pkg/client/cas.go
```diff
@@ -377,12 +380,9 @@ func (c *Client) BatchDownloadBlobs(ctx context.Context, dgs []digest.Digest) (m
 func (c *Client) makeBatches(dgs []digest.Digest) [][]digest.Digest {
 	var batches [][]digest.Digest
 	log.V(2).Infof("Batching %d digests", len(dgs))
-	sort.Slice(dgs, func(i, j int) bool {
-		return dgs[i].Size < dgs[j].Size
-	})
 	for len(dgs) > 0 {
-		batch := []digest.Digest{dgs[len(dgs)-1]}
```
atetubou, line 384: I think this, using back instead of front, optimization is for "large and small blobs hitting max exactly" test case, but that test case is bit artifical rather than pratical?

If reclient really needs this optimization, I will keep this optimization.

ola-rozenfeld: The way I'd like to see this change landing is: rename the makeBatches function to makeBatchesOptimizeSize, and add another function named makeBatchesOptimizeLocality. Use makeBatchesOptimizeLocality for downloads, and probably on Windows to start, until there are some load test results to show benefits for Linux as well.
Keep the old tests (limit them by platform, if needed), and add Windows-only new test cases for the new functionality.

atetubou: Due to passing array of digest is not sufficient for `makeBatchesOptimizeLocality`, I changed to pass additional flag to `makeBatches`. Also expose that as behavior flag and disabled locality ordering by default rather than setting it based on OS. How do you think this?

ola-rozenfeld: Better! I am in the process of a major rewrite to the upload/download logic (I will have a single uploader/downloader to manage uploads/downloads from many simultaneous actions), so I suspect I'll refactor this code, but I'll keep and respect your UtilizeLocality option. And if we want to make it the default later, we could.

go/pkg/client/client.go
```diff
@@ -78,7 +78,9 @@ type Client struct {
 	// ExecutableMode is mode used to create executable files.
 	ExecutableMode os.FileMode
 	// RegularMode is mode used to create non-executable files.
-	RegularMode     os.FileMode
+	RegularMode os.FileMode
+	// UtilizeLocality is to specify whether client downloads files utilizing disk access locality.
+	UtilizeLocality bool
```
ola-rozenfeld, line 83: Please make it a client Opt, like ChunkMaxSize or StartupCapabilities.

atetubou: done

atetubou:
@ola-rozenfeld is it possible to merge this first if https://github.com/bazelbuild/remote-apis-sdks/pull/203 took time?

I'd like to see another bottleneck after this PR in our client.

ola-rozenfeld:
Yes! Sorry about that, I thought it would all be in by now.

ola-rozenfeld (approved):

PR #202, opened 2020-10-07
Fixing deadlock in upload: context cancellation

ola-rozenfeld:


go/pkg/client/cas.go
```diff
@@ -113,6 +113,9 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker)
 
 	missing, present, err := c.getMissingPresent(ctx, newUploads)
 	if err != nil {
+		for _, d := range newUploads {
```
gkousik, line 116: nit: have a locally defined cleanup function and call it here and down below?

ola-rozenfeld: Oh, right, I have that function! Why didn't I call it?...

gkousik (approved):

PR #203, opened 2020-10-08
Unified uploader: handle context cancellations correctly.

ola-rozenfeld:


ola-rozenfeld:
Notes:
* I am in the process of adding more tests, will update.
* This required regressing the functionality of remembering Missing blobs on the Metadata (because there is no easy way to pass that through).

ola-rozenfeld:
Added more tests, and made sure to run with --features=race

ola-rozenfeld:
Fixes #206

ola-rozenfeld:
Added the unified downloader in the same CL, since they are similar and I believe together will be easier to review/address comments. Added more tests and improved existing tests. Also cleaned up some minor things:
- use errgroup instead of waitGroup in tests
- use WeightedSemaphore for casConcurrency instead of chan bool

go/pkg/client/cas_test.go
```diff
@@ -458,14 +458,57 @@ func TestUploadConcurrent(t *testing.T) {
 	}
 }
```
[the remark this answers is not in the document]
ola-rozenfeld: Done.

go/pkg/client/cas.go
```diff
@@ -25,32 +26,122 @@ import (
 	log "github.com/golang/glog"
 )
 
-// UploadIfMissing stores a number of uploadable items.
-// It first queries the CAS to see which items are missing and only uploads those that are.
-// Returns a slice of the missing digests.
-func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker) ([]digest.Digest, error) {
```
atetubou, line 31: could you keep interface unchanged?
we rely on missing digest returned by this function.

ola-rozenfeld: Done!

go/pkg/client/cas.go
```diff
@@ -722,53 +904,97 @@ func (c *Client) DownloadFiles(ctx context.Context, execRoot string, outputs map
...
 
-	eg, eCtx := errgroup.WithContext(ctx)
+	ctx := context.Background()
 	for i, batch := range batches {
-		i, batch := i, batch     // https://golang.org/doc/faq#closures_and_goroutines
-		c.casDownloaders <- true // Reserve an downloader thread.
-		eg.Go(func() error {
-			defer func() { <-c.casDownloaders }()
+		i, batch := i, batch // https://golang.org/doc/faq#closures_and_goroutines
+		go func() (err error) {
```
atetubou, line 910: err is ignored?

ola-rozenfeld: No, not ignored -- it is handled by the deferred afterDownload in lone 935.

atetubou: Then better to declare err in the beginning of this function and deferred call afterDownload there?

`return err` in this function seems meaningless and bit confusing to me.

ola-rozenfeld: This will not do the same thing -- note that I only defer afterDownload for the case of a single-blob batch; for the multi-blob batch case I handle everything manually. Perhaps, to make this clearer, I could split the function into two functions `readBatch` and `readSingle`, and only return/defer on the readSingle one? I'm not sure it will be clearer, though...

go/pkg/client/cas.go
```diff
@@ -62,48 +163,78 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*chunker.Chunker)
...
+		case resp := <-wait:
+			if resp.err != nil {
+				return nil, resp.err
+			}
+			if resp.missing {
+				missing = append(missing, resp.digest)
+			}
+			uploaded--
+			if uploaded == 0 {
+				return missing, nil
```
ramymedhat, line 234: I like how you handled reporting the missing blobs :-)

go/pkg/client/cas.go
ramymedhat: I'm also not clear on how the goroutine's returned err is handled. If you're using them then maybe use an errgroup?

go/pkg/client/cas.go
```diff
@@ -722,53 +904,97 @@ func (c *Client) DownloadFiles(ctx context.Context, execRoot string, outputs map
...
-						perm = c.ExecutableMode
-					}
-					if err := ioutil.WriteFile(filepath.Join(execRoot, out.Path), data, perm); err != nil {
-						return err
+					for _, r := range reqs[dg] {
+						perm := c.RegularMode
+						if r.output.IsExecutable {
+							perm = c.ExecutableMode
+						}
+						r.wait <- ioutil.WriteFile(filepath.Join(r.execRoot, r.output.Path), data, perm)
```
ramymedhat, line 929: Would it be possible to streamline where we respond to requests on the wait channel so its all done the same way? Currently sometimes we write to the channel directly, and in other places we use afterDownload.

go/pkg/client/cas.go
```diff
@@ -722,53 +904,97 @@ func (c *Client) DownloadFiles(ctx context.Context, execRoot string, outputs map
...
 					}
 				}
 			} else {
-				out := outputs[batch[0]]
-				path := filepath.Join(execRoot, out.Path)
-				log.V(3).Infof("Downloading single file with digest %s to %s", out.Digest, path)
-				if _, err := c.ReadBlobToFile(ctx, out.Digest, path); err != nil {
+				// We cannot release the lock after each file copy in this case, because
+				// the caller might move the file, and we don't have the contents in memory.
```
ramymedhat, line 934: Would it be simpler if we download the blob with the same digest multiple times? So that the download logic is keyed by the output path not the digest? Its rare enough that I think its not worth the extra logic in the code.

go/pkg/client/cas.go
```diff
@@ -25,32 +26,131 @@ import (
...
+		delete(dgMap, d)
+	}
+	for d := range dgMap {
+		present = append(present, d)
+	}
+	return missing, present, err
+}
+
+func (c *Client) uploadProcessor() {
+	var buffer []*uploadRequest
```
gkousik, line 67: I don't fully understand why we need this buffer. It is already buffered via casUploadRequests channel? Can you call `upload()` as and when you see new items in `casUploadRequests` ?

go/pkg/client/cas.go
```diff
@@ -25,32 +26,131 @@ import (
...
+	var newUploads []digest.Digest
+	for _, req := range reqs {
+		dg := req.c.Digest()
+		st, ok := c.casUploads[dg]
+		if ok {
+			st.mu.Lock()
+			if len(st.clients) > 0 {
+				st.clients = append(st.clients, req.wait)
+			} else {
+				req.wait <- &uploadResponse{err: nil, missing: false}
```
gkousik, line 127: IIUC, this is an error scenario? The digest exists in casUploaders, which means there should be some clients waiting on it but there aren't any. Should we log an error message if this happens?

go/pkg/client/cas.go
```diff
@@ -25,32 +26,131 @@ import (
...
+	for _, d := range missing {
+		delete(dgMap, d)
+	}
+	for d := range dgMap {
+		present = append(present, d)
+	}
+	return missing, present, err
+}
+
+func (c *Client) uploadProcessor() {
```
gkousik, line 66: nit (and feel free to ignore this): It maybe more readable to move the upload processor and the related functions to a separate file.

ola-rozenfeld (commented):
This is ready for another pass: I added the client option UnifiedCasOps which is false by default (I was thinking to change the default to true in a separate CL), so that it's easy to roll back.
I implemented the option by copy-pasting the old code -- I tried avoiding it, but it turned out to be quite difficult. I added test cases for both old and new behaviors.

go/api/command/command.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.22.0
+// 	protoc-gen-go v1.25.0
```
atetubou, line 3: It is better to have separate PR for this?
Or this is necessary for this PR?

go/pkg/tool/tool_test.go
```diff
@@ -215,7 +215,7 @@ func TestTool_UploadBlob(t *testing.T) {
 	cas := e.Server.CAS
 
 	tmpFile := path.Join(t.TempDir(), "blob")
-	if err := ioutil.WriteFile(tmpFile, []byte("Hello, World!"), 777); err != nil {
+	if err := ioutil.WriteFile(tmpFile, []byte("Hello, World!"), 0777); err != nil {
```
atetubou, line 218: It is better to have separate PR for this?
Or this is necessary for this PR?

go/api/command/command.pb.go
ola-rozenfeld: https://github.com/bazelbuild/remote-apis-sdks/pull/219

go/pkg/tool/tool_test.go
ola-rozenfeld: https://github.com/bazelbuild/remote-apis-sdks/pull/219

go/pkg/client/cas.go
gkousik: > If you only read from the channel, you are in danger of never stopping (if new items keep arriving)

I see, makes sense.

go/pkg/client/cas.go
```diff
@@ -25,13 +27,227 @@ import (
...
+	sort.Strings(iiKeys)
+	return &requestMetadata{
+		toolName:     strings.Join(tnKeys, ","),
+		actionID:     strings.Join(aiKeys, ","),
+		invocationID: strings.Join(iiKeys, ","),
+	}
+}
+
+func (c *Client) upload(reqs []*uploadRequest) {
+	const (
```
gkousik, line 149: nit: declare it at the top

go/pkg/client/cas.go
gkousik: Thanks, I think I understand this a little better now. So the `upload()` method is called sequentially by `uploadProcessor()` and `upload()` spawns multiple go-routines that actually do the upload.

An `uploadState()` will exist in `casUploaders`, if we have already seen an upload request for the digest. Now coming to this specific if condition:

```
if len(st.clients) > 0 {
...
} else {
...
}
```
IIUC, the else part will happen when one of the go-routines spawned by `upload()` finished doing the upload (either successfully or unsuccessfully), acquired `st.mu` before the code on line `161` did and finished notifying all the clients.

So now my question is, if `len(st.clients) == 0`, then it just means that we notified all the clients of the result. How do we know whether the upload itself was successful or not? We seem to be returning `&uploadResponse{err: nil, missing: false}` which is assuming that we didn't run into an error with the upload?

go/pkg/client/cas.go
```diff
@@ -25,13 +27,227 @@ import (
...
+	var metas []*requestMetadata
+	for _, req := range reqs {
+		dg := req.c.Digest()
+		st, ok := c.casUploads[dg]
+		if ok {
+			st.mu.Lock()
+			if len(st.clients) > 0 {
+				st.clients = append(st.clients, req.wait)
+			} else {
+				req.wait <- &uploadResponse{err: nil, missing: false}
```
gkousik, line 165: Minor nit: Include `digest` in the response here for consistency (or document why we don't need it - i.e., we need the digest only when `missing = true`).

go/pkg/client/cas.go
ola-rozenfeld: Done. I should just implement log.EveryN()...

go/pkg/client/cas.go
ola-rozenfeld: Correct, thank you, that was a mistake! I should have returned st.err. Done.

go/pkg/client/cas.go

gkousik (approved):

gkousik:
I would wait for Ramy's review as well.

go/pkg/client/cas.go
```diff
@@ -25,32 +26,132 @@ import (
...
-			dgs = append(dgs, dg)
-			chunkers[dg] = c
+	// Collect new uploads.
+	newStates := make(map[digest.Digest]*uploadState)
+	var newUploads []digest.Digest
+	for _, req := range reqs {
+		dg := req.c.Digest()
+		st, ok := c.casUploads[dg]
+		if ok {
+			st.mu.Lock()
```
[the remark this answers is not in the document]
ramymedhat: Ah gotcha, thanks for clarifying! Do you mind adding this to the comment on mu? There's an extensive comment there about what it doesn't protect, but may be useful to also describe what it does protect.

ramymedhat (approved):

PR #210, opened 2020-10-15
Revert concurrent uploads

gkousik:
Running a build with this version of SDK to test and make sure that a second cache-silo'ed build gets cache hits and that remote-execution calls are made.


ola-rozenfeld (approved):

ola-rozenfeld:
Reason we are reverting this: the revert is temporary until https://github.com/bazelbuild/remote-apis-sdks/pull/203 goes in. We need our own client release to use a version without the upload deduping (we suspect it might have caused stuck builds in outlier cases), hence the revert (other options included branching).

Other remedies include: implement unified uploads/downloads as an option; implement resumable uploads; implement progress-based timeouts for uploads/downloads (e.g. don't time the request out if it's slow, only if it's stuck).

PR #212, opened 2020-10-21
revive c.RPCOpts in GetTree

atetubou:
I think this is removed accidentaly in https://github.com/bazelbuild/remote-apis-sdks/commit/0dbc1039e6dad925ee9a461a66bbf1b2f35693f1 .

PR #213, opened 2020-10-21
fix for #212

atetubou:
I think this is more appropriate fix for https://github.com/bazelbuild/remote-apis-sdks/commit/0dbc1039e6dad925ee9a461a66bbf1b2f35693f1 .


gkousik (approved):
This LGTM, but I would wait for Ola's review as well :)

ola-rozenfeld:
Huh. Sorry about that, I was sure I changed it!

PR #215, opened 2020-10-22
Improving string representations for InputSpec members.

ola-rozenfeld:


ola-rozenfeld:
Resulting format:

```
{Path:foo/empty_dir Contents:[] IsExecutable:false IsEmptyDirectory:true}
{Path:foo/bar Contents:[98 97 114 45 99 111 110 116 101 110 116 115] IsExecutable:false IsEmptyDirectory:false}] 
```

```
[{Regex:*.bla Type:DirectoryInputType} {Regex:*.blo Type:FileInputType}]
```

PR #216, opened 2020-10-22
Log execution-id in addition to command-id in the SDK

gkousik:
This is to be consistent with remote execution clients.

### DIFF
--- a/go/pkg/rexec/rexec.go
+++ b/go/pkg/rexec/rexec.go
@@ -136,9 +136,9 @@ func (ec *Context) computeInputs() error {
 	}
 	ec.Metadata.EventTimes[command.EventComputeMerkleTree] = &command.TimeInterval{From: time.Now()}
 	defer func() { ec.Metadata.EventTimes[command.EventComputeMerkleTree].To = time.Now() }()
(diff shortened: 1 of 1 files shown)


atetubou (approved):

ramymedhat (commented):
Can we query gws logs with the execution ID?

gkousik:
Nope I don't think so. You could go from execution-id -> command-id (from reproxy_log), then query on command-id. I think that's ok because GWS logs can't be queried immediately anyway (there's a 10-15m delay) and by that time, you can probably lookup reproxy_log.

Alternatively I could modify this to log both command and execution-id. These are V(1) logs anyway, so doesn't hurt to log both.

ola-rozenfeld:
Maybe log the command ID on every line, and execution ID somewhere once in the beginning of a function. There are a couple of differences between them -- command id is stable; command id is much shorter. Logging both will be a lot less readable, imo.

gkousik:
Logged both command-id and execution-id as we discussed offline.

ramymedhat (approved):
Thanks! Very timely :)

PR #219, opened 2020-10-29
Fixing test file permission

ola-rozenfeld:


ola-rozenfeld:
I have no idea why the command proto changed, I guess maybe difference in Bazel versions?

atetubou (approved):

PR #221, opened 2020-10-30
Download stdout/stderr regardless of whether DownloadOutputs is set

ramymedhat:
not.


ola-rozenfeld (commented):
Wait, why are we doing this? Is it because we'll need it for not downloading intermediates?

PR #223, opened 2020-11-02
Supporting proper download cancellations

ola-rozenfeld:


go/pkg/client/cas.go
```diff
@@ -1054,6 +1055,10 @@ func (c *Client) downloadSingle(ctx context.Context, dg digest.Digest, reqs map[
 	}
 	r := rs[0]
 	rs = rs[1:]
```
ramymedhat, line 1057: Can you add a comment as to what you're checking for here? I assume making sure there is only one requestor for the blob so we can use their context?

go/pkg/client/cas.go
```diff
@@ -975,6 +975,7 @@ func copyFile(srcExecRoot, dstExecRoot, from, to string, mode os.FileMode) error
 type downloadRequest struct {
 	digest   digest.Digest
 	execRoot string
+	context  context.Context
```
ramymedhat, line 978: I tend to try to avoid embedding a context.Context in structs. I guess the other alternative is to monitor a channel? Would that be too complicated if it just covers the case of downloadSingle as you currently do?

go/pkg/client/cas.go
```diff
@@ -1054,6 +1055,10 @@ func (c *Client) downloadSingle(ctx context.Context, dg digest.Digest, reqs map[
 	}
 	r := rs[0]
 	rs = rs[1:]
+	if len(rs) == 0 {
+		// Download on same context as the issuing request, to support proper cancellation.
```
ramymedhat, line 1059: Would it be possible to set the right context before calling downloadSingle? I think its more clear to know that the function uses the context you passed to it and not potentially another context from a side channel.

go/pkg/client/cas_test.go
```diff
@@ -1551,3 +1551,39 @@ func TestDownloadFiles(t *testing.T) {
...
+	})
+	eg.Go(func() error {
+		cancel()
+		return nil
+	})
+	if err := eg.Wait(); err != nil {
+		t.Error(err)
+	}
+	if fake.BlobReads(fooDigest) != 0 {
+		t.Errorf("Expected no reads for foo, since request is cancelled, got %v.", fake.BlobReads(fooDigest))
```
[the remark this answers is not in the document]
ola-rozenfeld: Done.

go/pkg/client/cas.go
ola-rozenfeld: Yep, done!

go/pkg/client/cas.go
ola-rozenfeld: Yes, I thought it would be needlessly complicated. Instead of a context the request would include a channel, which the client would then ping, and we would create our own context.WithCancel, a goroutine waiting on that channel that would call cancel on the new context (see upload cancellations PR for comparison). I can do all that -- but I thought that would be needlessly complicated. Why not we embed a context on a struct in this case?

go/pkg/client/cas.go
ola-rozenfeld: Agreed, done.

go/pkg/client/cas.go
ramymedhat: There's a lot of threads about why the guideline exists, but I think Kaveh's response here is the most appropriate: https://groups.google.com/a/google.com/forum/#!searchin/go-nuts/context$20in$20struct/go-nuts/MAp-5Zt9z7M/qORACKu-AQAJ

ramymedhat (approved):

PR #224, opened 2020-11-03
Supporting single upload cancellation

ola-rozenfeld:


go/pkg/client/cas.go
```diff
@@ -91,10 +93,40 @@ func (c *Client) uploadProcessor() {
...
+			for _, r := range buffer {
+				if r.c != req.c || r.wait != req.wait {
+					newBuffer = append(newBuffer, r)
+				}
+			}
+			buffer = newBuffer
+			st, ok := c.casUploads[req.c.Digest()]
+			if ok {
+				st.mu.Lock()
+				var newClients []chan<- *uploadResponse
```
ramymedhat, line 115: Can you add a comment clarifying newClients? Something along the lines of it being a list of the remaining clients after the current cancellation requestor drops out?

go/pkg/client/cas.go
```diff
@@ -58,6 +59,7 @@ type uploadState struct {
 	// exactly once for a given state object (this is the whole point of the algorithm).
 	mu      sync.Mutex
 	clients []chan<- *uploadResponse
+	cancel  func()
```
ramymedhat, line 62: Can you update the comment above to indicate that mu now also protects the cancel function?

ramymedhat (approved):

go/pkg/client/cas.go
ola-rozenfeld: Done.

go/pkg/client/cas.go
ola-rozenfeld: Did one better and renamed the variable so it's clear.

PR #225, opened 2020-11-03
Bugfix: normalizing input paths relative to exec root

ola-rozenfeld:


go/pkg/tree/tree.go
```diff
@@ -63,6 +63,10 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 // recursively), and loads their contents into the provided map.
 func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileNode, chunkSize int, cache filemetadata.Cache) error {
 	absPath := filepath.Clean(filepath.Join(execRoot, path))
+	normPath, err := filepath.Rel(execRoot, absPath)
+	if err != nil {
```
ramymedhat, line 67: I don't think this actually ensures that absPath is under execRoot. filepath.Rel can return a path with ../

https://play.golang.com/p/V497MOsGPn_e

ramymedhat (approved):

PR #226, opened 2020-11-03
Implementing progress-based timeouts for Read/Write.

ola-rozenfeld:


ramymedhat (commented):
I may be misunderstanding the current tests, but can we also have a test to simulate partial progress instead of just timing out on the first n calls? That way we are testing the scenario of a slow connection or a connection with probabilistic packet loss?

ramymedhat (approved):

PR #227, opened 2020-11-03
Adding action id to CAS verbose logging

ola-rozenfeld:


ola-rozenfeld:
I haven't tried this on a large build yet, but I wouldn't be surprised if it slowed things down considerably. This might be okay, though, as verbose logging should be used sparingly, only to debug things...

go/pkg/client/client_context.go
```diff
@@ -18,6 +18,19 @@ const (
 	remoteHeadersKey = "build.bazel.remote.execution.v2.requestmetadata-bin"
 )
 
+// LogContextInfof(ctx, x, ...) is equivalent to log.V(x).Infof(...) except it
+// also logs context metadata, if available.
+func LogContextInfof(ctx context.Context, v log.Level, format string, args ...interface{}) {
```
ramymedhat, line 23: nit: Can you rather extend the Verbose type with a function that logs the context info?

log.V(2).ContextInfof(...)?

ramymedhat: nvm this won't work with a package level function, and an embedded struct in the client context package is not really worth it. LGTM.

ramymedhat (approved):

PR #228, opened 2020-11-03
Make Chunker.Next read "generic" and independent from digest size.

rubensf:
These are both pre-works for adding write compression support to the
remote-apis-sdks.

Chunker.Next should be "generic" as it allows for a drop in implementation
of a reader that compresses on the fly. I still left the special case of
caching data in memory as it saves the effort from memory copies.

Making the chunker reads work independently from the digest size is also
useful as it allows us to not have to pre-compute the digest of the
compressed blobs. The current draft of the RE API never requires the
digest of the compressed blob at any point, and this saves us the
trouble of needing to read the data twice.

Notice that this implies that the chunker won't be actually matching
data against the supplied digest at any moment. The digest is now purely
information storage, rather than necessary for chunker logic.

As a caveat, for simplicity, I made it that we only cache in memory
files that are smaller than the *chunk* size rather than the IO buffer
size.

### DIFF
--- a/go/pkg/chunker/chunker.go
+++ b/go/pkg/chunker/chunker.go
@@ -112,7 +112,7 @@ func (c *Chunker) ChunkSize() int {
 // TODO(olaola): implement Seek(offset) when we have resumable uploads.
 func (c *Chunker) Reset() {
 	if c.r != nil {
-		c.r.Seek(0)
+		c.r.SeekOffset(0)
 	}
 	c.offset = 0
 	c.reachedEOF = false
--- a/go/pkg/chunker/chunker.go
+++ b/go/pkg/chunker/chunker.go
@@ -2,12 +2,10 @@
 package chunker
 
 import (
-	"bufio"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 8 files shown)


go/pkg/reader/reader_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func setTmpFile(t *testing.T, content []byte) (string, func()) {
+	t.Helper()
+	execRoot, err := ioutil.TempDir("", "")
```
ola-rozenfeld, line 15: nit: We already have something like this in filemetadata_test.go. We should just centralize it in some test util package. Also, no real need to clean up temp files.

go/pkg/reader/reader_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+			n, err := io.ReadFull(r, data)
+			if n != tc.dataBuffSize {
+				t.Errorf("Read() = %d bytes, expected %d", n, tc.dataBuffSize)
+			}
+			if err != nil {
+				t.Errorf("Read() = %v err, expected nil", err)
+			}
+			if diff := cmp.Diff(data, tc.blob[:tc.dataBuffSize]); diff != "" {
+				t.Errorf("Read() = incorrect result, diff(-wang, +got): %v", diff)
```
ola-rozenfeld, line 82: typo: want

rubensf: Done.

go/pkg/reader/reader_test.go
rubensf: Rebased this on top of newly added #234.

rubensf:
Sorry about the `outdated` comments! I forget github would make code reviewing harder when commit hashes "disappear" (due to eg rebasing, which changes the hash I guess?)

go/pkg/chunker/chunker.go
```diff
@@ -149,47 +156,54 @@ func (c *Chunker) Next() (*Chunk, error) {
...
+
+		// We don't need to check the amount of bytes read, as ReadFull will yell if
+		// it's diff than len(data).
+		data = make([]byte, c.chunkSize)
+		n, err := io.ReadFull(c.r, data)
+		data = data[:n]
+		switch err {
+		case io.ErrUnexpectedEOF:
+			// Cache the contents to avoid further IO for small files.
+			if c.offset == 0 {
```
ola-rozenfeld, line 191: Why not do this for both EOF cases?

rubensf: Hm, I did it like this because [ReadFull](https://golang.org/pkg/io/#ReadFull) only returns io.EOF if no bytes are read from `c.r`, which shouldn't realistically reach this point for `offset == 0` (it'll be caught at the Digest.Size == 0 check above, or it'll be some sort of error for the compressed bytes to be zero sized). 

In retrospect, the code is cleaner to read and not any less correct by grouping both behaviors. Done!

ola-rozenfeld (approved):

PR #229, opened 2020-11-05
Support symlinks when constructing a Merkle tree.

k-ye:
Added `TreeSymlinkOpts` to control the behavior of tree construction.

To be backward compatible, symlink is preserved only when `TreeSymlinkOpts.Preserved` is explicitly set to `true`. By default, symlinks are converted into files.

Issue: #146

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -85,7 +85,9 @@ type Client struct {
 	// UtilizeLocality is to specify whether client downloads files utilizing disk access locality.
 	UtilizeLocality UtilizeLocality
 	// UnifiedCASOps specifies whether the client uploads/downloads files in the background
-	UnifiedCASOps       UnifiedCASOps
+	UnifiedCASOps UnifiedCASOps
+	// TreeSymlinkOpts controls how symlinks are handled when constructing a tree.
+	TreeSymlinkOpts     TreeSymlinkOpts
 	serverCaps          *repb.ServerCapabilities
 	useBatchOps         UseBatchOps
 	casConcurrency      int64
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -173,7 +173,9 @@ func (s UnifiedCASOps) Apply(c *Client) {
 }
 
 func (o *TreeSymlinkOpts) Apply(c *Client) {
-	c.TreeSymlinkOpts = *o
+	if o != nil {
+		c.TreeSymlinkOpts = *o
+	}
 }
 
 // MaxBatchDigests is maximum amount of digests to batch in batched operations.
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -87,7 +87,7 @@ type Client struct {
 	// UnifiedCASOps specifies whether the client uploads/downloads files in the background
 	UnifiedCASOps UnifiedCASOps
 	// TreeSymlinkOpts controls how symlinks are handled when constructing a tree.
-	TreeSymlinkOpts     TreeSymlinkOpts
+	TreeSymlinkOpts     *TreeSymlinkOpts
 	serverCaps          *repb.ServerCapabilities
 	useBatchOps         UseBatchOps
 	casConcurrency      int64
@@ -173,9 +173,7 @@ func (s UnifiedCASOps) Apply(c *Client) {
 }
 
 func (o *TreeSymlinkOpts) Apply(c *Client) {
-	if o != nil {
-		c.TreeSymlinkOpts = *o
-	}
+	c.TreeSymlinkOpts = o
 }
 
 // MaxBatchDigests is maximum amount of digests to batch in batched operations.
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -172,6 +172,10 @@ func (s UnifiedCASOps) Apply(c *Client) {
 	c.UnifiedCASOps = s
 }
 
+func (o *TreeSymlinkOpts) Apply(c *Client) {
+	c.TreeSymlinkOpts = *o
+}
+
 // MaxBatchDigests is maximum amount of digests to batch in batched operations.
 type MaxBatchDigests int
 
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
@@ -78,6 +79,43 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 	return false
 }
 
+// getTargetRelPath returns the part of target that is relative to execRoot,
+// iff. target is under execRoot. Otherwise it returns an error.
+func getTargetRelPath(execRoot, symlink, target string) (string, error) {
+	if !filepath.IsAbs(target) {
+		target = filepath.Clean(filepath.Join(execRoot, target))
+	}
+	rel, err := filepath.Rel(execRoot, target)
+	if err != nil {
+		return "", err
+	}
+	if strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
+		return "", fmt.Errorf("symlink=%q (target=%q) is not under execRoot=%q", symlink, target, execRoot)
+	}
+	return rel, nil
+}
+
+// preprocessSymlink returns two things: if the routine should continue, and if
+// there is an error to be reported back.
+func preprocessSymlink(execRoot, symlink string, meta *filemetadata.SymlinkMetadata, opts *TreeSymlinkOpts) (bool, error) {
+	if meta.IsDangling {
+		// For now, we do not treat a dangling symlink as an error. In the case
+		// where the symlink is not preserved (i.e. needs to be converted to a
+		// file), we simply ignore this path in the finalized tree.
+		return opts.Preserved, nil
+	}
+	if !opts.Preserved {
+		// We will convert the symlink to a normal file, so it doesn't matter
+		// where target is under execRoot or not.
+		return true, nil
+	}
+
+	if _, err := getTargetRelPath(execRoot, symlink, meta.Target); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
 // loadFiles reads all files specified by the given InputSpec (descending into subdirectories
 // recursively), and loads their contents into the provided map.
 func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileSysNode, chunkSize int, cache filemetadata.Cache, opts *TreeSymlinkOpts) error {
@@ -91,14 +129,14 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 	}
 	meta := cache.Get(absPath)
 	isSymlink := meta.Symlink != nil
-	if isSymlink && meta.Symlink.IsDangling {
-		if !opts.Preserved {
+	if isSymlink {
+		cont, err := preprocessSymlink(execRoot, path, meta.Symlink, opts)
+		if err != nil {
+			return err
+		}
+		if !cont {
 			return nil
 		}
-		// Right now we never treat a dangling symlink as an error. If we
-		// choose not to preserve the symlink, the path is simply skipped.
-		// Otherwise we will create a corresponding symlink node in the tree
-		// (even if the target does not exist).
 	} else if meta.Err != nil {
 		return meta.Err
 	}
@@ -123,13 +161,18 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 		return nil
 	} else if t == command.SymlinkInputType {
 		fs[path] = &fileSysNode{
-			// TODO: How to handle a target with absolute path?
 			symlink: &symlinkNode{target: meta.Symlink.Target},
 		}
-		// NOTE: The target is not followed. It is the users' responsibility
-		// to explicitly list the target in their InputSpec. This applies to
-		// symlinks to either a file or a directory.
-		return nil
+		if meta.Symlink.IsDangling {
+			return nil
+		}
+		// If we can proceed here, we've already verified that target is
+		// under execRoot, so this should never fail.
+		relTarget, err := getTargetRelPath(execRoot, path, meta.Symlink.Target)
+		if err != nil {
+			panic(fmt.Sprintf("Invalid target=%q err=%v", relTarget, err))
+		}
+		return loadFiles(execRoot, excl, relTarget, fs, chunkSize, cache, opts)
 	}
 	// Directory
 	files, err := ioutil.ReadDir(absPath)
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -64,6 +64,15 @@ type TreeSymlinkOpts struct {
 	// By default, a symlink is converted into its targeted file.
 	// If true, preserve the symlink.
 	Preserved bool
+	// If true, the symlink target (if not dangling) is followed.
+	FollowsTarget bool
+}
+
+// NewTreeSymlinkOpts returns a default NewTreeSymlinkOpts object.
+func NewTreeSymlinkOpts() *TreeSymlinkOpts {
+	return &TreeSymlinkOpts{
+		FollowsTarget: true,
+	}
 }
 
 // shouldIgnore returns whether a given input should be excluded based on the given InputExclusions,
@@ -120,7 +129,7 @@ func preprocessSymlink(execRoot, symlink string, meta *filemetadata.SymlinkMetad
 // recursively), and loads their contents into the provided map.
 func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileSysNode, chunkSize int, cache filemetadata.Cache, opts *TreeSymlinkOpts) error {
 	if opts == nil {
-		opts = &TreeSymlinkOpts{}
+		opts = NewTreeSymlinkOpts()
 	}
 	absPath := filepath.Clean(filepath.Join(execRoot, path))
 	normPath, err := getRelPath(execRoot, absPath)
@@ -163,7 +172,7 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 		fs[path] = &fileSysNode{
 			symlink: &symlinkNode{target: meta.Symlink.Target},
 		}
-		if meta.Symlink.IsDangling {
+		if meta.Symlink.IsDangling || !opts.FollowsTarget {
 			return nil
 		}
 		// If we can proceed here, we've already verified that target is
@@ -185,7 +194,7 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 		return nil
 	}
 	for _, f := range files {
-		if e := loadFiles(execRoot, excl, filepath.Join(normPath, f.Name()), fs, chunkSize, cache, config); e != nil {
+		if e := loadFiles(execRoot, excl, filepath.Join(normPath, f.Name()), fs, chunkSize, cache, opts); e != nil {
 			return e
 		}
 	}
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
@@ -68,8 +67,8 @@ type TreeSymlinkOpts struct {
 	FollowsTarget bool
 }
 
-// NewTreeSymlinkOpts returns a default NewTreeSymlinkOpts object.
-func NewTreeSymlinkOpts() *TreeSymlinkOpts {
+// DefaultTreeSymlinkOpts returns a default DefaultTreeSymlinkOpts object.
+func DefaultTreeSymlinkOpts() *TreeSymlinkOpts {
 	return &TreeSymlinkOpts{
 		FollowsTarget: true,
 	}
@@ -94,14 +93,7 @@ func getTargetRelPath(execRoot, symlink, target string) (string, error) {
 	if !filepath.IsAbs(target) {
 		target = filepath.Clean(filepath.Join(execRoot, target))
 	}
-	rel, err := filepath.Rel(execRoot, target)
-	if err != nil {
-		return "", err
-	}
-	if strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
-		return "", fmt.Errorf("symlink=%q (target=%q) is not under execRoot=%q", symlink, target, execRoot)
-	}
-	return rel, nil
+	return getRelPath(execRoot, target)
 }
 
 // preprocessSymlink returns two things: if the routine should continue, and if
@@ -129,7 +121,7 @@ func preprocessSymlink(execRoot, symlink string, meta *filemetadata.SymlinkMetad
 // recursively), and loads their contents into the provided map.
 func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileSysNode, chunkSize int, cache filemetadata.Cache, opts *TreeSymlinkOpts) error {
 	if opts == nil {
-		opts = NewTreeSymlinkOpts()
+		opts = DefaultTreeSymlinkOpts()
 	}
 	absPath := filepath.Clean(filepath.Join(execRoot, path))
 	normPath, err := getRelPath(execRoot, absPath)
@@ -169,17 +161,19 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 		}
 		return nil
 	} else if t == command.SymlinkInputType {
+		relTarget, err := getTargetRelPath(execRoot, path, meta.Symlink.Target)
+		if err != nil {
+			return err
+		}
 		fs[path] = &fileSysNode{
-			symlink: &symlinkNode{target: meta.Symlink.Target},
+			// We cannot directly use meta.Symlink.Target, because it could be
+			// an absolute path. Since the remote worker will map the exec root
+			// to a different directory, we must strip away the local exec root.
+			// See https://github.com/bazelbuild/remote-apis-sdks/pull/229#discussion_r524830458
+			symlink: &symlinkNode{target: relTarget},
 		}
 		if meta.Symlink.IsDangling || !opts.FollowsTarget {
 			return nil
-		}
-		// If we can proceed here, we've already verified that target is
-		// under execRoot, so this should never fail.
-		relTarget, err := getTargetRelPath(execRoot, path, meta.Symlink.Target)
-		if err != nil {
-			panic(fmt.Sprintf("Invalid target=%q err=%v", relTarget, err))
 		}
 		return loadFiles(execRoot, excl, relTarget, fs, chunkSize, cache, opts)
 	}
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -165,7 +165,7 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 		if err != nil {
 			return err
 		}
-		fs[path] = &fileSysNode{
+		fs[normPath] = &fileSysNode{
 			// We cannot directly use meta.Symlink.Target, because it could be
 			// an absolute path. Since the remote worker will map the exec root
 			// to a different directory, we must strip away the local exec root.
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -62,8 +62,7 @@ type TreeStats struct {
 type TreeSymlinkOpts struct {
 	// By default, a symlink is converted into its targeted file.
 	// If true, preserve the symlink.
-	Preserved            bool
-	AllowDanglingSymlink bool
+	Preserved bool
 }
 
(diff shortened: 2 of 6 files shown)


go/pkg/tree/tree.go
```diff
@@ -46,6 +46,16 @@ type Stats struct {
 	// TODO(olaola): number of FileMetadata cache hits/misses go here.
 }
 
+// ConstructTreeConfig controls the behavior when constructing a tree from the local filesystem.
+type ConstructTreeConfig struct {
```
atetubou, line 50: ConstructConfig? Tree is in package name, so redandant.

go/pkg/tree/tree.go
```diff
@@ -99,15 +109,15 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
...
+		if e := loadFiles(execRoot, excl, filepath.Join(path, f.Name()), fs, chunkSize, cache, config); e != nil {
 			return e
 		}
 	}
 	return nil
 }
 
 // ComputeMerkleTree packages an InputSpec into uploadable inputs, returned as Chunkers.
-func ComputeMerkleTree(execRoot string, is *command.InputSpec, chunkSize int, cache filemetadata.Cache) (root digest.Digest, inputs []*chunker.Chunker, stats *Stats, err error) {
+func ComputeMerkleTree(execRoot string, is *command.InputSpec, chunkSize int, cache filemetadata.Cache, config ConstructTreeConfig) (root digest.Digest, inputs []*chunker.Chunker, stats *Stats, err error) {
```
atetubou, line 120: Is it possible to pass each config as optional args by making this variadic function?

go/pkg/tree/tree.go
```diff
@@ -46,6 +46,16 @@ type Stats struct {
...
 }
 
+// ConstructTreeConfig controls the behavior when constructing a tree from the local filesystem.
+type ConstructTreeConfig struct {
+	// By default, symlinks are implicitly treated as a regular file, with the
+	// link metadata stripped away. Setting this to true causes the symlinks to
+	// be distinguished from the normal files.
+	DistinguishSymlinkFromFile bool
+	// If true, the dangling symlinks will not be treated as an error.
+	AllowDanglingSymlink bool
```
atetubou, line 56: I prefer to remove this and allow dangling symlink when symlink is enabled.

go/pkg/tree/tree.go
```diff
@@ -46,6 +46,16 @@ type Stats struct {
 	// TODO(olaola): number of FileMetadata cache hits/misses go here.
 }
 
+// ConstructTreeConfig controls the behavior when constructing a tree from the local filesystem.
+type ConstructTreeConfig struct {
+	// By default, symlinks are implicitly treated as a regular file, with the
+	// link metadata stripped away. Setting this to true causes the symlinks to
+	// be distinguished from the normal files.
+	DistinguishSymlinkFromFile bool
```
atetubou, line 54: Can we use just 'EnableSymlink' here?

go/pkg/tree/tree.go
k-ye: I don't think so, because the default client behavior is to treat this as an error, and we want backward compatibility?

go/pkg/tree/tree.go
k-ye: I believe that's going to restrict the config params to be of the same type?

If your concern is around the temporary break of `cas`, we can add another function, e.g. `ComputeMerkleTreeWithConfig`?

go/pkg/tree/tree.go
atetubou: Enabling symlink is new behavior, so I think there is no previous behavior we want to keep compatibility?

go/pkg/tree/tree.go
k-ye: Done.

go/pkg/tree/tree.go
k-ye: Done

go/pkg/tree/tree.go
k-ye: I think these two are independent (i.e. it's fine if one enables `AllowDanglingSymlink` but not `EnableSymlink`). There is a check to return error for dangling symlinks:

https://github.com/bazelbuild/remote-apis-sdks/blob/af8c2eea6772090ab71c0eda356df2f011893626/go/pkg/tree/tree.go#L68-L70

go/pkg/tree/tree.go
atetubou: I imagined to use functional option pattern, but `ComputeMerkleTreeWithConfig` is fine too.

go/pkg/tree/tree.go
atetubou: I see, but I still think it is better to make it small where user can configure. Adding `AllowDanglingSymlink` later is easy if we found it is necessary, but removing is hard.

go/pkg/tree/tree.go
k-ye: > I imagined to use functional option pattern

Ah, I guess you mean something like

```go
type ConstructOption = func(ConstrucConfig*)

func ComputeMerkleTree(execRoot string, is *command.InputSpec, chunkSize int, cache filemetadata.Cache,  opts ... ConstructOption) {
  config := &ConstrucConfig{}
  for o := range opts {
    o(config)
  }
  ...
}
```

That seems cleaner.

atetubou: yes.

go/pkg/tree/tree.go
k-ye: My understanding is that the SDK does not want to allow dangling symlinks by default, therefore this config. Let's see @ola-rozenfeld 's opinion?

go/pkg/tree/tree.go
k-ye: Done

go/pkg/tree/tree.go
k-ye: I've switched to the functional pattern. Meanwhile I also removed the config for the dangling symlink

atetubou:
Could you add code using the config as well, at least for upload part? I think it is better to have one PR for that.

go/pkg/tree/tree.go
k-ye: From https://github.com/bazelbuild/remote-apis-sdks/pull/188#issuecomment-702233019 I think the SDK team wants the handling of symlink and allowing dangling symlinks to be separate.

ola-rozenfeld: Yes, I'd prefer to have these as separate options, to preserve backward compatibility. Also, make sure that the default options are clear and initialized correctly (AllowDanglingSymlink=true).
Also, I'd actually prefer the introduction of the options and the implementation to be in the same PR, it's easier to review that way IMO.

ola-rozenfeld: Suggestion: now that tree is a part of Client after a recent PR, it would make a lot of sense to me to just make both of these Client Opts instead.

k-ye: Ack.

go/pkg/tree/tree.go
```diff
@@ -50,6 +50,36 @@ type Stats struct {
...
+type constructConfig struct {
+	enableSymlink        bool
+	allowDanglingSymlink bool
+}
+
+// ConstructOption controls the behavior when constructing a tree from the local filesystem.
+type ConstructOption = func(c *constructConfig)
+
+// EnableSymlink causes the symlinks to be distinguished from the normal files.
+func EnableSymlink() ConstructOption {
```
atetubou, line 62: better to accept bool arg?

go/pkg/tree/tree.go
```diff
@@ -50,6 +50,36 @@ type Stats struct {
...
+	return func(c *constructConfig) {
+		c.enableSymlink = true
+	}
+}
+
+// AllowDanglingSymlink allows dangling symlinks
+func AllowDanglingSymlink() ConstructOption {
```
atetubou, line 69: better to accept bool arg?

go/pkg/client/tree.go
```diff
@@ -89,6 +110,16 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
 			},
 		}
 		return nil
+	} else if t == command.SymlinkInputType {
+		fs[path] = &fileSysNode{
+			// TODO: How to handle a target with absolute path?
```
k-ye, line 115: [Here](https://github.com/bazelbuild/remote-apis/blob/aa8e718768c2973274c2a23e9dcc27b42afea794/build/bazel/remote/execution/v2/remote_execution.proto#L799-L801) says that an absolute path target can be supported. Do you think we still need to check if the target is within exec root or not?

go/pkg/tree/tree.go
k-ye: Ah, since the `tree` package is merged into `client`, I've refactored this part to have a `TreeSymlinkOpts`. PTAL

go/pkg/tree/tree.go
k-ye: ditto

go/pkg/client/tree.go
atetubou: I'm fine to keep symlink as-is here, as I don't remember we have the case handling in tree absolute path symlink.
I'm fine with 2 too.

go/pkg/client/tree.go
```diff
@@ -65,17 +81,22 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
...
 	t := command.FileInputType
-	if smd := meta.Symlink; smd != nil && smd.IsDangling {
-		return nil
-	}
-	if meta.Err != nil {
+	isSymlink := meta.Symlink != nil
+	isDanglingSymlink := isSymlink && meta.Symlink.IsDangling
+	if isDanglingSymlink {
+		if !(opts.Preserved && opts.AllowDanglingSymlink) {
+			return nil
```
atetubou, line 92: symlink is ignored silently here?

k-ye: Yeah, i think the behavior is consistent with the old code:

* if we preserved symlink, and allow for dangling symlinks, then the produced tree will contain a `SymlinkNode`, which points to an invalid target.
* Otherwise, return `nil`, which is what the old code did.

atetubou: Then I'm wondering what value does `AllowDanglingSymlink` have. Also `IgnoreDanglingSymlink` sounds proper name?

k-ye: Yeah. Given how `loadFile` is implemented, if the user doesn't specify the target, a symlink effectively becomes "dangling" in the tree. I removed this option. It seems that neither the old code nor the new one needs to treat dangling symlinks as an error.

atetubou: update PR description then?

go/pkg/client/tree_test.go
```diff
@@ -904,10 +1024,12 @@ func TestComputeMerkleTree(t *testing.T) {
 
 			e, cleanup := fakes.NewTestEnv(t)
 			defer cleanup()
-
+			if tc.treeOpts != nil {
```
atetubou, line 1027: nit: it is better to do nil check in Apply function?

k-ye: done

atetubou: seems not pushed?

go/pkg/client/tree.go
k-ye: Done

go/pkg/client/tree_test.go
k-ye: https://github.com/bazelbuild/remote-apis-sdks/pull/229/commits/6bdd9894ad2c6daa16b7d743f4c67858a5469647 ?

go/pkg/client/tree.go
ola-rozenfeld: I like it, but for the record, I just realized (a tad too late), that we could merge the upcoming compression change and this change without breaking compatibility using your original idea. I want to ask you if you guys still prefer it, because we could revert #235 and do it:

If we introduce TreeConstructionOption similar to client.Opt (or as you did), then we can have our new CompressionCutoff as part of these options and add these options to ConstructMerkleTree without breaking compatibility and without the need to move the tree package into client.

The cons of that option: it will result in a bit of a duplicated code for us when creating the compressed Chunker; it necessitates more work now (reverting #235 and what you did here, again); the client Options will be broken into two packages/types with similar code and the caller will need to *translate* from one to another (because ultimately, with all these options, they are Client options set through flags).

The pros of this option:
- won't break compatibility
- it passes the minimal required thing (only the construction options, not the whole client), allowing us to reuse the tree functions in a server implementation more easily (e.g. fake server package)

I'm not sure what we should do. I'm okay with both. What do you guys think?

rubensf: > it will result in a bit of a duplicated code for us when creating the compressed Chunker

I think this will be minimal and worth the effort (a dozen lines of dupped at most?)

I don't like the options translation as much but I generally like having tree be independent of Client - albeit I can't really see independent use cases, they're pretty independent in concept. Moreover, I *do* like not breaking compatibility. But I'm also fine with leaving #235 as is.

go/pkg/client/tree_test.go
atetubou: Sorry, I missed that.

go/pkg/client/tree.go
atetubou: I already manged to update deps for the PR in our repo, so I don't need the revert of the PR.

atetubou (approved):

go/pkg/client/tree.go
k-ye: > I want to ask you if you guys still prefer it, because we could revert #235 and do it:

Like @atetubou has said, since we've already resolved the incompatibility issue, I think it's fine to go with this new code, unless there are other concerns here?

ola-rozenfeld: Great! Thank you, just checking!

go/pkg/client/tree.go
ola-rozenfeld: We can error out on absolute path symlinks for now, and support them later if needed. However, traversing them is necessary. The users specify an input spec such as foo/,bar/, and it's our responsibility to decide what exactly we should upload based on that. If foo/bla is a directory symlink to /baz, we need to traverse the target to load all the files there into the Merkle tree, otherwise the symlink will be dangling on the bot.
Note that relative path symlinks that escape the input root are an error as well.

go/pkg/client/client.go
```diff
@@ -85,7 +85,9 @@ type Client struct {
 	// UtilizeLocality is to specify whether client downloads files utilizing disk access locality.
 	UtilizeLocality UtilizeLocality
 	// UnifiedCASOps specifies whether the client uploads/downloads files in the background
-	UnifiedCASOps       UnifiedCASOps
+	UnifiedCASOps UnifiedCASOps
+	// TreeSymlinkOpts controls how symlinks are handled when constructing a tree.
+	TreeSymlinkOpts     TreeSymlinkOpts
```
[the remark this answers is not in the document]
k-ye: ah i see, done.

go/pkg/client/client.go
```diff
@@ -170,6 +172,12 @@ func (s UnifiedCASOps) Apply(c *Client) {
 	c.UnifiedCASOps = s
 }
 
+func (o *TreeSymlinkOpts) Apply(c *Client) {
+	if o != nil {
```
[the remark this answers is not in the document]
k-ye: done

go/pkg/client/tree.go
```diff
@@ -65,17 +80,27 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 
 // loadFiles reads all files specified by the given InputSpec (descending into subdirectories
 // recursively), and loads their contents into the provided map.
-func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileSysNode, chunkSize int, cache filemetadata.Cache) error {
+func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileSysNode, chunkSize int, cache filemetadata.Cache, opts TreeSymlinkOpts) error {
```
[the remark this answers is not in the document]
k-ye: done

k-ye (commented):
> We can error out on absolute path symlinks for now, and support them later if needed. However, traversing them is necessary...  (https://github.com/bazelbuild/remote-apis-sdks/pull/229#discussion_r521422354)

Ack, I've decided to handle this once and for all. I will check if `target` is under `execRoot`, regardless if it's an absolute path or not (see `getTargetRelPath()` and `preprocessSymlink()`): 

* Escapes `execRoot`: report this as an error.
* Inside `execRoot`: OK, and we will traversing the target recursively, *provided that* it is not dangling.

The check is a bit non-trivial, so I added a `tree_whitelist_test.go` just to verify `getTargetRelPath`.

go/pkg/client/tree.go
```diff
@@ -89,6 +155,20 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
...
+			symlink: &symlinkNode{target: meta.Symlink.Target},
+		}
+		if meta.Symlink.IsDangling {
+			return nil
+		}
+		// If we can proceed here, we've already verified that target is
+		// under execRoot, so this should never fail.
+		relTarget, err := getTargetRelPath(execRoot, path, meta.Symlink.Target)
+		if err != nil {
+			panic(fmt.Sprintf("Invalid target=%q err=%v", relTarget, err))
```
k-ye, line 169: I used `panic` here instead of returning `err`, because I think this is an internal invariant. If we ever reach here, it means there's a bug in the code.

go/pkg/client/tree.go
```diff
@@ -63,19 +79,69 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 	return false
 }
 
+// getTargetRelPath returns the part of target that is relative to execRoot,
+// iff. target is under execRoot. Otherwise it returns an error.
```
atetubou, line 83: nit: `.` after `iff` is intentional?

k-ye, line 83:
```suggestion
// iff target is under execRoot. Otherwise it returns an error.
```

k-ye: removed..

go/pkg/client/tree_test.go
```diff
@@ -44,6 +48,16 @@ func mustMarshal(p proto.Message) []byte {
 	return b
 }
 
+func prependExecRootPlaceholder(target string) string {
+	return fmt.Sprintf("%s/%s", execRootPlaceholder, target)
```
atetubou, line 52: use `filepath.Join` or `path.Join`?

go/pkg/client/tree_whitebox_test.go
```diff
@@ -0,0 +1,52 @@
+package client
```
atetubou, line 1: maybe `client_test`?

go/pkg/client/tree_test.go
k-ye: This function is used in `SymlinkNode`, where the API says the target path is always separated by `/`: https://github.com/bazelbuild/remote-apis/blob/aa8e718768c2973274c2a23e9dcc27b42afea794/build/bazel/remote/execution/v2/remote_execution.proto#L798-L807

That said, I noticed that the existing tests's `inputPath` are also constructed with `/`, instead of `filepath.Join`. I wonder why this is not a problem on Windows..

go/pkg/client/tree_whitebox_test.go
k-ye: I cannot, because I'm trying to test an exported function (hence `whitebox` in the filename). Otherwise I would just place this whole thing inside `tree_test.go`.

I do see another way of doing this in https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd. Do you know which is preferred?

go/pkg/client/tree_test.go
atetubou: I think this repo doesn't have Windows coverage.

go/pkg/client/tree_whitebox_test.go
atetubou: Ah, I see.

go/pkg/client/tree_test.go
k-ye: hmm, i thought `cas` already worked on Win?

go/pkg/client/tree_whitebox_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+import "testing"
+
+func TestGetTargetRelPath(t *testing.T) {
+	execRoot := "/execRoot/dir"
+	symlink := "sym"
+	tests := []struct {
+		desc      string
+		target    string
+		isError   bool
```
atetubou, line 11: nit: I think `wantErr` is more common in Go.

go/pkg/client/tree.go
```diff
@@ -89,6 +155,20 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
...
+		if meta.Symlink.IsDangling {
+			return nil
+		}
+		// If we can proceed here, we've already verified that target is
+		// under execRoot, so this should never fail.
+		relTarget, err := getTargetRelPath(execRoot, path, meta.Symlink.Target)
+		if err != nil {
+			panic(fmt.Sprintf("Invalid target=%q err=%v", relTarget, err))
+		}
+		return loadFiles(execRoot, excl, relTarget, fs, chunkSize, cache, opts)
```
atetubou, line 171: What happen if symlink target is already in other upload paths?

k-ye: It would result in a repeated read, which probably isn't a big deal? I.e. it will be no different from the case where we specify a normal file path, and a directory that contains that file.

go/pkg/client/tree_whitebox_test.go
k-ye: i see, done

go/pkg/client/tree.go
atetubou: I see, but isolate/isolated don't upload target of symlink if it is not included in upload paths.
So I think it is fine to make whether upload target of symlink or not user's responsibility.

k-ye: > So I think it is fine to make whether upload target of symlink or not user's responsibility.

See https://github.com/bazelbuild/remote-apis-sdks/pull/229#discussion_r521422354? I added a `FollowsTarget` option. The SDK sets this to `true` by default. FYI @ola-rozenfeld

go/pkg/client/tree.go
ola-rozenfeld: We follow the guildeline of "don't panic" in Go code. In particular, it makes sense here because this is a library, and never want to crash the calling executable. If we have a bug, we should surface an error message that will make it easy to catch and fix it.

go/pkg/client/tree.go
ola-rozenfeld: Yep, makes sense, thank you!

go/pkg/client/tree_test.go
ola-rozenfeld: This repo works on Windows -- there are some tests that are Windows or Linux specific (see e.g. cache_posix_test.go https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/filemetadata/BUILD.bazel), but this test should work on both.

go/pkg/client/tree.go
```diff
@@ -89,6 +164,20 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
 			},
 		}
 		return nil
+	} else if t == command.SymlinkInputType {
+		fs[path] = &fileSysNode{
+			symlink: &symlinkNode{target: meta.Symlink.Target},
```
ola-rozenfeld, line 173: I think there is a problem here -- if the symlink target is absolute (and we have Preserve=true), then even when it is actually under the exec root we cannot preserve it as is, because that would fail on the bot (because the bot will map the exec root to a different absolute path).
In this edge case, I think that we will need to convert the symlink target to a relative path instead.

Alternatively, we could disallow absolute symlink targets in general when Preserve=true.

go/pkg/client/tree_test.go
```diff
@@ -490,6 +505,92 @@ func TestComputeMerkleTree(t *testing.T) {
...
+			additionalBlobs: [][]byte{fooBlob, fooDirBlob},
+			wantCacheCalls: map[string]int{
+				"fooDir/foo": 1,
+				"fooSym":     1,
+			},
+			wantStats: &client.TreeStats{
+				InputDirectories: 2,
+				InputFiles:       1,
+				InputSymlinks:    1,
+				TotalInputBytes:  fooDg.Size + fooDirDg.Size,
```
ola-rozenfeld, line 531: This seems low... maybe not counting the symlink nodes? Ditto below.

k-ye:
rebased

go/pkg/client/tree.go
ola-rozenfeld: Also, I suspect that when you fix this, you will no longer need the exec root placeholder workaround in tests.
But if you do want to test absolute symlinks, I think the IsAbsolute attribute of inputPath already does what you want, you don't need the exec root placeholder.

go/pkg/client/tree.go
```diff
@@ -45,11 +52,29 @@ type TreeStats struct {
...
+type TreeSymlinkOpts struct {
+	// By default, a symlink is converted into its targeted file.
+	// If true, preserve the symlink.
+	Preserved bool
+	// If true, the symlink target (if not dangling) is followed.
+	FollowsTarget bool
+}
+
+// NewTreeSymlinkOpts returns a default NewTreeSymlinkOpts object.
+func NewTreeSymlinkOpts() *TreeSymlinkOpts {
```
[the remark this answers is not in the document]
k-ye: done

go/pkg/client/tree.go
```diff
@@ -63,19 +88,69 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 	return false
 }
 
+// getTargetRelPath returns the part of target that is relative to execRoot,
```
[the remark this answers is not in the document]
k-ye: Ah yeah, but let me keep `getTargetRelPath()`, because it has a slightly different interface and is used in two places..

go/pkg/client/tree.go
k-ye: Thanks for pointing this out. Yeah the test is simplified..

> I think the IsAbsolute attribute of inputPath already does what you want, you don't need the exec root placeholder.

Note that `IsAbsolute` only affects how the test data  are constructed in the filesystem, but doesn't apply to the `SymlinkNode` protos. Anyway, the placeholder is removed now.

go/pkg/client/tree_test.go
k-ye: Hmm, you mean the `TotalInputBytes` should count the bytes for symlinks? I didn't do that because 1) the symlink node itself isn't associated with a `Digest`, and 2) will counting in both the symlink and the target result in double counting?

go/pkg/client/tree_test.go
```diff
@@ -570,6 +687,72 @@ func TestComputeMerkleTree(t *testing.T) {
 				TotalInputBytes:  fooDg.Size + fooDirDg.Size + barDg.Size + barDirDg.Size,
 			},
 		},
+		{
+			desc: "Directory absoluate symlink (preserved)",
```
ola-rozenfeld, line 691: typo: absolute

go/pkg/client/tree.go
```diff
@@ -93,6 +160,22 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
 			},
 		}
 		return nil
+	} else if t == command.SymlinkInputType {
+		relTarget, err := getTargetRelPath(execRoot, path, meta.Symlink.Target)
+		if err != nil {
+			return err
+		}
+		fs[path] = &fileSysNode{
```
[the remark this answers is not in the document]
k-ye: Done

go/pkg/client/tree_test.go
k-ye: Done

go/pkg/client/tree.go
```diff
@@ -93,6 +160,22 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
...
+			// We cannot directly use meta.Symlink.Target, because it could be
+			// an absolute path. Since the remote worker will map the exec root
+			// to a different directory, we must strip away the local exec root.
+			// See https://github.com/bazelbuild/remote-apis-sdks/pull/229#discussion_r524830458
+			symlink: &symlinkNode{target: relTarget},
+		}
+		if meta.Symlink.IsDangling || !opts.FollowsTarget {
+			return nil
```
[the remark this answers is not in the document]
k-ye: hmm, i don't have a Win at this moment to test this. But i think the test case has been creating dangling symlinks for a while. Have you noticed that it has failed on Win? https://github.com/bazelbuild/remote-apis-sdks/blob/47f78418f148770fb07cce6e82a02ccaf104509a/go/pkg/client/tree_test.go#L65-L67

go/pkg/client/tree.go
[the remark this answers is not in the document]
k-ye: Removed

go/pkg/client/tree.go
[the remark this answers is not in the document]
ola-rozenfeld: Oh, you're right -- I guess that does work and I misremembered. Thank you!

ola-rozenfeld (approved):

PR #231, opened 2020-11-06
Refactor where IsDirectory is stored in the file metadata

k-ye:
This is a prep step to address https://github.com/bazelbuild/remote-apis-sdks/pull/229#issuecomment-722923295. I find the code to handle symlinks a bit too complicated when `IsDirectory` is put inside the `FileError`, especially when symlink and directory are both true.


atetubou (approved):
looks good to me

gkousik (approved):

PR #232, opened 2020-11-06
Add compressed writes to cas.go.

rubensf:
This follows the current tentative API being worked on in
bazelbuild/remote-apis#168. While there's technically room for it to
change, it has reached a somewhat stable point worth implementing.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -25,6 +25,9 @@ import (
 	log "github.com/golang/glog"
 )
 
+// DefaultCompressedBytestreamThreshold is the default threshold for transferring blobs compressed on ByteStream.Write RPCs.
+const DefaultCompressedBytestreamThreshold = 1024
+
 // UploadIfMissing stores a number of uploadable items.
 // It first queries the CAS to see which items are missing and only uploads those that are.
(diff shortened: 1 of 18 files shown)


go/pkg/client/cas.go
```diff
@@ -135,7 +142,26 @@ func (c *Client) WriteProto(ctx context.Context, msg proto.Message) (digest.Dige
 func (c *Client) WriteBlob(ctx context.Context, blob []byte) (digest.Digest, error) {
 	ch := chunker.NewFromBlob(blob, int(c.ChunkMaxSize))
 	dg := ch.Digest()
-	return dg, c.WriteChunked(ctx, c.ResourceNameWrite(dg.Hash, dg.Size), ch)
+
```
ola-rozenfeld, line 145: That is not enough to change if we want all writes to be compressed. Things don't usually go through this intermediate-layer function. Most use-cases use UploadIfMissing. I'd suggest creating a Client wrapper function for creating a Chunker from file (which initializes the correct type of chunker based on client parameters) and use it everywhere that Chunkers are created.

rubensf:
Closing in favor of #240

PR #233, opened 2020-11-07
tree: pull EmptyDirectoryMarker out of fileNode

k-ye:
Another cleanup for #229 .

Also lower-cased the member names in `fileNode`, `treeNode`.

### DIFF
--- a/go/pkg/tree/tree.go
+++ b/go/pkg/tree/tree.go
@@ -27,8 +27,12 @@ type treeNode struct {
 }
 
 type fileNode struct {
-	Chunker              *chunker.Chunker
-	IsExecutable         bool
+	Chunker      *chunker.Chunker
+	IsExecutable bool
+}
+
+type fileSysNode struct {
+	file                 *fileNode
 	EmptyDirectoryMarker bool
 }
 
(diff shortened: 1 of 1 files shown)


go/pkg/tree/tree.go
```diff
@@ -27,8 +27,12 @@ type treeNode struct {
...
 
 type fileNode struct {
-	Chunker              *chunker.Chunker
-	IsExecutable         bool
+	Chunker      *chunker.Chunker
+	IsExecutable bool
+}
+
+type fileSysNode struct {
+	file                 *fileNode
```
atetubou, line 35: `File`?

go/pkg/tree/tree.go
```diff
@@ -27,8 +27,12 @@ type treeNode struct {
 }
 
 type fileNode struct {
-	Chunker              *chunker.Chunker
-	IsExecutable         bool
+	Chunker      *chunker.Chunker
+	IsExecutable bool
+}
+
+type fileSysNode struct {
```
atetubou, line 34: why do you need to have this separately?

k-ye: I'll later add `symlinkNode` into it. `fileSysNode` is a union representing a regular file, and empty directory or a symlink. (It's really similar to `treeNode`, except that it only contains a single item)

go/pkg/tree/tree.go
k-ye: Done (but i feel like it should be the other way around, since these are all private types)

go/pkg/tree/tree.go
atetubou: I see.

go/pkg/tree/tree.go
atetubou: Yeah, I agree with you.

atetubou (approved):

go/pkg/tree/tree.go
k-ye: OK, I took it upon myself to change all these to lower case. @ola-rozenfeld  @gkousik

ola-rozenfeld (approved):

PR #235, opened 2020-11-10
Move Tree into Client package.

rubensf:
Besides tree_test_, all other uses of tree.go already have a client -
it's related enough to justify the move.

The motivation here is so that tree can prepare chunks that are aware
whether or not they should be compressed.


ola-rozenfeld (approved):

atetubou:
Could you keep interface compatiblity as long as possible (e.g. separate new interface addition and old code removal)?
Our deps is not managed in a repository and if you don't keep interface compatibility, that makes updating deps difficult for us.

PR #238, opened 2020-11-12
Separate "Chunker" and "UploadEntry".

rubensf:
This intentionally breaks the dependency that Tree has to know before
hand information about how to separate data into chunks for uploading,
when chunking is just an artifact of how Bytestream operates on data.

I could technically have left the chunker constructors and chunkSize
in `tree.go` calls, but they'd end up being misleading, so I removed
them albeit the higher breaking cost.


go/pkg/chunker/chunker.go
```diff
@@ -34,69 +69,66 @@ type Chunker struct {
...
+	size := fmt.Sprintf("<%d bytes>", c.ue.digest.Size)
 	if c.path == "" {
 		return size
 	}
 	return fmt.Sprintf("%s: %s", size, c.path)
 }
 
-// Digest returns the digest of the full data of this chunker.
-func (c *Chunker) Digest() digest.Digest {
-	return c.digest
```
ola-rozenfeld, line 97: We could return ue.digest -- I assume you remove this because it is no longer used?

go/pkg/chunker/chunker.go
```diff
@@ -34,69 +69,66 @@ type Chunker struct {
...
 	reachedEOF bool
-	path       string
+
+	ue         *UploadEntry
+	compressed bool
 }
 
-// NewFromBlob initializes a Chunker from the provided bytes buffer.
-func NewFromBlob(blob []byte, chunkSize int) *Chunker {
+func NewFromUEntry(ue *UploadEntry, compressed bool, chunkSize int) (*Chunker, error) {
```
ola-rozenfeld, line 80: nit: If this is our only constructor, we can just call it New.

ola-rozenfeld (commented):
LGTM, just a couple of nits! Thank you!

go/pkg/chunker/chunker.go
```diff
@@ -21,6 +21,42 @@ var IOBufferSize = 10 * 1024 * 1024
 // ErrEOF is returned when Next is called when HasNext is false.
 var ErrEOF = errors.New("ErrEOF")
 
+type UploadEntry struct {
```
ola-rozenfeld, line 24: Can you please separate this into a different package? My reasoning is: the chunker is something very particular that only a few things depend upon (only the actual upload code), but many more things need to depend on UploadEntry (rexec, for instance). So it's separation of concerns.

go/pkg/chunker/chunker.go
```diff
@@ -31,72 +67,69 @@ type Chunker struct {
...
-// any time. This means that late errors due to mismatch of digest and
-// contents are possible.
-func NewFromFile(path string, dg digest.Digest, chunkSize int) *Chunker {
-	if chunkSize < 1 {
-		chunkSize = DefaultChunkSize
-	}
-	if chunkSize > IOBufferSize {
-		chunkSize = IOBufferSize
-	}
+func chunkerFromBlob(blob []byte, compressed bool) *Chunker {
```
ola-rozenfeld, line 108: Are these functions reused? If not, please inline. Ditto below.

go/pkg/chunker/chunker.go
```diff
@@ -31,72 +67,69 @@ type Chunker struct {
...
 		chunkSize = DefaultChunkSize
 	}
-	contents := make([]byte, len(blob))
-	copy(contents, blob)
-	return &Chunker{
-		contents:  contents,
-		chunkSize: chunkSize,
-		digest:    digest.NewFromBlob(blob),
+	var c *Chunker
+	if ue.contents != nil {
```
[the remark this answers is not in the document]
rubensf: Done

go/pkg/chunker/chunker.go
```diff
@@ -31,70 +30,53 @@ type Chunker struct {
...
-// NewFromProto initializes a Chunker from the marshalled proto message.
-func NewFromProto(msg proto.Message, chunkSize int) (*Chunker, error) {
-	blob, err := proto.Marshal(msg)
-	if err != nil {
-		return nil, err
+		if chunkSize > IOBufferSize {
+			chunkSize = IOBufferSize
+		}
+	} else {
+		return nil, errors.New("Invalid UEntry. Content and path cannot both be nil.")
```
ola-rozenfeld, line 65: Update the error message.

go/pkg/fakes/cas.go
```diff
@@ -598,7 +599,12 @@ func (f *CAS) Read(req *bspb.ReadRequest, stream bsgrpc.ByteStream_ReadServer) e
 		return status.Errorf(codes.NotFound, "test fake missing blob with digest %s was requested", dg)
 	}
 
-	ch := chunker.NewFromBlob(blob, 2*1024*1024)
+	f.maybeBlock(dg)
```
ola-rozenfeld, line 602: Why are we blocking twice?

ola-rozenfeld (approved):

PR #239, opened 2020-11-13
Add compressed file reader.

rubensf:
This is similar, but replaces #230. It's built on top of #238. I've made it a separate PR in case we don't go with #238, or #238 changes substantially.

### DIFF
--- a/go/pkg/chunker/chunker.go
+++ b/go/pkg/chunker/chunker.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/reader"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
+	"github.com/klauspost/compress/zstd"
 )
 
 // DefaultChunkSize is the default chunk size for ByteStream.Write RPCs.
@@ -19,6 +20,9 @@ var IOBufferSize = 10 * 1024 * 1024
 
 // ErrEOF is returned when Next is called when HasNext is false.
 var ErrEOF = errors.New("ErrEOF")
+
+// Compressor for full blobs
+var fullCompressor, _ = zstd.NewWriter(nil)
 
 // Chunker can be used to chunk an input into uploadable-size byte slices.
 // A single Chunker is NOT thread-safe; it should be used by a single uploader thread.
@@ -94,6 +98,7 @@ func (c *Chunker) ChunkSize() int {
 // TODO(olaola): implement Seek(offset) when we have resumable uploads.
 func (c *Chunker) Reset() {
 	if c.r != nil {
+		// We're ignoring the error here, as not to change the fn signature.
 		c.r.SeekOffset(0)
 	}
 	c.offset = 0
--- a/go/pkg/reader/reader.go
+++ b/go/pkg/reader/reader.go
@@ -2,10 +2,13 @@ package reader
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os"
+
+	"github.com/klauspost/compress/zstd"
 )
 
 type Initializable interface {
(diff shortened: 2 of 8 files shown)


go/pkg/reader/reader.go
```diff
@@ -95,3 +99,69 @@ func (fio *fileSeeker) Initialize() error {
 	fio.initialized = true
 	return nil
 }
+
+type compressedSeeker struct {
+	fs   ReadSeeker
+	encd *zstd.Encoder
+	// This keeps the encrypted data
```
mostynb, line 106: I guess you mean "compressed data" here instead?

rubensf: Done.

go/pkg/chunker/chunker.go
```diff
@@ -21,6 +22,9 @@ var IOBufferSize = 10 * 1024 * 1024
 // ErrEOF is returned when Next is called when HasNext is false.
 var ErrEOF = errors.New("ErrEOF")
 
+// Compressor for full blobs
+var fullCompressor, _ = zstd.NewWriter(nil)
```
ola-rozenfeld, line 26: Is this one thread safe?

go/pkg/reader/reader.go
```diff
@@ -95,3 +99,69 @@ func (fio *fileSeeker) Initialize() error {
...
+	if retErr == nil {
+		retErr = err
+	}
+	return m, err
+}
+
+func (cfs *compressedSeeker) SeekOffset(offset int64) error {
+	cfs.buf.Reset()
+	err := cfs.encd.Close()
+	cfs.encd.Reset(cfs.buf)
```
ola-rozenfeld, line 157: Why not return err immediately if it is not nil?

go/pkg/chunker/chunker.go
mostynb: `Encoder.EncodeAll` is thread-safe, and that's all that is used below.

rubensf: @mostynb is right here - I expanded the comments :)

go/pkg/reader/reader.go
rubensf: Hah, I had a pending comment reply on #230. Basically it'd allow you to re-use a reader - but that's not what we are planning to do anymore.

go/pkg/reader/reader.go
```diff
@@ -107,3 +111,75 @@ func (fio *fileSeeker) Initialize() error {
...
+		fs:   fs,
+		encd: encd,
+		buf:  buf,
+	}, err
+}
+
+func (cfs *compressedSeeker) Read(p []byte) (int, error) {
+	var n int
+	var errR, errW error
+	for cfs.buf.Len() < len(p) && errR == nil && errW == nil {
```
ola-rozenfeld, line 143: Re: errors: I meant something like:

```
defer cfs.encd.Close()
for cfs.buf.Len() < len(p) {
  n, err := cfs.fs.Read(p)
  if err != nil {
    return 0, err
  }
  if _, err = cfs.encd.Write(p[:n]); err != nil {
    return 0, err
  }
}
return cfs.buf.Read(p)
```

Wouldn't that be equivalent? Or am I missing something?

ola-rozenfeld: And if you really need to close the encoder conditionally on no previous errors, it should be doable with:

```
func (cfs *compressedSeeker) Read(p []byte) (n int, err error) {
  defer func() {
    if err != nil {
      cfs.encd.Close()
     }
  }()
...
}
```

rubensf: The first block is different whereas in I'm still doing the Write after the loop Read, regardless if the loop Read returned an error. Basically I'm making sure that I'm compressing *all* the data I have available. Besides, I like the code having fewer indentation.

Additionally, I want to Close the enconder on an *existing* error (ie EOF or any other error that'd make continuing the compression impossible). It would be better to use the defer if I was preemptively returning errors in your first suggestion - otherwise it merely adds indentation.

ola-rozenfeld (approved):

PR #243, opened 2020-11-18
Closing the reader in the end of a read.

ola-rozenfeld:
Bugfix: this should fix the broken Windows unit test.

### DIFF
--- a/go/pkg/reader/reader.go
+++ b/go/pkg/reader/reader.go
@@ -15,6 +15,7 @@ type Initializable interface {
 
 type ReadSeeker interface {
 	io.Reader
+	io.Closer
 	Initializable
 	SeekOffset(offset int64)
 }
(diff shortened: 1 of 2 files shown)


go/pkg/chunker/chunker.go
```diff
@@ -134,6 +134,7 @@ func (c *Chunker) FullData() ([]byte, error) {
 	}
 	// Cache contents so that the next call to FullData() doesn't result in file read.
 	c.contents, err = ioutil.ReadAll(c.r)
+	c.r.Close()
```
gkousik, line 137: Should this be `defer c.r.Close()` at the beginning of this function?

ola-rozenfeld: I don't think so. I'm only closing when I do a full read (and reach EOF).

ola-rozenfeld:
Removed the Chunker change -- will do that in a separate CL, and fix unit tests first.

gkousik (approved):

PR #245, opened 2020-11-20
Fix the way to compute the target path relative to the symlink

k-ye:
I believe what I did in #229 was wrong when handling the symlink target.

1. When we construct the `SymlinkNode`, `Target` should be relative to **the directory of the symlink**. #229 computed the path relative to `execRoot` instead. This is problematic when the target is an absolute path in the filesystem.
2. When we recursively call `loadFiles()` on the target, this time `path` should still be the part of target relative to `execRoot`.

### DIFF
--- a/go/pkg/client/tree_test.go
+++ b/go/pkg/client/tree_test.go
@@ -57,6 +57,15 @@ type inputPath struct {
 func construct(dir string, ips []*inputPath) error {
 	for _, ip := range ips {
 		path := filepath.Join(dir, ip.path)
+		if ip.emptyDir {
+			if err := os.MkdirAll(path, 0777); err != nil {
+				return err
+			}
+			continue
+		}
+		if err := os.MkdirAll(filepath.Dir(path), 0777); err != nil {
+			return err
+		}
 		if ip.isSymlink {
 			target := ip.symlinkTarget
 			if ip.isAbsolute {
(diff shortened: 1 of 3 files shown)


ola-rozenfeld (approved):
Right, good catch! I'm surprised no tests in tree_test were changed. In general, I'd much rather test the public function there than the small private function.

k-ye:
https://github.com/bazelbuild/remote-apis-sdks/pull/245#pullrequestreview-535059321

One sec, let me add some test..

k-ye:
> One sec, let me add some test..

Done

go/pkg/client/tree_test.go
```diff
@@ -692,28 +697,27 @@ func TestComputeMerkleTree(t *testing.T) {
 			input: []*inputPath{
 				{path: "foobarDir/foo", fileContents: fooBlob, isExecutable: true},
 				{path: "foobarDir/bar", fileContents: barBlob},
-				{path: "foobarSymDir", isSymlink: true, isAbsolute: true, symlinkTarget: "foobarDir"},
+				{path: "base/foobarSymDir", isSymlink: true, isAbsolute: true, symlinkTarget: "foobarDir"},
```
k-ye, line 700: This might seem like it should be `../foobarDir`, but because `isAbsolute` is `true`, the actual target it generates is `"${execRoot}/foobarDir"`.

ola-rozenfeld (approved):

PR #246, opened 2020-11-24
Make CI run multiple tests.

rubensf:
While this has a toil on CI performance, this project is still small
enough that this is cost is easily affordable.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -11,7 +11,7 @@ tasks:
       - "//..."
     test_flags:
       - "--features=race"
-      - "--runs_per_test=5"
+      - "--runs_per_test=10"
       - "--jobs=100"
     test_targets:
       - "//..."
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -9,6 +9,10 @@ tasks:
       - "go vet ./..."
     build_targets:
       - "//..."
+    test_targets:
+      - "//..."
+# Check that this succeeds even in races.
+  ubuntu1804_race:
     test_flags:
       - "--features=race"
       - "--runs_per_test=10"
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -13,6 +13,8 @@ tasks:
       - "//..."
 # Check that this succeeds even in races.
   ubuntu1804_race:
+    platform: ubuntu1804
+    name: "bazel race check"
     test_flags:
       - "--features=race"
       - "--runs_per_test=10"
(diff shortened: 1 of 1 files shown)


ola-rozenfeld (commented):
Just curious, how long does it take with this change vs without?

rubensf:
~1min without these changes.
I tried with [runs_per_test=5](https://buildkite.com/bazel/remote-apis-sdks/builds/1292) and it goes to ~4.5 minutes, [=10](https://buildkite.com/bazel/remote-apis-sdks/builds/1290) to ~8-9minutes, but the 5 runs per test didn't reliably catch the concurrency issue we're experiencing.

In any case, we can run these tests locally much faster without checking for eg concurrency, so I don't believe the extra time to be problematic.

ifoox:
Would it be useful to set this up as a parallel target (like the gobuild one below) so that you could get fast results without the race detector, and then get results with it after 10 mins?

rubensf:
I'm not opposed to the idea if y'all think it useful, but I think the *real* faster way is just running `bazel test ...` or `go test ./...` which saves even more time, before pushing it to github. I think of these more as "sanity checks".

ola-rozenfeld:
I agree with Igor. I find the go formatter here to be more useful than in my client (it catches errors such as forgetting the right amount of parameters in fmt.Printf); also, some external developers might forget to run the tests, or run them via go and not bazel which might cause some differences. Also, the presubmits are for various platforms (or should be...).
So, overall, I often find that the presubmit helps a lot, and the faster it is (or parts of it), the better.

rubensf:
Sounds good - done.

ola-rozenfeld (approved):

gkousik:
Did the race test succeed locally for you? Is it only failing on buildkite?

ola-rozenfeld:
Oh, it's supposed to fail, until I fix my bug. This is WAI.

ola-rozenfeld (commented):
Suggestion: split them further with:

```
  # Deflake with multiple runs per test.
  ubuntu1804_deflake:
    platform: ubuntu1804
    name: "bazel deflake check"
    test_flags:
    - "--runs_per_test=100"
    test_targets:
    - "//..."
  # Check that gorace succeeds.
  ubuntu1804_race:
    platform: ubuntu1804
    name: "bazel race check"
    test_flags:
    - "--features=race"
    test_targets:
    - "//..."
```

This is what I used for testing #256. It currently doesn't work because TestCompressedReads fails racing, but it will be faster once #256 is in.

ola-rozenfeld:
I think if you rebase now and use my suggestion, it should pass now.

PR #247, opened 2020-11-25
Turn off compression by default.

rubensf:
While the RE API addition of the API isn't ready, we can't
check for server Capabilities whether compression is supported.


mostynb:
> While the RE API addition of the API isn't ready, we can't
> check for server Capabilities whether compression is supported.

I think you should be able to check if compression is supported using the Capabilities service already, if remote-apis-sdks uses the proto with the compressed-blobs PR applied. The reponses from servers which don't know about this field should be interpreted as "compression not supported".

rubensf:
We could do that, but I think it makes sense to do that for testing rather than having `master` rely on a non-merged-yet PR.

ola-rozenfeld (approved):

PR #250, opened 2020-11-26
Fix write compression.

rubensf:
On the midst of PR chains and rebasing, I managed to
leave the functionality off.

### DIFF
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -26,6 +26,7 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 
 // WriteChunked uploads chunked data with a given resource name to the CAS.
 func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chunker) error {
+	fmt.Println("name: ", name)
 	closure := func() error {
 		ch.Reset() // Retry by starting the stream from the beginning.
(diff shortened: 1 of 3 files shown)


go/pkg/client/bytestream.go
```diff
@@ -26,6 +26,7 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 
 // WriteChunked uploads chunked data with a given resource name to the CAS.
 func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chunker) error {
+	fmt.Println("name: ", name)
```
ola-rozenfeld, line 29: debug printf

ola-rozenfeld (commented):
But how did the tests pass, then?

go/pkg/client/bytestream.go
rubensf: Done - thanks for catching!

rubensf (commented):
The fake cas server is meant to reply with the appropriate blob type (compressed or uncompressed). The issue was that compressed writes were never being done.

Now the Write should be doing compressed writes when necessary, as confirmed by the debug print I just deleted :)

I arguably should expand the fake cas server to signal that it should expect a request to come compressed or uncompressed.

ola-rozenfeld (approved):

PR #253, opened 2020-11-30
Further write compression patch.

rubensf:
The existing code had a couple issues:
1) An EOF from the underlying uncompressed data would trigger finalizing
a read of the compressed data - even though the compressed data itself
wasn't ready or finalized reading.
2) *After* the `Close` call on the encoder, which triggers the
compressed data to be available for reading, is when you have to
actually read the final chunks and pieces of data.

The test code was also faulty as it checked the bytes up to the length
of the code compressed data (instead of the test compressed data), which
was 0.

This issue never arove on the fakes servers as cas_test doesn't seem to
have File tests - this compression issue wouldn't show up on in memory
blob tests. Just in case, I also added further checks on fake CAS to
ensure compressed calls were actually being made.

### DIFF
--- a/go/pkg/client/cas_test.go
+++ b/go/pkg/client/cas_test.go
@@ -54,8 +54,9 @@ func TestSplitEndpoints(t *testing.T) {
 	casServer := grpc.NewServer()
 	blob := []byte("foobar")
 	fake := &fakes.Reader{
-		Blob:   blob,
-		Chunks: []int{6},
+		Blob:             blob,
+		Chunks:           []int{6},
+		ExpectCompressed: false,
 	}
 	bsgrpc.RegisterByteStreamServer(casServer, fake)
 	go execServer.Serve(l1)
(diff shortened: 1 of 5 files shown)


go/pkg/client/cas_test.go
```diff
@@ -299,6 +302,12 @@ func TestWrite(t *testing.T) {
 	for _, cmp := range []client.CompressedBytestreamThreshold{-1, 0} {
 		for _, tc := range tests {
 			t.Run(fmt.Sprintf("%s - CompressionThresh:%d", tc.name, cmp), func(t *testing.T) {
+				if int(cmp) == 0 {
```
ola-rozenfeld, line 305: nit:
fake = int(cmp) == 0

go/pkg/client/cas_test.go
```diff
@@ -251,8 +252,10 @@ func TestRead(t *testing.T) {
 		if tc.limit == 0 {
 			// Limit tests don't work well with compression, as the limit refers to the compressed bytes
 			// while offset, per spec, refers to uncompressed bytes.
+			tc.fake.ExpectCompressed = true
```
ola-rozenfeld, line 255: nit: Another plug for refactoring the compression to be a test parameter. If you keep doing the setting/resetting a parameter, something is wrong. Separate cleanup CL is okay.

ola-rozenfeld (approved):

go/pkg/client/cas_test.go
rubensf: Thanks!

PR #254, opened 2020-12-02
Fixing a couple of flaky tests.

ola-rozenfeld:
Also improving the context logging by logging at depth 1 (which prints the caller function line number).

### DIFF
--- a/go/pkg/client/cas_test.go
+++ b/go/pkg/client/cas_test.go
@@ -581,6 +581,11 @@ func TestUploadConcurrentCancel(t *testing.T) {
 	for i := range blobs {
 		blobs[i] = []byte(fmt.Sprint(i))
 	}
+	var input []*uploadinfo.Entry
+	for _, blob := range blobs {
+		input = append(input, uploadinfo.EntryFromBlob(blob))
+	}
+	input = append(input, input...)
 	type testCase struct {
 		name string
 		// Whether to use batching.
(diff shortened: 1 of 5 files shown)


ola-rozenfeld:
Unfortunately, retries_test is still a little bit flaky, but I think this is also not due to a bug in the code, but rather strict test assumptions.

go/pkg/client/cas_test.go
```diff
@@ -616,18 +621,12 @@ func TestUploadConcurrentCancel(t *testing.T) {
 			fake.ReqSleepDuration = reqMaxSleepDuration
 			fake.ReqSleepRandomize = true
 			c := e.Client.GrpcClient
-			client.CASConcurrency(defaultCASConcurrency).Apply(c)
```
rubensf, line 619: It looks like this was completely removed-  is this intentional? I guess it is since I see that `tc.concurrency` is being applied right after.

go/pkg/client/retries_test.go
```diff
@@ -219,7 +219,7 @@ func setup(t *testing.T) *flakyFixture {
 	f.client, err = client.NewClient(f.ctx, instance, client.DialParams{
 		Service:    f.listener.Addr().String(),
 		NoSecurity: true,
-	}, client.StartupCapabilities(false), client.ChunkMaxSize(2), client.RPCTimeouts(map[string]time.Duration{"default": 50 * time.Millisecond}))
+	}, client.StartupCapabilities(false), client.ChunkMaxSize(2), client.RPCTimeouts(map[string]time.Duration{"default": 50 * time.Millisecond, "Execute": 0}))
```
rubensf, line 222: Execute always times out?

go/pkg/client/cas_test.go
ola-rozenfeld: Yep, idk why I had it here to begin with.

go/pkg/client/retries_test.go
ola-rozenfeld: Execute should basically never have a timeout. This should have still worked, and it (and other operations) are the only bug that still remains (will fail in ~1/200 runs). Sometimes a call will time out of the 50ms before it reaches my fake server, and I don't know why.

rubensf (approved):

PR #255, opened 2020-12-03
Splittied unified operations into many parameters.

ola-rozenfeld:


ramymedhat (approved):

gkousik (approved):
I suppose we can expose flags from the consumers of SDK?

ola-rozenfeld:
Yeah, like we already do. I thought to not add flags now because I will probably need to remove them later, but if we discover that there is a legitimate need to tune this for different scenarios and there isn't an optimal value, then exposing SDK flags makes sense too.

PR #256, opened 2020-12-04
Deflaking,improving performance,refactoring.

ola-rozenfeld:
Also improving retry logging. Includes adding race/deflakes to presubmit
-- decided to make that two separate jobs for efficiency -- gorace detects
different kinds of errors than runs_per_test; combining them is too
costly (currently, impossible because we spawn too many goroutines for
gorace to work).

### DIFF
--- a/go/pkg/client/retries_test.go
+++ b/go/pkg/client/retries_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"reflect"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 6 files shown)


ola-rozenfeld:
This now includes a bugfix in code as well (see client.go changes).

go/pkg/client/retries_test.go
```diff
@@ -33,6 +34,7 @@ type flakyServer struct {
 	mu               sync.RWMutex   // protects numCalls.
 	numCalls         map[string]int // A counter of calls the server encountered thus far, by method.
 	retriableForever bool           // Set to true to make the flaky server return a retriable error forever, rather than eventually a non-retriable error.
+	sleepDelay       time.Duration
```
rubensf, line 37: Add comments for these, since the others also have comments

rubensf (approved):

go/pkg/client/retries_test.go
ola-rozenfeld: Done.

PR #259, opened 2020-12-07
Add tool to run action multiple times to check for determinism.

ramymedhat:


go/pkg/tool/tool.go
```diff
@@ -31,54 +31,126 @@ const (
...
+		GrpcClient:        c.GrpcClient,
+	}
+	cmd, err := c.prepCommand(ctx, client, actionDigest, inputRoot)
+	if err != nil {
+		return err
+	}
+	opt := &command.ExecutionOptions{AcceptCached: false, DownloadOutputs: false, DownloadOutErr: true}
+	firstRes, firstMd := client.Run(ctx, cmd, opt, oe)
+	for i := 1; i < attempts; i++ {
+		testOnlyStartDeterminismExec()
```
ola-rozenfeld, line 58: Could we run them concurrently to save time?

ramymedhat: That's a good idea. Do you mind if I do this in a follow up PR? I'll file a bug for it.

ola-rozenfeld: Sure!

ola-rozenfeld (approved):

PR #260, opened 2020-12-07
Add digest verification to compressed reads.

rubensf:


go/pkg/client/cas.go
```diff
@@ -849,7 +876,11 @@ func (c *Client) readBlobStreamed(ctx context.Context, hash string, sizeBytes, o
 		return n, fmt.Errorf("Failed to finalize writing downloaded data downstream: %v", err)
 	}
 	close(done)
-	if wt.writtenBytes != sz {
+	// We can't reliably calculate hash without the full blob
```
mostynb, line 879: nitpick: remove "reliably" from this sentence

go/pkg/client/cas.go
```diff
@@ -812,24 +813,46 @@ func (c *Client) ReadBlobStreamed(ctx context.Context, d digest.Digest, w io.Wri
 }
 
 type writerTracker struct {
-	io.Writer
-	writtenBytes int64
+	w     io.Writer
+	pw    *io.PipeWriter
```
mostynb, line 817: I wonder if you could avoid the extra goroutine by using an io.MultiWriter instead of io.PipeWriter here?

ola-rozenfeld (approved):

go/pkg/client/cas.go
rubensf: Done.

go/pkg/client/cas.go
rubensf: I don't think I could avoid it using a io.MultiWriter the problem was how I set up the `digest.NewFromReader` to run synchronously. While I could change it to be able to stream it, creating a new struct in `digest.go` (to keep encoding code out of `cas.go`) & testing it is some work, and I think there are more important optimizations to do here, eg not close/open a new encoder for every blob.

PR #261, opened 2020-12-08
Add RealBytesUploaded to Metadata

rubensf:
Admittedly, the testing here is subpar. I'll add an AI to myself to add more thorough testing to account for eg only 1 client is counting the data and sanity check that the compressed bytes are a exact number expected.

### DIFF
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -15,19 +15,21 @@ import (
 )
 
 // WriteBytes uploads a byte slice.
-func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error {
+func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) (int64, error) {
 	ue := uploadinfo.EntryFromBlob(data)
 	ch, err := chunker.New(ue, false, int(c.ChunkMaxSize))
 	if err != nil {
-		return err
+		return 0, err
 	}
 	return c.WriteChunked(ctx, name, ch)
 }
 
 // WriteChunked uploads chunked data with a given resource name to the CAS.
-func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chunker) error {
+func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
+	var totalBytes int64
 	closure := func() error {
 		ch.Reset() // Retry by starting the stream from the beginning.
+		totalBytes = int64(0)
 		// TODO(olaola): implement resumable uploads.
 
 		stream, err := c.Write(ctx)
@@ -55,13 +57,15 @@ func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chun
 			if err != nil {
 				return err
 			}
+			totalBytes += int64(len(req.Data))
 		}
 		if _, err := stream.CloseAndRecv(); err != nil {
 			return err
 		}
 		return nil
 	}
-	return c.Retrier.Do(ctx, closure)
+	err := c.Retrier.Do(ctx, closure)
+	return totalBytes, err
 }
 
 // ReadBytes fetches a resource's contents into a byte slice.
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -15,13 +15,14 @@ import (
 )
 
 // WriteBytes uploads a byte slice.
-func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) (int64, error) {
+func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error {
 	ue := uploadinfo.EntryFromBlob(data)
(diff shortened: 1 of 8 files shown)


ola-rozenfeld:
I'll take a look after presubmits pass.

rubensf:
They were passing... Then I noticed a forgot a debug statement, deleted it, accidentally left uncompilable mumble jumble and forgot to check again :').

go/pkg/client/cas.go
```diff
@@ -142,16 +143,22 @@ func (c *Client) uploadProcessor() {
...
-			err:     err,
+			digest:     st.ue.Digest,
+			bytesMoved: bytesMoved,
+			missing:    missing,
+			err:        err,
 		}
+
+		// Only first client is reported to have moved the bytes to prevent double accounting.
+		bytesMoved = 0
+		// Only first client is reported the digest missing to prevent double accounting.
```
ola-rozenfeld, line 160: nit: unify these comments into one :-)

go/pkg/client/cas.go
```diff
@@ -360,6 +371,7 @@ func (c *Client) uploadNonUnified(ctx context.Context, data ...*uploadinfo.Entry
 						return err
 					}
 					bchMap[dg] = data
+					totalBytesTransferred += int64(len(data))
```
ola-rozenfeld, line 374: That should be protected somehow (different goroutines writing to same variable).
Please run test with features=race to verify.

go/pkg/client/cas.go
```diff
@@ -469,33 +486,33 @@ func (c *Client) UploadIfMissing(ctx context.Context, data ...*uploadinfo.Entry)
 // * Do we want to allow []byte uploads, or require the user to construct Chunkers?
 // * How to consistently distinguish in the API between should we use GetMissing or not?
 // * Should BatchWrite be a public method at all?
-func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte) error {
+func (c *Client) WriteBlobs(ctx context.Context, blobs map[digest.Digest][]byte) ([]digest.Digest, int64, error) {
```
ola-rozenfeld, line 489: We don't really need to change the interface of these middle-layer functions, right? We only use UploadIfMissing in the top layer. Ditto below.

go/pkg/client/cas.go
```diff
@@ -1505,7 +1522,7 @@ func (c *Client) shouldCompress(sizeBytes int64) bool {
 	return int64(c.CompressedBytestreamThreshold) >= 0 && int64(c.CompressedBytestreamThreshold) <= sizeBytes
 }
 
-func (c *Client) writeRscName(dg digest.Digest) string {
+func (c *Client) writeRscName(dg digest.Digest) (ret string) {
```
ola-rozenfeld, line 1525: Why this change?

go/pkg/client/bytestream.go
```diff
@@ -15,19 +15,21 @@ import (
 )
 
 // WriteBytes uploads a byte slice.
-func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error {
+func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) (int64, error) {
```
ola-rozenfeld, line 18: Actually, do we need to change this one? I don't think we use it in UloadIfMissing. Let's only change the things we have to change for UploadIfMissing.

go/pkg/client/cas.go
ola-rozenfeld: nit: using sync/atomic AddUint64 is shorter.

go/pkg/client/bytestream.go
```diff
@@ -15,19 +15,21 @@ import (
...
 	if err != nil {
-		return err
+		return 0, err
 	}
 	return c.WriteChunked(ctx, name, ch)
 }
 
 // WriteChunked uploads chunked data with a given resource name to the CAS.
-func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chunker) error {
+func (c *Client) WriteChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
```
ola-rozenfeld, line 28: Actually, now that I think about it, I'm really not sure this function needs to be public at all. It's very low-level. But I suggest a separate CL only to make this one private (so that it can easily be rolled back if it breaks someone external).
This way, we only change the signature of UploadIfMissing.

go/pkg/client/cas.go
rubensf: Done

go/pkg/client/bytestream.go
rubensf: I don't think we need to change this one, but this sounds like delaying until we inevitably want more metrics?
Done

go/pkg/client/bytestream.go
rubensf: Separate CL it is!

ola-rozenfeld (approved):

PR #262, opened 2020-12-09
rexec: Update to REAPI v2.2

santigl:
Versions 2.1 and 2.2 of the [Remote Execution API](https://github.com/bazelbuild/remote-apis/blob/1e9ccef3705c2c41a447aee0ddbe6810ccb7949d/build/bazel/remote/execution/v2/remote_execution.proto) introduce two changes:

* The Platform properties list can now be also set in the `Action` message (allowing servers to read the properties without dereferencing the `Command` message), and
* the `Action.output_{files, directories}` fields are now combined into `Action.output_paths`.

This bumps up the required version of `github.com/bazelbuild/remote-apis` to use v2.2 of `remote_execution.proto` and updates `exec.go` to populate those fields following the most recent version of the spec.

This is a breaking change for older implementations, but I wasn't sure what the preferred way of addressing this would be. I'll be happy to address this PR with the necessary adjustments before merging.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,9 +7,8 @@ require (
 	github.com/aclements/go-moremath v0.0.0-20190830160640-d16893ddf098 // indirect
 	github.com/ajstarks/deck v0.0.0-20191009173945-82d717002242 // indirect
 	github.com/ajstarks/svgo v0.0.0-20190826172357-de52242f3d65 // indirect
-	github.com/bazelbuild/remote-apis v0.0.0-20200904140912-1aeb39973178
+	github.com/bazelbuild/remote-apis v0.0.0-20201113154229-1e9ccef3705c
 	github.com/bradfitz/gomemcache v0.0.0-20190913173617-a41fca850d0b // indirect
-	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e // indirect
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
 	github.com/cznic/cc v0.0.0-20181122101902-d673e9b70d4d // indirect
 	github.com/cznic/fileutil v0.0.0-20181122101858-4d67cfea8c87 // indirect
@@ -29,6 +28,7 @@ require (
 	github.com/godbus/dbus v4.1.0+incompatible // indirect
 	github.com/golang/freetype v0.0.0-20170609003504-e2365dfdc4a0 // indirect
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
+	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/golang/protobuf v1.4.2
 	github.com/gonum/blas v0.0.0-20181208220705-f22b278b28ac // indirect
 	github.com/gonum/floats v0.0.0-20181209220543-c233463c7e82 // indirect
@@ -38,7 +38,6 @@ require (
 	github.com/google/go-cmp v0.5.1
 	github.com/google/uuid v1.1.2
 	github.com/googleapis/gax-go v2.0.2+incompatible // indirect
-	github.com/ianlancetaylor/demangle v0.0.0-20181102032728-5e5cf60278f6 // indirect
 	github.com/jung-kurt/gofpdf v1.13.0 // indirect
 	github.com/klauspost/compress v1.11.2
 	github.com/kylelemons/godebug v1.1.0
@@ -52,6 +51,7 @@ require (
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/syndtr/goleveldb v1.0.0 // indirect
 	golang.org/x/build v0.0.0-20191031202223-0706ea4fce0c // indirect
+	golang.org/x/exp v0.0.0-20200224162631-6cc2880d07d6 // indirect
 	golang.org/x/image v0.0.0-20191009234506-e7c1f5e7dbb8 // indirect
 	golang.org/x/mobile v0.0.0-20191031020345-0945064e013a // indirect
 	golang.org/x/net v0.0.0-20200904194848-62affa334b73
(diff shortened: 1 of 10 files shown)


go/pkg/client/exec.go
```diff
@@ -192,16 +194,13 @@ func buildCommand(ac *Action) *repb.Command {
...
+		OutputPaths: append(ac.OutputFiles, ac.OutputDirs...),
 		Platform: &repb.Platform{
 			Properties: []*repb.Platform_Property{{Name: containerImagePropertyName, Value: ac.DockerImage}},
 		},
 	}
-	copy(cmd.OutputFiles, ac.OutputFiles)
-	copy(cmd.OutputDirectories, ac.OutputDirs)
-	sort.Strings(cmd.OutputFiles)
-	sort.Strings(cmd.OutputDirectories)
+	sort.Strings(cmd.OutputPaths)
```
edbaunton, line 202: Do we also need to maintain the old fields for older servers and implementations?

ola-rozenfeld:
Unfortunately, RBE is actually not compatible with this API version. So we will have to be backwards-compatible in this change. Ideally, if we want to be the most efficient possible, we check the server API version using the Capabilities API (which we currently already do in client startup), and fill the correct fields based on that.

santigl:
That sounds good, @ola-rozenfeld.

I exported the `ServerCaps` from `Client` and added conditionals around the `HighApiVersion` value that the server reports to determine whether we should use the new fields.

(Apologies in advance if there is something in that code that is bad practice/style in Go.)

go/pkg/rexec/rexec.go
```diff
@@ -127,6 +127,32 @@ func (ec *Context) downloadOutputs(execRoot string) *command.Result {
 	return command.NewResultFromExitCode((int)(ec.resPb.ExitCode))
 }
 
+// HighAPIVersionNewerThanOrEqualTo returns whether the latest version reported
+// as supported in ServerCapabilities matches or is more recent than a
+// reference major/minor version.
+func HighAPIVersionNewerThanOrEqualTo(serverCapabilities *repb.ServerCapabilities, major int32, minor int32) bool {
```
ola-rozenfeld, line 133: Please make this private. Later, I'll move it out into the client, probably be something like client.serverSupports(apiVersion)->supportStatus

ola-rozenfeld:
Btw, I'll be out on vacation until Jan 5 -- if this is urgent, please ask @rubensf to finish this review.
Happy holidays!

go/pkg/rexec/rexec.go
santigl: Sounds good. That approach would be much nicer!

go.mod
```diff
@@ -7,9 +7,8 @@ require (
 	github.com/aclements/go-moremath v0.0.0-20190830160640-d16893ddf098 // indirect
 	github.com/ajstarks/deck v0.0.0-20191009173945-82d717002242 // indirect
 	github.com/ajstarks/svgo v0.0.0-20190826172357-de52242f3d65 // indirect
-	github.com/bazelbuild/remote-apis v0.0.0-20200904140912-1aeb39973178
+	github.com/bazelbuild/remote-apis v0.0.0-20201113154229-1e9ccef3705c
```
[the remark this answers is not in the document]
santigl: Changed it to bring this fresh fix: https://github.com/bazelbuild/remote-apis/commit/9e72daff42c941baaf43a4c370e2607a984c58a7.

santigl:
> Btw, I'll be out on vacation until Jan 5 -- if this is urgent, please ask @rubensf to finish this review.
> Happy holidays!

Thank you for the review, Ola. Happy holidays!

santigl:
@ola-rozenfeld, @rubensf: When you have some time, could I get another review of this?

go/pkg/rexec/rexec.go
```diff
@@ -127,6 +127,32 @@ func (ec *Context) downloadOutputs(execRoot string) *command.Result {
...
+		return false
+	}
+
+	latestSupportedMajor := serverCapabilities.HighApiVersion.GetMajor()
+	latestSupportedMinor := serverCapabilities.HighApiVersion.GetMinor()
+
+	return (latestSupportedMajor > major || (latestSupportedMajor == major && latestSupportedMinor >= minor))
+}
+
+func actionHasPlatformProperties(serverCapabilities *repb.ServerCapabilities) bool {
```
ola-rozenfeld, line 144: nit: I think all of this logic moved of the capabilities.go module, instead of exporting the ServerCaps. Then you can have client.ActionHasPublicProperties() and client.CommandHasOutputPaths() without arguments.

ola-rozenfeld:
This leaves me a little bit worried that once we add output path support (or other features) in RBE, this code will automatically invalidate all remote caches. But I'm not sure what can be done about this rather than carefully plan this on the RBE side. I mean, in theory we could also have a migration period and pre-populate the caches.
But in any case, I think your code makes sense and the migration is our problem. Just be advised that you might experience the same issue with your server as well (new command/action = 0 cache hits the first time).

go/pkg/rexec/rexec.go
santigl: I moved those methods to `Client` and figured they could use different names there: `Supports{ActionPlatformProperties, CommandOutputPaths}()`. What do you think?

santigl:
> This leaves me a little bit worried that once we add output path support (or other features) in RBE, this code will automatically invalidate all remote caches. But I'm not sure what can be done about this rather than carefully plan this on the RBE side. I mean, in theory we could also have a migration period and pre-populate the caches.
> But in any case, I think your code makes sense and the migration is our problem. Just be advised that you might experience the same issue with your server as well (new command/action = 0 cache hits the first time).

Right, the cache invalidation is unfortunate. Although for some servers being able to start inlining platform properties might offer a bit of an incentive for paying the price of those initial cache misses.

Thanks a lot, Ola!

ola-rozenfeld (approved):

PR #264, opened 2020-12-09
Add LogicalBytesUploaded

rubensf:


go/pkg/command/command.go
```diff
@@ -465,8 +465,11 @@ type Metadata struct {
 	OutputDigests map[string]digest.Digest
 	// Missing digests that are uploaded to CAS.
 	MissingDigests []digest.Digest
-	// RealBytesMoved is the real number of bytes put on the wire for upload. It may differ from the
-	// sum of MissingDigest.Sizes due to compression. It does not, however, account for blobs metadata.
+	// LogicalBytesMoved is the sum of sizes in bytes of the blobs that were uploaded. It should be
```
ola-rozenfeld, line 468: The names are *Uploaded, not moved. Comment doesn't match names.

ola-rozenfeld (approved):

go/pkg/command/command.go
rubensf: Done

PR #266, opened 2020-12-14
Add downloaded bytestream bytes metrics

rubensf:


go/pkg/client/cas.go
```diff
@@ -1245,17 +1268,27 @@ func (c *Client) downloadBatch(ctx context.Context, batch []digest.Digest, reqs
...
 	bchMap, err := c.BatchDownloadBlobs(ctx, batch)
 	if err != nil {
-		afterDownload(batch, reqs, err)
+		afterDownload(batch, reqs, map[digest.Digest]int64{}, err)
 		return
 	}
 	for _, dg := range batch {
 		data := bchMap[dg]
+		lbMoved := dg.Size
+		rbMoved := int64(len(data))
```
ola-rozenfeld, line 1277: This cannot ever be different from lbMoved here, right? This is BatchDownload, we don't compress these.

go/pkg/client/cas.go
```diff
@@ -1071,56 +1071,62 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
...
+	lbMoved, rbMoved, err := c.downloadOutputs(ctx, outputs, execRoot, cache)
+	return outputs, lbMoved, rbMoved, err
 }
 
-// DownloadActionOutputs downloads the output files and directories in the given action result.
-func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionResult, execRoot string, cache filemetadata.Cache) error {
+// DownloadActionOutputs downloads the output files and directories in the given action result. It returns the amount of downloaded bytes.
+// It returns the number of logical and real bytes downloaded, which may be different from sum
+// of sizes of the files due to dedupping and compression.
+func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionResult, execRoot string, cache filemetadata.Cache) (logicalBytesMoved, realBytesMoved int64, err error) {
```
ola-rozenfeld, line 1103: Suggestion: introduce a DownloadMetadata (or even better, a generalized MovedBytesMetadata or something) with the fields: BytesRequested, RealBytesMoved, LogicalBytesMoved and CachedBytes. There, BytesRequested=sum of all other fields (something is Cached if we don't need to move any bytes to fulfill the request, while the difference between Logical and Real is compression).
That way, you can refactor to just return this MovedBytesMetadata everywhere, and no further changes will be needed if we add/change fields (the functions API will no longer change).
It will also be more consistent across functions (here, sometimes you return one number, sometimes two, sometimes zero). I'd actually be okay with just returning this struct everywhere.
WDYT?

rubensf: I think "no further changes will be needed" is ambitious, but i'll give it a try.

rubensf:
@gkousik, mind taking this review from here?

go/pkg/client/cas.go
```diff
@@ -36,6 +36,14 @@ const DefaultCompressedBytestreamThreshold = -1
 // DefaultMaxHeaderSize is the defaut maximum gRPC header size.
 const DefaultMaxHeaderSize = 8 * 1024
 
+// MovedBytesMetadata represents the bytes moved in CAS related requests.
+type MovedBytesMetadata struct {
+	Requested    int64
+	LogicalMoved int64
+	RealMoved    int64
+	Cached       int64
```
gkousik, line 44: Do we really need Cached? We have two scenarios:
1. Unified uploader
2. Compression

When logicalmoved == realmoved, no compression, actually downloaded realmoved bytes
When logicalmoved > realmoved && realmoved > 0, compression, actually downloaded realmoved bytes
When logicalmoved > realmoved && realmoved == 0, didn't download anything (used unified uploader)

go/pkg/client/cas.go
```diff
@@ -1270,8 +1302,26 @@ func afterDownload(batch []digest.Digest, reqs map[digest.Digest][]*downloadRequ
...
+			RealMoved:    rbMoved,
+		}
+		for i, r := range rs {
+			// bytesMoved will be zero for error cases.
+			// We only report it to the first client to prevent double accounting.
+			r.wait <- &downloadResponse{stats: stats, err: err}
+			if i == 0 {
+				stats.Cached = stats.LogicalMoved
+				stats.RealMoved = 0
+				stats.LogicalMoved = 0
```
gkousik, line 1323: See the other comment about getting rid of "Cached" field. If you do that, then I think you may have to set LogicalMoved for the other clients as well.

go/pkg/client/cas.go
```diff
@@ -1299,7 +1364,8 @@ func (c *Client) downloadSingle(ctx context.Context, dg digest.Digest, reqs map[
 	// The lock is released when all file copies are finished.
 	// We cannot release the lock after each individual file copy, because
 	// the caller might move the file, and we don't have the contents in memory.
-	defer func() { afterDownload([]digest.Digest{dg}, reqs, err) }()
+	bytesMoved := map[digest.Digest]int64{}
+	defer func() { afterDownload([]digest.Digest{dg}, reqs, bytesMoved, err) }()
```
gkousik, line 1368: not related to this CL but why don't we just call

`defer afterDownload([]digest.Digest{dg}, reqs, bytesMoved, err)()` ?

go/pkg/client/cas.go
rubensf: Fair point - but let's use the other comment to decide on keeping/not keeping Cached :)

go/pkg/client/cas_test.go
```diff
@@ -250,22 +250,34 @@ func TestRead(t *testing.T) {
...
 			if tc.offset == 0 && tc.limit == 0 {
-				got, err := c.ReadBlob(ctx, digest.NewFromBlob(want))
+				got, bMoved, err := c.ReadBlob(ctx, digest.NewFromBlob(want))
 				if err != nil {
 					t.Errorf("c.ReadBlob(ctx, digest) gave error %s, want nil", err)
 				}
 				if !bytes.Equal(want, got) {
 					t.Errorf("c.ReadBlob(ctx, digest) gave diff: want %v, got %v", want, got)
 				}
+				if tc.compress && bMoved == int64(len(got)) && len(tc.fake.Blob) != 0 {
```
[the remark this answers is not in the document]
rubensf: It's deterministic given some parameters, eg compression speed used. IIUC, upgrading the underlying compression lib or just tweaking the settings we use for it could arguably change the compressed value to something that'd still be valid zstandard. So I think using concrete values to check against would be more trouble than it's worth. Wdyt?

go/pkg/client/cas.go
```diff
@@ -1412,8 +1480,11 @@ func (c *Client) download(data []*downloadRequest) {
 
 // This is a legacy function used only when UnifiedDownloads=false.
 // It will be removed when UnifiedDownloads=true is stable.
-func (c *Client) downloadNonUnified(ctx context.Context, execRoot string, outputs map[digest.Digest]*TreeOutput) error {
+// It returns the number of logical and real bytes downloaded, which may be different from sum
+// of sizes of the files due to dedupping and compression.
```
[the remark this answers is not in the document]
rubensf: Technically, no - it's done by `downloadOutputs` which is on the call path for this. But I guess we're describing this function, which is independent from potential callers :p. Removed.

go/pkg/client/cas.go
```diff
@@ -36,6 +36,33 @@ const DefaultCompressedBytestreamThreshold = -1
...
 const DefaultMaxHeaderSize = 8 * 1024
 
+// MovedBytesMetadata represents the bytes moved in CAS related requests.
+type MovedBytesMetadata struct {
+	// Requested is the sum of the sizes in bytes for all the uncompressed
+	// blobs needed by the execution. It includes bytes that might have
+	// been deduped and thus not passed through the wire.
+	Requested int64
+	// LogicalMoved is the sum of the sizes in bytes for all uncompressed
+	// blobs passed through the wire. It may be different
```
ola-rozenfeld, line 46: I think this sentence lost its ending?

go/pkg/client/cas.go
```diff
@@ -36,6 +36,33 @@ const DefaultCompressedBytestreamThreshold = -1
...
+	// RealMoved is the the sum of sizes in bytes for all blobs passed
+	// through the wire in the format they were passed through (eg
+	// compressed).
+	RealMoved int64
+	// Cached is amount of logical bytes that we did not have to move
+	// through the wire because they were de-duped.
+	Cached int64
+}
+
+func sumStats(a, b MovedBytesMetadata) MovedBytesMetadata {
```
ola-rozenfeld, line 57: nit: we agreed to pass all structs by pointer with the sole exception of Digest. Ditto everywhere else. Same for returning.

go/pkg/client/cas.go
```diff
@@ -940,12 +969,13 @@ func (c *Client) readBlobStreamed(ctx context.Context, hash string, sizeBytes, o
 }
 
 // ReadProto reads a blob from the CAS and unmarshals it into the given message.
-func (c *Client) ReadProto(ctx context.Context, d digest.Digest, msg proto.Message) error {
-	bytes, err := c.ReadBlob(ctx, d)
+// Returns the size of the proto and the amount of bytes moved through the wire.
+func (c *Client) ReadProto(ctx context.Context, d digest.Digest, msg proto.Message) (int64, int64, error) {
```
ola-rozenfeld, line 973: nit: maybe return the metadata everywhere instead of one of two ints? That way, we never have to change the API again.

go/pkg/client/cas.go
rubensf: Done

go/pkg/client/cas.go
rubensf: Fixed.

go/pkg/client/cas.go
rubensf: I think that would be convenient, but `Requested` and `Cached` are higher level concepts than "real bytes" and "compressed bytes", and don't necessarily make much sense at this level. Wdyt?

ola-rozenfeld:
I'll take another look once the tests pass.

rubensf:
*sigh* fixed.

go/pkg/client/cas.go
ola-rozenfeld: Idk, I'm only willing to be that we will get invariably confused at which functions return which numbers... Also, I'm particularly worried about changing public function APIs unnecessarily.
The only reason Cached doesn't apply here is because it is only dealt with in the *IfMissing flow, right? But everything else should apply, no?
Anyway, up to you, I'm just worried that it's confusing and will still change a bunch...

go/pkg/client/cas.go
```diff
@@ -36,6 +36,32 @@ const DefaultCompressedBytestreamThreshold = -1
...
+	// RealMoved is the the sum of sizes in bytes for all blobs passed
+	// through the wire in the format they were passed through (eg
+	// compressed).
+	RealMoved int64
+	// Cached is amount of logical bytes that we did not have to move
+	// through the wire because they were de-duped.
+	Cached int64
+}
+
+func sumStats(a, b *MovedBytesMetadata) {
```
ola-rozenfeld, line 58: nit: If b == nil { return }
and then return nil from everywhere when error is returned.

go/pkg/client/cas.go
```diff
@@ -1106,56 +1135,69 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
 }
 
 // DownloadDirectory downloads the entire directory of given digest.
-func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoot string, cache filemetadata.Cache) (map[string]*TreeOutput, error) {
+// It returns the number of logical and real bytes downloaded, which may be different from sum
+// of sizes of the files due to dedupping and compression.
+func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoot string, cache filemetadata.Cache) (map[string]*TreeOutput, MovedBytesMetadata, error) {
```
ola-rozenfeld, line 1140: Return a *MovedBytesMetadata here and elsewhere. That way, you initialize them with &MovedBytesMetadata{...}, too. Basically, you only ever deal with pointers.

go/pkg/client/cas.go
```diff
@@ -1228,7 +1276,12 @@ type downloadRequest struct {
 	context context.Context
 	output  *TreeOutput
 	meta    *requestMetadata
-	wait    chan<- error
+	wait    chan<- *downloadResponse
+}
+
+type downloadResponse struct {
+	stats MovedBytesMetadata
```
ola-rozenfeld, line 1283: By pointer

ola-rozenfeld (commented):
Some more nits, but other than that LGTM! Thank you!

go/pkg/client/cas_test.go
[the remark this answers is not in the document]
gkousik: I"m hoping we won't do this often and even then we wouldn't be changing the defaults?

go/pkg/client/cas.go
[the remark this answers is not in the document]
gkousik: nit: fullstop at the end of thisline.

go/pkg/client/cas.go
gkousik: We discussed this offline, so lets keep it there :)

go/pkg/client/cas.go
gkousik: Resolving since we are going to be keeping Cached around.

gkousik (approved):
Sorry for the slooooow review :)

go/pkg/client/cas.go
rubensf: I changed it to your suggestion and tbh it does look way nicer now :)

go/pkg/client/cas_test.go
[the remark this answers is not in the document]
rubensf: So I think my options here to use the actual expected values for `got` and `bMoved` are:
1) Have it explicitly in the test cases. However, I think it's much simpler and cleaner to just run all test cases for compression both on and off.
2) I'd still have to add a `if tc.compress` to check against decide to check against the compressed or non-compressed real values.

So I went in a slightly different path here: as per Ola's suggestions to always return `MovedBytesMetadata`, I now do already have the logical and real bytes, so the checks are slightly simpler (the first one is for both compressed and non-compressed cases, the second one is only for compressed cases).

Wdyt?

go/pkg/client/cas_test.go
```diff
@@ -250,22 +250,34 @@ func TestRead(t *testing.T) {
...
+					t.Errorf("c.ReadBlob(ctx, digest) = _, %v - same amount of real and logical bytes despite compression on", bMoved)
+				}
+				if !tc.compress && bMoved != int64(len(got)) {
```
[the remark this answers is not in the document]
rubensf: Discussion on the other thread :)

rubensf (commented):
Heh dw I've had much longer ones!

go/pkg/client/cas.go
```diff
@@ -1141,56 +1178,67 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
...
+// It returns the number of logical and real bytes downloaded, which may be different from sum
+// of sizes of the files due to dedupping and compression.
+func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoot string, cache filemetadata.Cache) (map[string]*TreeOutput, *MovedBytesMetadata, error) {
 	dir := &repb.Directory{}
-	if err := c.ReadProto(ctx, d, dir); err != nil {
-		return nil, fmt.Errorf("digest %v cannot be mapped to a directory proto: %v", d, err)
+	stats := &MovedBytesMetadata{}
+
+	protoStats, err := c.ReadProto(ctx, d, dir)
+	sumStats(stats, protoStats)
```
ola-rozenfeld, line 1188: Go nit: this should be stats.AddFrom(protostats)
And I think your use-case would improve if it returned a self pointer.

ola-rozenfeld (approved):
Just one nit, LGTM otherwise. Also, reminder to do the features=race thing just in case. I think we forgot to submit the PR that added it...

ola-rozenfeld (approved):

PR #267, opened 2020-12-21
Add shared zstd encoder pools.

rubensf:
On some sample testing this compression performs ~4x faster
than "wasting" the encoders.

### DIFF
--- a/go/pkg/reader/reader.go
+++ b/go/pkg/reader/reader.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/klauspost/compress/zstd"
+	"github.com/mostynb/zstdpool-syncpool"
 )
 
 type Initializable interface {
@@ -153,12 +154,7 @@ type compressedSeeker struct {
 }
 
 var encoderInit sync.Once
-var encoders sync.Pool
-
-func newEncoder() interface{} {
-	e, _ := zstd.NewWriter(nil)
-	return e
-}
+var encoders *sync.Pool
 
 // NewCompressedFileSeeker creates a ReadSeeker based on a file path.
 func NewCompressedFileSeeker(path string, buffsize int) (ReadSeeker, error) {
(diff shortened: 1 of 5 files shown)


go/pkg/reader/reader.go
```diff
@@ -148,8 +148,16 @@ type compressedSeeker struct {
...
-	buf    *syncedBuffer
-	closed bool
+	buf *syncedBuffer
+	err error
+}
+
+var encoderInit sync.Once
+var encoders sync.Pool
+
+func newEncoder() interface{} {
```
mostynb, line 158: I think this is leaky: zstd.NewWriter spawns goroutines that refer to some internal data, then when the zstd.Encoder is purged from the sync.Pool that internal data + goroutines leak.

Here is an example workaround: https://github.com/mostynb/zstdpool-syncpool

And some background: https://github.com/klauspost/compress/issues/264

go/pkg/reader/reader.go
```diff
@@ -148,8 +148,16 @@ type compressedSeeker struct {
...
+	e, _ := zstd.NewWriter(nil)
```
mostynb, line 159: You might want to use zstd.WithEncoderConcurrency(1), to avoid creating more goroutines than are necessary.

rubensf: Ah thanks!

go/pkg/reader/reader.go
rubensf: I missed not only the goroutines would leak but also the encoder itself since GC wouldn't ever catch it.

I imported your lib in - thanks!

mostynb: No problem- feedback welcome. It hasn't been thoroughly tested yet, but I figure it makes sense to share the effort in figuring out how to use this zstd api safely/efficiently (hence the tiny lib).

gkousik (approved):

go/pkg/reader/reader.go
rubensf: Will certainly send PRs if I find any issues :)

PR #269, opened 2021-01-12
Fixing header overflow bug in unified operations

ola-rozenfeld:


go/pkg/client/cas.go
```diff
@@ -164,6 +164,47 @@ func updateAndNotify(st *uploadState, bytesMoved int64, err error, missing bool)
...
+	var keys []string
+	for k := range labels {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return strings.Join(keys, ",")
+}
+
+// capToMaxHeaderSize ensures total length does not exceed max header size.
+func capToMaxHeaderSize(m *requestMetadata, limit int) {
```
rubensf, line 177: Nit: Rename `limit` to `headerSize` or the function name to `capToLimit`/`capReqToLimit`

go/pkg/client/cas.go
```diff
@@ -173,24 +214,13 @@ func getUnifiedMetadata(metas []*requestMetadata) *requestMetadata {
...
-	return &requestMetadata{
-		toolName:     strings.Join(tnKeys, ","),
-		actionID:     strings.Join(aiKeys, ","),
-		invocationID: strings.Join(iiKeys, ","),
+	m := &requestMetadata{
+		toolName:     getUnifiedLabel(toolNames),
+		actionID:     getUnifiedLabel(actionIDs),
+		invocationID: getUnifiedLabel(invocationIDs),
 	}
+	capToMaxHeaderSize(m, 15*1024)
```
rubensf, line 222: should this limit be defined somewhere? a constant (which has a descriptive name)? with a comment attached? or a flag controlled input?

ola-rozenfeld: Good point, I had no idea that the limit existed, the only mention I found is https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md, where a default of 8K is suggested. AFAIK, it is configurable on the server side, but unless we somehow pass it back to the client in Capabilities I'm not sure how would the client get this limit. Now that I checked, I think 8K is a safer value in general, even though RBE allows 16K.

go/pkg/client/cas.go

go/pkg/client/cas.go
```diff
@@ -173,24 +214,16 @@ func getUnifiedMetadata(metas []*requestMetadata) *requestMetadata {
...
-		actionID:     strings.Join(aiKeys, ","),
-		invocationID: strings.Join(iiKeys, ","),
+	m := &requestMetadata{
+		toolName:     getUnifiedLabel(toolNames),
+		actionID:     getUnifiedLabel(actionIDs),
+		invocationID: getUnifiedLabel(invocationIDs),
 	}
+	// We cap to a bit less than the maximum header size in order to allow
+	// for some proto fields serialization overhead.
+	const kDefaultMaxHeaderSize = 8 * 1024
```
rubensf, line 224: I'd make it a global package variable rather than a function one - the compiler is probably smart enough to inline it, but still :)

rubensf (approved):

go/pkg/client/cas.go

PR #271, opened 2021-01-27
Adding a single retry if a blob digest doesn't match expected

ola-rozenfeld:


ola-rozenfeld:
Note: this doesn't add any similar checks to BatchReadBlobs (which would be a good idea for another PR, because we are 100% relying on the server to do the right thing).

ola-rozenfeld:
@atetubou : this is the workaround you asked for, PTAL. Thank you!

go/pkg/client/cas.go
```diff
@@ -779,57 +779,80 @@ func marshalledRequestSize(d digest.Digest) int64 {
...
+		return nil, err
+	}
+	res := buf.Bytes()
+	// Compare the digest of the read blob to expected and retry the call once on difference.
+	// We can't reliably calculate digest without the full blob.
+	unexpectedResult := func() bool {
+		return err == nil && (len(res) != int(sz) || (offset == 0 && limit <= 0 && dg != recvDg))
+	}
+	if unexpectedResult() {
+		buf.Reset() // Retry once.
```
atetubou, line 827: could you do one more retry?

go/pkg/client/cas.go
```diff
@@ -779,57 +779,80 @@ func marshalledRequestSize(d digest.Digest) int64 {
...
+	buf := bytes.NewBuffer(make([]byte, 0, sz+bytes.MinRead))
+	_, recvDg, err := c.readBlobStreamed(ctx, dg, offset, limit, buf)
+	if err != nil {
+		return nil, err
+	}
+	res := buf.Bytes()
+	// Compare the digest of the read blob to expected and retry the call once on difference.
+	// We can't reliably calculate digest without the full blob.
+	unexpectedResult := func() bool {
+		return err == nil && (len(res) != int(sz) || (offset == 0 && limit <= 0 && dg != recvDg))
```
atetubou, line 824: better to check `err` just after it is got?

go/pkg/client/cas.go

go/pkg/client/cas.go
ola-rozenfeld: You're right, this looks a lot cleaner!

go/pkg/client/cas.go
```diff
@@ -779,57 +779,81 @@ func marshalledRequestSize(d digest.Digest) int64 {
...
 	}
-	sz += bytes.MinRead // Pad size so bytes.Buffer does not reallocate.
-	buf := bytes.NewBuffer(make([]byte, 0, sz))
-	_, err := c.readBlobStreamed(ctx, hash, sizeBytes, offset, limit, buf)
-	return buf.Bytes(), err
+	// Pad size so bytes.Buffer does not reallocate.
+	buf := bytes.NewBuffer(make([]byte, 0, sz+bytes.MinRead))
+	// Compare the digest of the read blob to expected and retry the call on difference.
+	numRetries := 2
+	recvDg := digest.Empty
```
atetubou, line 818: better to use `var recvDg digest.Digest`?

go/pkg/client/cas.go
```diff
@@ -779,57 +779,81 @@ func marshalledRequestSize(d digest.Digest) int64 {
...
 	f, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, c.RegularMode)
 	if err != nil {
 		return 0, err
 	}
 	defer f.Close()
-	return c.readBlobStreamed(ctx, hash, sizeBytes, 0, 0, f)
+	numRetries := 2
+	n, recvDg := int64(0), digest.Empty
+	for i := 0; i < numRetries; i++ {
+		f.Truncate(0)
```
atetubou, line 845: check error?

go/pkg/client/cas.go
```diff
@@ -779,57 +779,81 @@ func marshalledRequestSize(d digest.Digest) int64 {
...
+		f.Seek(0, 0)
```
atetubou, line 846: check error?

go/pkg/client/cas.go

go/pkg/client/cas.go

go/pkg/client/cas.go
ola-rozenfeld: Sure.
Btw, I had a little bug in the logic which I now fixed, PTAL! Thank you!

atetubou (approved):

atetubou:
Thanks, but update title of PR? We do multiple retry now.

go/pkg/client/cas.go
```diff
@@ -900,43 +934,32 @@ func (wt *writerTracker) Close() error {
...
-	if sizeBytes == 0 {
+func (c *Client) readBlobStreamed(ctx context.Context, d digest.Digest, offset, limit int64, w io.Writer) (int64, digest.Digest, error) {
+	if d.Size == 0 {
 		// Do not download empty blobs.
-		return 0, nil
+		return 0, digest.Empty, nil
 	}
 	wt := newWriteTracker(w)
-	name, wc, done, err := c.maybeCompressReadBlob(hash, sizeBytes, wt)
+	name, wc, done, err := c.maybeCompressReadBlob(d.Hash, d.Size, wt)
```
rubensf, line 943: `maybeCompressReadBlob` could also take digest now

go/pkg/client/cas.go
```diff
@@ -779,57 +779,88 @@ func marshalledRequestSize(d digest.Digest) int64 {
...
-	buf := bytes.NewBuffer(make([]byte, 0, sz))
-	_, err := c.readBlobStreamed(ctx, hash, sizeBytes, offset, limit, buf)
-	return buf.Bytes(), err
+	// Pad size so bytes.Buffer does not reallocate.
+	buf := bytes.NewBuffer(make([]byte, 0, sz+bytes.MinRead))
+	// Compare the digest of the read blob to expected and retry the call on difference.
+	numAttempts := 3
+	var err error
+	var recvDg digest.Digest
+	for i := 0; i < numAttempts; i++ {
```
rubensf, line 820: Does it make sense to `retrier.WithPolicy(ctx, func(error) { return true }, retrier.Immediatelly(3), downloadAndCheck)`?
downloadAndCheck can be the same, and Reset/(Truncate and Seek) can be abstracted in a `Reset() error`.

Or even, I think you could keep the original check in `readBlobStreamed`, add a type to the hash error, and do

```
func myReset_File() { Truncate; Seek }
func myReset_Buffer() { Reset }

retrier.WithPolicy(ctx, func(err error) { errors.Is(err, HashMismatch) }, retrier.Immediatelly(3), func() {
myReset()
readBlobStreamed()
})
```
?

ola-rozenfeld: As discussed offline: removed the public ReadBlobStreamed function and unified the behaviour under a resettableWriter interface.

go/pkg/client/cas.go

go/pkg/client/cas.go
```diff
@@ -900,43 +892,79 @@ func (wt *writerTracker) Close() error {
...
-		return n, err
-	}
-	if err = wc.Close(); err != nil {
-		return n, err
-	}
-	if err := <-wt.ready; err != nil {
-		return n, err
-	}
-	close(wt.ready)
-	sz := sizeBytes - offset
```
rubensf, line 924: Did you drop this becasue it's arguably redundant with hash checking?
I think we should keep it anyway - and actually move it before the hash check. I just heard a thought that we can maybe have issues due to truncated reads which lead to hash mismatches.

go/pkg/client/retries_test.go
```diff
@@ -287,6 +298,39 @@ func TestReadRetries(t *testing.T) {
...
+				client.RPCTimeouts(map[string]time.Duration{"default": 500 * time.Millisecond}).Apply(f.client)
+			}
+
+			blob := []byte("blob")
+			path := filepath.Join(t.TempDir(), strings.Replace(t.Name(), "/", "_", -1))
+			n, err := f.client.ReadBlobToFile(f.ctx, digest.NewFromBlob(blob), path)
+			if err != nil {
+				t.Errorf("client.ReadBlobToFile(ctx, digest) gave error %s, want nil", err)
+			}
+			if n != 4 {
```
rubensf, line 320: `if n != len(blob)`, and in the error msg.

ola-rozenfeld: Done.

go/pkg/client/cas.go
```diff
@@ -900,43 +892,79 @@ func (wt *writerTracker) Close() error {
...
+		if err = <-wt.ready; err != nil {
+			return err
+		}
+		close(wt.ready)
+		if err = <-done; err != nil {
+			return fmt.Errorf("Failed to finalize writing downloaded data downstream: %v", err)
+		}
+		close(done)
+		// We can't reliably calculate hash without the full blob
+		if offset == 0 && limit <= 0 && wt.dg != d {
```
rubensf, line 969: Not this cl's fault, but I think instead of `offset == 0 && limit <= 0`, we should check `d.SizeBytes == sz`. The limit being > 0 doesn't mean we couldn't potentially be attempting to pull the entire blob.

go/pkg/client/cas.go
ola-rozenfeld: Done.

rubensf (approved):

go/pkg/client/cas.go

PR #273, opened 2021-02-04
Fix compressed read retries in compressed blobs.

rubensf:
The current configuration will try to use the offset of the _compressed_
blob as the new offset for retrying reads. However, the compressed RE
API wants the offset to refer to the _uncompressed_ blobs. Since the
compressed offset will usually be a lower number than the uncompressed
offset, we generally end up with a bigger blob than expected.

Since this Read interruptions are mostly transient (eg context deadlines
exceeded), our higher level retries have been mostly hiding the issue.

Also took the chance to beef up some comments around the thread tricky
compression code.

I'll _eventually_ add retries tests to compressed blobs writes retry tests :)

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -18,7 +18,6 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/chunker"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/filemetadata"
-	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
 	"github.com/golang/protobuf/proto"
 	"github.com/klauspost/compress/zstd"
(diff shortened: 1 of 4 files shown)


rubensf:
In the process, I've removed the retrier based on a wrong digest, since this pretty much should have been our flake reason...

Wdyt?

go/pkg/client/retries_test.go
```diff
@@ -276,58 +288,71 @@ func TestWriteRetries(t *testing.T) {
...
-			blob := []byte("blob")
-			got, err := f.client.ReadBlob(f.ctx, digest.NewFromBlob(blob))
-			if err != nil {
-				t.Errorf("client.ReadBlob(ctx, digest) gave error %s, want nil", err)
-			}
-			if diff := cmp.Diff(blob, got, cmpopts.EquateEmpty()); diff != "" {
-				t.Errorf("client.ReadBlob(ctx, digest) gave diff (-want, +got):\n%s", diff)
-			}
-		})
+		for _, comp := range []bool{false, true} {
```
ola-rozenfeld, line 291: How much longer did it make the test? If it added a few seconds, then suggestion: make compression retry test a separate Test function from this one that includes the sleeping, because these two are really orthogonal. Ditto for files.

go/pkg/client/cas.go
```diff
@@ -938,40 +961,53 @@ func (c *Client) readBlobStreamed(ctx context.Context, d digest.Digest, offset,
 	if limit > 0 && limit < sz {
 		sz = limit
 	}
+	if err = w.Reset(); err != nil {
```
ola-rozenfeld, line 964: Wait, if you moved this Reset outside of the closure, that tells me you no longer need it to be resettable. And you're removing the functionality of retrying on failed digest verification, too.
So, suggestion: base this PR on a revert of my https://github.com/bazelbuild/remote-apis-sdks/pull/271. Should be much simpler, imo.

go/pkg/client/cas.go
```diff
@@ -843,24 +846,36 @@ func NewCompressedWriteBuffer(w io.WriteCloser) (io.WriteCloser, chan error, err
...
+		// separate thread. As such, we also need a way to signal the main
+		// thread that the decoding has finished - which will have some delay
+		// from the last Write call.
 		_, err := decoder.WriteTo(w)
 		if err != nil {
 			// Because WriteTo returned early, the pipe writers still
 			// have to go somewhere or they'll block execution.
 			io.Copy(ioutil.Discard, r)
 		}
-		w.Close()
```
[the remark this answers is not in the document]
rubensf: Under `if d.Size == sz {` in the new file... `if err := wt.Close()`. 

And I thought the CI was doing `features=race` already, _sighs_. Caught a big, thanks for reminding me!

And i don't think we have windows CI.

go/pkg/client/retries_test.go
rubensf: The non race tests still take < 10s (less than 1s of noticeable increase).

I'm not sure they're orthogonal - I'd argue they _should_ be the same. As much as possible, comp vs noncomp should behave the same way.

go/pkg/client/cas.go
ola-rozenfeld: We'll have to clean this up in the fixit next week, then. There's a lot of stuff here now that no longer should be that will just be confusing. We no longer need the whole ResettableWriter struct and interface, and the refactoring I did that made everything more complicated...
But I'm okay with leaving it as a cleanup task for the next CL.

go/pkg/client/retries_test.go
ola-rozenfeld: Sure, if it doesn't make it noticeably longer, this is fine!

ola-rozenfeld (approved):

PR #276, opened 2021-02-11
Bugfix: the variables in defer need to be evaluated at the end.

ola-rozenfeld:


go/pkg/rexec/rexec.go
```diff
@@ -363,7 +363,7 @@ func (ec *Context) DownloadOutErr() {
 // DownloadOutputs downloads the outputs of the command in the context to the specified directory.
 func (ec *Context) DownloadOutputs(execRoot string) {
 	st := ec.Result.Status
-	stats, res := ec.downloadOutputs(ec.cmd.ExecRoot)
+	stats, res := ec.downloadOutputs(execRoot)
```
rubensf, line 366: can we instead rename `execRoot` to `downloadDir` or something like that? IIUC this path has nothing to do with exec root - and it was clearly misleading :)

go/pkg/client/cas.go
```diff
@@ -1483,7 +1483,7 @@ func (c *Client) downloadSingle(ctx context.Context, dg digest.Digest, reqs map[
 	// We cannot release the lock after each individual file copy, because
 	// the caller might move the file, and we don't have the contents in memory.
 	bytesMoved := map[digest.Digest]*MovedBytesMetadata{}
-	defer afterDownload([]digest.Digest{dg}, reqs, bytesMoved, err)
+	defer func() { afterDownload([]digest.Digest{dg}, reqs, bytesMoved, err) }()
```
rubensf, line 1486: *shamecube* dropped the ball here

rubensf (approved):

go/pkg/rexec/rexec.go
ola-rozenfeld: Done.

PR #277, opened 2021-02-11
Surfacing gRPC error details in the error message. Attempt #2.

ola-rozenfeld:


go/pkg/client/tree.go
```diff
@@ -1,4 +1,3 @@
-// Package tree provides functionality for constructing a Merkle tree of uploadable inputs.
```
rubensf, line 1: Maybe just move this under "package" but leave the comment?

go/pkg/client/cas.go
```diff
@@ -28,6 +28,7 @@ import (
 
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 	log "github.com/golang/glog"
+	_ "google.golang.org/genproto/googleapis/rpc/errdetails" // the proto needs to be compiled in for UnmarshallAny.
```
rubensf, line 31: Maybe more context than "unmarshallany"? since that method isn't called in this file.

go/pkg/client/cas.go
```diff
@@ -103,6 +104,50 @@ type uploadState struct {
 	cancel  func()
 }
 
+// StatusError is the same as status.Error except it includes the error details in the error message.
+type StatusError struct {
```
rubensf, line 108: Maybe have this error outside of `cas.go`? Maybe still under `package client`, but cas.go is large and IIUC this isn't tied at dependant.

rubensf (approved):

go/pkg/client/tree.go
ola-rozenfeld: This is a remnant of the time when tree used to actually be a separate package. Now that it's a part of client, I see no point to have a top-of-file comment, because that would just confuse the godoc (which comment should it show when generating the client docs?). I'm surprised this did not fail our external gofmt.

go/pkg/client/cas.go
ola-rozenfeld: Yep, agreed. Added status.go file.

go/pkg/client/cas.go
ola-rozenfeld: Right, changed.

go/pkg/client/tree.go
rubensf: I meant not as a top of file doc -- maybe something after imports. To still have some context, we already have too little documentation as is :)

ola-rozenfeld: Okay, we'll see if that works.

PR #278, opened 2021-02-16
Refactoring to allow passing in the tool version to the remote server

ola-rozenfeld:


ola-rozenfeld:
This is non backward compatible! Both re-client and internal usage will require modifications on import of this commit.
However, I think it simplifies the API and is overall for the best. I should have done it this way to begin with...

go/pkg/client/client_context.go
```diff
@@ -19,55 +19,80 @@ const (
 	remoteHeadersKey = "build.bazel.remote.execution.v2.requestmetadata-bin"
 )
 
+// ContextMetadata is optionally attached to RPC requests.
+type ContextMetadata struct {
```
rubensf, line 23: Can we just use `RequestMetadata`[1] instead? Or make it a field of `ContextMetadata`?

1: https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L1677

go/pkg/client/cas.go
```diff
@@ -208,50 +202,53 @@ func getUnifiedLabel(labels map[string]bool) string {
 }
 
 // capToLimit ensures total length does not exceed max header size.
-func capToLimit(m *requestMetadata, limit int) {
-	total := len(m.toolName) + len(m.actionID) + len(m.invocationID)
+func capToLimit(m *ContextMetadata, limit int) {
```
rubensf, line 205: Don't you have to now account for all of ContextMetadata here?

rubensf (commented):
This all does look better! 

(sorry for sending comments in two passes :')

go/pkg/client/client_context.go
ola-rozenfeld: I didn't want to do that specifically, that's why I had the internal struct. I wanted to decouple this from the API protos. Besides, the RequestMetadata proto is clunkier to use (with the internal structure that it has.)

go/pkg/client/cas.go
ola-rozenfeld: That would be difficult, so I still only do the merging and capping for the two fields, ActionID and InvocationID specifically. I even regress combining the tool name here (because it made little sense anyway).

rubensf: I agree it's difficult (maybe hacky?), but like `m.ToolName`, we have to account for it when checking the full header size?

go/pkg/client/client_context.go
rubensf: Sgtm!

go/pkg/client/cas.go
ola-rozenfeld: Right, sorry! Done.

rubensf (approved):

PR #282, opened 2021-02-18
Optimise out uploading the empty blob in a couple of places. Closes #184

ola-rozenfeld:


go/pkg/client/cas.go
```diff
@@ -565,6 +574,10 @@ func (c *Client) WriteProto(ctx context.Context, msg proto.Message) (digest.Dige
 func (c *Client) WriteBlob(ctx context.Context, blob []byte) (digest.Digest, error) {
 	ue := uploadinfo.EntryFromBlob(blob)
 	dg := ue.Digest
+	if dg.IsEmpty() {
+		LogContextInfof(ctx, log.Level(2), "Skipping upload of empty blob")
```
rubensf, line 578: Include blob digest? For all the logs.

rubensf (approved):

go/pkg/client/cas.go
ola-rozenfeld: It's the empty digest, though... But sure, will do.

PR #285, opened 2021-02-19
Add shared zstd decoder pools.

rubensf:
Testing on internal builds yielded improvements :)

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -21,6 +21,7 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
 	"github.com/golang/protobuf/proto"
 	"github.com/klauspost/compress/zstd"
+	"github.com/mostynb/zstdpool-syncpool"
 	"github.com/pborman/uuid"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


go/pkg/client/cas.go
```diff
@@ -868,9 +872,18 @@ func NewCompressedWriteBuffer(w io.Writer) (io.WriteCloser, chan error, error) {
 	// the writer data.
 	r, nw := io.Pipe()
 
+	decoderInit.Do(func() {
+		decoders = syncpool.NewDecoderPool(zstd.WithDecoderConcurrency(1))
+	})
+
 	// TODO(rubensf): Reuse decoders when possible to save the effort of starting/closing goroutines.
```
mostynb, line 879: I guess you can remove this TODO?

go/pkg/client/cas.go
```diff
@@ -881,13 +894,14 @@ func NewCompressedWriteBuffer(w io.Writer) (io.WriteCloser, chan error, error) {
 		// separate thread. As such, we also need a way to signal the main
 		// thread that the decoding has finished - which will have some delay
 		// from the last Write call.
-		_, err := decoder.WriteTo(w)
+		_, err := decoderW.Decoder.WriteTo(w)
```
mostynb, line 897: I think this could just be `decoderW.WriteTo(w)` since the Decoder is embedded. And similarly for the `decoderW.Decoder.Reset` call further up.

go/pkg/client/cas.go
rubensf: Yup - thanks for catching!

go/pkg/client/cas.go
rubensf: Fixed, thanks!

ola-rozenfeld (approved):

PR #287, opened 2021-03-01
remotetool: add command to upload a tree into the CAS

Yannic:


go/pkg/tool/tool.go
```diff
@@ -274,18 +273,47 @@ func (c *Client) DownloadBlob(ctx context.Context, blobDigest, path string) (str
...
 	}
+	return uploader.CloseAndWait()
+}
 
-	log.Infof("Uploading blob of %v from %v.", dg, path)
-	ue := uploadinfo.EntryFromFile(dg, path)
-	if _, _, err := c.GrpcClient.UploadIfMissing(ctx, ue); err != nil {
+// UploadTree uploads a tree from the specified path into the remote cache.
+func (c *Client) UploadTree(ctx context.Context, concurrency uint64, path string) error {
+	uploader := newParallelUploader(ctx, c.GrpcClient, concurrency)
```
rubensf, line 293: I wonder why couldn't you use `UploadIfMissing` directly instead of making a parallel uploaded? `UploadIfMissing` should already has [parallel support](https://github.com/bazelbuild/remote-apis-sdks/blob/fdb5dae38ca890ae7311dc0d320e2d327df10485/go/pkg/client/cas.go#L419), controlled by [CASConcurrency](https://github.com/bazelbuild/remote-apis-sdks/blob/fdb5dae38ca890ae7311dc0d320e2d327df10485/go/pkg/flags/flags.go#L46)

go/cmd/remotetool/main.go
```diff
@@ -53,14 +54,16 @@ var supportedOps = []OpType{
...
 
 var (
-	operation    = flag.String("operation", "", fmt.Sprintf("Specifies the operation to perform. Supported values: %v", supportedOps))
-	digest       = flag.String("digest", "", "Digest in <digest/size_bytes> format.")
-	pathPrefix   = flag.String("path", "", "Path to which outputs should be downloaded to.")
-	inputRoot    = flag.String("input_root", "", "For reexecute_action: if specified, override the action inputs with the specified input root.")
-	execAttempts = flag.Int("exec_attempts", 10, "For check_determinism: the number of times to remotely execute the action and check for mismatches.")
+	operation         = flag.String("operation", "", fmt.Sprintf("Specifies the operation to perform. Supported values: %v", supportedOps))
+	digest            = flag.String("digest", "", "Digest in <digest/size_bytes> format.")
+	pathPrefix        = flag.String("path", "", "Path to which outputs should be downloaded to.")
```
gkousik, line 63: I think this documentation should be updated?

go/pkg/tool/tool.go
gkousik: +1 to this.

go/pkg/tool/tool_test.go
```diff
@@ -276,3 +280,50 @@ func TestTool_UploadBlob(t *testing.T) {
...
+	// First request should upload the blob.
+	if cas.WriteReqs() != 4 {
+		t.Fatalf("Expected 4 writes, got %v", cas.WriteReqs())
+	}
+
+	// Retries should check whether the blob already exists and skip uploading if it does.
+	if err := toolClient.UploadTree(context.Background(), tempDir); err != nil {
+		t.Fatalf("UploadTree('%v') failed: %v", tmpFile, err)
+	}
+	if cas.WriteReqs() != 4 {
```
gkousik, line 326: If we are skipping the blob upload during second call to the same `tempDir` path, then shouldn't we have 0 write requests?

Yannic: The stats aren't reset, so they continue to be 4 from the initial upload. If the second call uploads anything we'd get > 4 here.

go/pkg/tool/tool.go
Yannic: I'm using this tool to upload very large trees with 10,000s files. Just iterating over all of them takes significant time, so it's more efficient to start uploading while iterating over the files in the tree.

I'll have a look if I can achieve this with the existing parallel uploading.

ola-rozenfeld:
Was this superseded by https://github.com/bazelbuild/remote-apis-sdks/blob/e74bc3da2efc333cae5d79e88cfb0e0baeeaf4f1/go/pkg/tool/tool.go#L341 ?

PR #288, opened 2021-03-03
Extract NewClientFromConnection() from NewClient()

Jwata:
This change will allow us create a client with custom gRPC connections.
I simply extracted the logic from `NewClient()`.

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -600,6 +600,11 @@ func NewClient(ctx context.Context, instanceName string, params DialParams, opts
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


gkousik:
Since this is an exported function, can you add a test to client_test.go for this? A basic test to make sure the options are applied and proper parameters are set would do.

gkousik (changes requested):
Feel free to merge after adding the unit test.

Jwata:
@gkousik 
Thank you for your review.
I added unit tests. It seems I need your confirmation.

rubensf (approved):

gkousik (approved):

PR #290, opened 2021-03-08
Add utility function to return output file digests for a given exec context

gkousik:
This function is useful when we only want to know the output digests
without actually downloading them.

Test: Added unit tests, also integ test with re-client.

### DIFF
--- a/go/pkg/rexec/rexec.go
+++ b/go/pkg/rexec/rexec.go
@@ -4,6 +4,7 @@ package rexec
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -336,11 +337,11 @@ func (ec *Context) ExecuteRemotely() {
 	if ec.resPb != nil {
 		ec.setOutputMetadata()
 		ec.Result = command.NewResultFromExitCode((int)(ec.resPb.ExitCode))
-		log.V(1).Infof("%s %s> Downloading outputs...", cmdID, executionID)
 		if ec.opt.DownloadOutErr {
 			ec.Result = ec.downloadOutErr()
 		}
 		if ec.Result.Err == nil && ec.opt.DownloadOutputs {
+			log.V(1).Infof("%s %s> Downloading outputs...", cmdID, executionID)
 			stats, res := ec.downloadOutputs(ec.cmd.ExecRoot)
 			ec.Metadata.LogicalBytesDownloaded += stats.LogicalMoved
 			ec.Metadata.RealBytesDownloaded += stats.RealMoved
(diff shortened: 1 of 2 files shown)


go/pkg/rexec/rexec.go
```diff
@@ -384,6 +385,30 @@ func (ec *Context) DownloadOutputs(outputDir string) {
...
+	if ec.resPb == nil {
+		return nil, nil
+	}
+
+	ft, err := ec.client.GrpcClient.FlattenActionOutputs(ec.ctx, ec.resPb)
+	if err != nil {
+		return nil, err
+	}
+	res := map[string]digest.Digest{}
+	for p, t := range ft {
```
rubensf, line 402: use more comprehensive names "filepath", "filetree", "tree" (??). I'm not as familiar with the code so it's ambiguous

gkousik: Done.

go/pkg/rexec/rexec.go
```diff
@@ -384,6 +385,30 @@ func (ec *Context) DownloadOutputs(outputDir string) {
...
+		return nil, nil
+	}
+
+	ft, err := ec.client.GrpcClient.FlattenActionOutputs(ec.ctx, ec.resPb)
+	if err != nil {
+		return nil, err
+	}
+	res := map[string]digest.Digest{}
+	for path, outTree := range ft {
+		fp := path
```
rubensf, line 403: Do you need `fp := path`? Why not just `path`?

go/pkg/rexec/rexec.go
```diff
@@ -384,6 +385,30 @@ func (ec *Context) DownloadOutputs(outputDir string) {
 	}
 }
 
+// GetOutputFileDigests returns a map of output file paths to digests.
+// This function is supposed to be run after a successful cache-hit / remote-execution
+// has been run with the given execution context. If called before the completion of
+// remote-execution, the function returns a nil result.
+func (ec *Context) GetOutputFileDigests(useAbsPath bool) (map[string]digest.Digest, error) {
```
rubensf, line 392: useAbsPath is convenient but does it make sense? You can arguably parse the output with other tools.

I'm not exactly opposed to it but I think long term it'd work better as a SDK wide option (eg show full path on outputs for convenience).

go/pkg/rexec/rexec.go
gkousik: I guess not, changed.

go/pkg/rexec/rexec.go
gkousik: I mean currently I don't have a usecase for the absolute paths, I only want relative paths. That said, I think it is a useful piece of functionality for this method to optionally return either the relative / absolute paths for the output files. There are going to be multiple users for the SDK.

rubensf: Right, I am just wondering if this is the correct approach vs a SDK Client option for all places we show output back to the user :)

... but as usual, we can decide later.

rubensf (approved):

PR #292, opened 2021-03-09
Fix SDK to use working-directory relative output paths.

ramymedhat:
This is the correct implementation according to the remote api.

See
https://github.com/bazelbuild/remote-apis/blob/0943dc4e70e1414735a85a3167557392c177ff45/build/bazel/remote/execution/v2/remote_execution.proto#L609
and
https://github.com/bazelbuild/remote-apis/blob/0943dc4e70e1414735a85a3167557392c177ff45/build/bazel/remote/execution/v2/remote_execution.proto#L574


go/pkg/client/cas.go
```diff
@@ -1232,34 +1232,35 @@ func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoo
...
+	outStats, err := c.downloadOutputs(ctx, outputs, root, cache)
 	stats.addFrom(outStats)
 	return outputs, stats, err
 }
 
 // DownloadActionOutputs downloads the output files and directories in the given action result. It returns the amount of downloaded bytes.
 // It returns the number of logical and real bytes downloaded, which may be different from sum
 // of sizes of the files due to dedupping and compression.
-func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionResult, execRoot string, cache filemetadata.Cache) (*MovedBytesMetadata, error) {
+func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionResult, execRoot, workingDir string, cache filemetadata.Cache) (*MovedBytesMetadata, error) {
```
rubensf, line 1243: Same, `execRoot, workingDir` -> `outDir`.

This function doesn't need to worry about these RE API details.

rubensf (commented):
Three points:
- We can free cas.go from the concepts of `execRoot` and `workingDir`. They don't affect functionality of "download these files to this location".
- rexec.go should be calling `ec.downloadOutputs(filepath.Join(ec.cmd.ExecRoot, ec.cmd.WorkingDir))`.
- I think for now we should be adding a flag `--legacy_download_to_execroot` to control this on rexec.go in case we have clients that might have implemented this wrong. We can then remove it ~in the future.

go/pkg/client/cas.go
```diff
@@ -1232,34 +1232,35 @@ func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoo
...
+		if err := os.RemoveAll(filepath.Join(root, dir.Path)); err != nil {
 			return nil, err
 		}
 	}
-	return c.downloadOutputs(ctx, outs, execRoot, cache)
+	return c.downloadOutputs(ctx, outs, root, cache)
 }
 
-func (c *Client) downloadOutputs(ctx context.Context, outs map[string]*TreeOutput, execRoot string, cache filemetadata.Cache) (*MovedBytesMetadata, error) {
+func (c *Client) downloadOutputs(ctx context.Context, outs map[string]*TreeOutput, root string, cache filemetadata.Cache) (*MovedBytesMetadata, error) {
```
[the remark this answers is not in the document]
ramymedhat: Done.

go/pkg/client/cas.go
ramymedhat: Done.

go/pkg/client/cas.go
```diff
@@ -1209,7 +1209,7 @@ func (c *Client) FlattenActionOutputs(ctx context.Context, ar *repb.ActionResult
 // DownloadDirectory downloads the entire directory of given digest.
 // It returns the number of logical and real bytes downloaded, which may be different from sum
 // of sizes of the files due to dedupping and compression.
-func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoot string, cache filemetadata.Cache) (map[string]*TreeOutput, *MovedBytesMetadata, error) {
+func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, root string, cache filemetadata.Cache) (map[string]*TreeOutput, *MovedBytesMetadata, error) {
```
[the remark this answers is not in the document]
ramymedhat: Yeah that's why I changed execRoot to root, but outDir is clearer. Done.

go/pkg/rexec/rexec.go
```diff
@@ -130,7 +130,7 @@ func (ec *Context) downloadOutErr() *command.Result {
 func (ec *Context) downloadOutputs(execRoot string) (*rc.MovedBytesMetadata, *command.Result) {
```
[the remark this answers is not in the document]
ramymedhat: we override execroot in DownloadOutputs. I'll leave this as is now since we still support execRoot rooted downloads, and we can clean this up when we remove the new option. That way we can check for the option in one central place: downloadOutputs.

go/pkg/client/cas.go
```diff
@@ -1309,12 +1309,12 @@ func (c *Client) downloadOutputs(ctx context.Context, outs map[string]*TreeOutpu
 		if src.IsEmptyDirectory {
 			return fullStats, fmt.Errorf("unexpected empty directory: %s", src.Path)
 		}
-		if err := copyFile(execRoot, execRoot, src.Path, out.Path, perm); err != nil {
+		if err := copyFile(outDir, outDir, src.Path, out.Path, perm); err != nil {
```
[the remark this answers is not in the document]
ramymedhat: Done.

ramymedhat:
> > downloadOutputs is a function in the ExecutionContext, so it should be aware of the WorkingDir and not have it be passed unless there is a reason we don't use it (which is the case for ExecRoot).
> 
> in which case, we wouldn't need to pass anything to it since it's context aware. The way we use it, it isn't context aware. We might as well rename the arguments there to "outputDir".
> 
> > If clients break because of this change they can pin to an earlier version of the sdk?
> 
> yes but they miss out new updates? I don't think that's too much trouble for us to support, even more when all affected code lives in calls to `downloadOuputs`. Lmk if you think otherwise.

SG. Added the option to use legacy behavior.

go/pkg/rexec/rexec.go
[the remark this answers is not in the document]
rubensf: TL;DR My argument is that this shouldn't be named `execRoot`, just `outDir`, like in the `cas.go` functions.

I think I was confused myself and thus ambiguous - the Download functions in `rexec.go` are context aware, so they do need to know the exec root and the working directory as part of the general context of the action. But the _remote_ location of an artifact does not need to be the _local_ location of an artifact. So this argument shouldn't be _named_ execRoot.

In retrospect, to avoid ambiguity, we probably should use the name `downloadDir` or `downDir` rather than `outDir` everywhere relevant, but I'm ok if you'd rather not go around changing the code once again after I told you to do it once :sweat_smile:

go/pkg/client/cas_test.go
```diff
@@ -1221,6 +1221,195 @@ func TestDownloadActionOutputs(t *testing.T) {
 	}
 }
 
+func TestDownloadActionOutputsWithWorkingDir(t *testing.T) {
```
rubensf, line 1224: It looks like this is a copy of `TestDownloadActionOutputs`. I don't think this is necessary. There was no functionality change on `cas.go`. Might as well just tweak it to use `wd` as the `outDir` or try different `outDirs`.

rubensf (approved):
nits

rubensf (approved):

ramymedhat:
Realized local exec remote cache is not handled in the change to use working-dir-relative outputs. Fixed.

rubensf (approved):

go/pkg/client/tree.go
```diff
@@ -464,12 +464,16 @@ func getRelPath(base, path string) (string, error) {
 // ComputeOutputsToUpload transforms the provided local output paths into uploadable Chunkers.
 // The paths have to be relative to execRoot.
 // It also populates the remote ActionResult, packaging output directories as trees where required.
-func (c *Client) ComputeOutputsToUpload(execRoot string, paths []string, cache filemetadata.Cache) (map[digest.Digest]*uploadinfo.Entry, *repb.ActionResult, error) {
+func (c *Client) ComputeOutputsToUpload(execRoot, workingDir string, paths []string, cache filemetadata.Cache) (map[digest.Digest]*uploadinfo.Entry, *repb.ActionResult, error) {
```
atetubou, line 467: Could you keep old interface for gradual migration in our repository?

go/pkg/client/tree.go
```diff
@@ -464,12 +464,16 @@ func getRelPath(base, path string) (string, error) {
...
 	outs := make(map[digest.Digest]*uploadinfo.Entry)
 	resPb := &repb.ActionResult{}
 	for _, path := range paths {
-		absPath := filepath.Clean(filepath.Join(execRoot, path))
-		normPath, err := getRelPath(execRoot, absPath)
+		absPath := filepath.Clean(filepath.Join(execRoot, workingDir, path))
+		_, err := getRelPath(execRoot, absPath)
```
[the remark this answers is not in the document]
ramymedhat: Done

go/pkg/client/tree.go
ramymedhat: Is this still relevant? Can you elaborate on what gradual migration you're thinking of?

go/pkg/client/cas.go
```diff
@@ -1238,34 +1238,34 @@ func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, execRoo
...
+	outStats, err := c.downloadOutputs(ctx, outputs, outDir, cache)
 	stats.addFrom(outStats)
 	return outputs, stats, err
 }
 
 // DownloadActionOutputs downloads the output files and directories in the given action result. It returns the amount of downloaded bytes.
 // It returns the number of logical and real bytes downloaded, which may be different from sum
 // of sizes of the files due to dedupping and compression.
-func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionResult, execRoot string, cache filemetadata.Cache) (*MovedBytesMetadata, error) {
+func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionResult, outDir string, cache filemetadata.Cache) (*MovedBytesMetadata, error) {
```
rubensf, line 1249: FWIW after a reread of the comments, if you do want to go through the trouble, I think `downDir` is a more appropriate name - DownloadOutputs does not necessarily download outputs to the output directory (eg it can download them to a temporary location for racing strategies)

go/pkg/rexec/rexec_test.go
```diff
@@ -31,79 +31,104 @@ func TestExecCacheHit(t *testing.T) {
...
-		contents, err := ioutil.ReadFile(path)
-		if err != nil {
-			t.Errorf("error reading from %s: %v", path, err)
-		}
-		if !bytes.Equal(contents, []byte("output")) {
-			t.Errorf("expected %s to contain \"output\", got %v", path, contents)
-		}
+		})
 	}
+
```
rubensf, line 131: del?

rubensf (approved):
nits

go/pkg/client/cas.go
ramymedhat: I think outDir is reasonable enough and I really don't want to go renaming this again :)

go/pkg/rexec/rexec_test.go
ramymedhat: Done

PR #293, opened 2021-04-02
calculate filemeta data in parallel

atetubou:
This is to utilize cpu for hash calculation when there are many
large files.

Mostly for https://crbug.com/1193375

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -9,6 +9,9 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/command"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
@@ -123,7 +126,7 @@ func preprocessSymlink(execRoot, symlinkNormDir string, meta *filemetadata.Symli
 
 // loadFiles reads all files specified by the given InputSpec (descending into subdirectories
 // recursively), and loads their contents into the provided map.
-func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs map[string]*fileSysNode, cache filemetadata.Cache, opts *TreeSymlinkOpts) error {
+func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs *concurrentFileSysNodeMap, cache filemetadata.Cache, opts *TreeSymlinkOpts) error {
 	if opts == nil {
 		opts = DefaultTreeSymlinkOpts()
 	}
@@ -159,25 +162,25 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs 
 		return meta.Err
 	}
 	if t == command.FileInputType {
-		fs[normPath] = &fileSysNode{
+		fs.set(normPath, &fileSysNode{
 			file: &fileNode{
 				ue:           uploadinfo.EntryFromFile(meta.Digest, absPath),
 				isExecutable: meta.IsExecutable,
 			},
-		}
+		})
 		return nil
 	} else if t == command.SymlinkInputType {
 		relTarget, err := getTargetRelPath(execRoot, symlinkNormDir, meta.Symlink.Target)
 		if err != nil {
 			return err
 		}
-		fs[normPath] = &fileSysNode{
+		fs.set(normPath, &fileSysNode{
 			// We cannot directly use meta.Symlink.Target, because it could be
 			// an absolute path. Since the remote worker will map the exec root
 			// to a different directory, we must strip away the local exec root.
 			// See https://github.com/bazelbuild/remote-apis-sdks/pull/229#discussion_r524830458
 			symlink: &symlinkNode{target: relTarget},
-		}
+		})
 		if meta.Symlink.IsDangling || !opts.FollowsTarget {
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


atetubou:
Sorry, this seems not sufficient.

atetubou:
Sorry, let me re-open this again.

I confirmed this somewhat improves the performance of ComputeMerkleTree on Swarming bot.
See https://crbug.com/1193375#c17 for detail.

go/pkg/client/tree.go
```diff
@@ -190,21 +193,39 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
...
+		f := f
+		eg.Go(func() error {
+			return loadFiles(execRoot, excl, filepath.Join(normPath, f.Name()), fs, cache, opts)
+		})
 	}
-	return nil
+	return eg.Wait()
+}
+
+type concurrentFileSysNodeMap struct {
```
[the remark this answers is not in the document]
rubensf: Ah nvm, I didn't notice that this is write-only then read-only.

rubensf (approved):

go/pkg/client/tree.go
```diff
@@ -190,21 +193,39 @@ func loadFiles(execRoot string, excl []*command.InputExclusion, path string, fs
 	}
 
 	if len(files) == 0 {
-		fs[normPath] = &fileSysNode{emptyDirectoryMarker: true}
+		fs.set(normPath, &fileSysNode{emptyDirectoryMarker: true})
 		return nil
 	}
+
+	var eg errgroup.Group
 	for _, f := range files {
```
ramymedhat, line 201: You can move the semaphore here.

go/pkg/filemetadata/cache.go
```diff
@@ -17,11 +18,17 @@ type fmCache struct {
...
+	// Semaphoe to limit the number of parallel hash calculation.
+	computeSema chan struct{}
 }
 
 // NewSingleFlightCache returns a singleton-backed in-memory cache, with no validation.
 func NewSingleFlightCache() Cache {
-	return &fmCache{Backend: cache.GetInstance()}
+	return &fmCache{
+		Backend:     cache.GetInstance(),
+		computeSema: make(chan struct{}, runtime.NumCPU()),
```
ramymedhat, line 30: Did you experiment with going above num CPUs? I imagine you could handle more parallelism than num CPUs.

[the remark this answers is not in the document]
ramymedhat: Alternatively you can create a new constructor that constrains the number of parallel calls to N. The default constructor should be unconstrained.

ramymedhat (changes requested):

go/pkg/client/tree.go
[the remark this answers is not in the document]
rubensf: I think that should be the main sync.Map https://golang.org/pkg/sync/#Map
And the comments say:
```
The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.
```
IIUC the comments, I don't think this is a good use case for syncmap.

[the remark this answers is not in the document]
ramymedhat: I guess you will save a couple of lines for type checks. Map with mutex sgtm.

go/pkg/filemetadata/cache.go
```diff
@@ -17,11 +18,17 @@ type fmCache struct {
 	Backend     *cache.Cache
 	cacheHits   uint64
 	cacheMisses uint64
+
+	// Semaphoe to limit the number of parallel hash calculation.
+	computeSema chan struct{}
```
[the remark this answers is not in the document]
atetubou: done

go/pkg/filemetadata/cache.go
atetubou: No, but I think this is CPU intensive and we won't get more benefit with higher concurrency?

go/pkg/client/tree.go
atetubou: I don't think semaphore here does reasonable limit for parallelism as we call this function recursively?

go/pkg/filemetadata/cache.go
ramymedhat: There is an IO component that may allow higher parallelism I would imagine. There's a syscall to stat the file, and then reading to digest.

go/pkg/client/tree.go
ramymedhat: Ah I missed that. Then I guess a separate constructor would work.

go/pkg/filemetadata/cache.go
rubensf: I think having parallelism be an argument for `NewSingleFlightCache` can't hurt. We can come up with a heuristic for the optimal value depending on OS and CPU separately.

atetubou: In our issue, I'm suspecting cpu usage from hash calculation for large files is dominant part.
So I choose cpu num as default not to cause many context switch there.

And I added option to specify parallelism in case user needs to change.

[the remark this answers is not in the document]
atetubou: Added option. I feel num cpu is better default than unconstrained to avoid many context switch?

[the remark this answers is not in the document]
ramymedhat: Thanks for adding the option! I would prefer to leave the current default behavior intact until we confirm NumCPU works well for reclient. Note that our files are more biased towards being small, so not sure what the impact will be there.

go/pkg/filemetadata/cache.go
```diff
@@ -17,11 +21,17 @@ type fmCache struct {
 	Backend     *cache.Cache
 	cacheHits   uint64
 	cacheMisses uint64
+
+	// Semaphoe to limit the number of parallel hash calculation.
```
ramymedhat, line 25: Semaphore

ramymedhat (approved):
Actually on second thought, this patch will change the behavior on the client side because the number of goroutines will increase now that we're parallelizing hash calculations. We will need to test and tweak it anyway. So a default of NumCPUs parallelism sgtm.

PR #298, opened 2021-04-12
Revamp loadFiles for performance.

rubensf:
This gets rid of a few bottlenecks:

1. Recursion -> iteration. I'm not sure how much of a bottleneck this is, 
but in practice recursion is much more expensive than iteration.                                                                                                                                                                                                                    
2. Replace ioutil.ReadDir, which will both sort and stat files, so
unnecessarily slow.                                                                                                                                                                                                                                                                 
3. Remove lots of filepath operations, including Cleans. All the Cleans
we called were redundant with filepath pkg which calls Clean
or a lot of its functions.

As a side note, for future improvements on loadFiles, it may be worth
it to skip on some filepath functions. Per pprof analysis, 20% of CPU
time is still spent on Clean functions which are automatically called.

I've also taken the chance to refactor loadFiles into something
more readable IMO. This does add a bit of code dupping (specifically,
shouldIgnore and meta.Err checks), but the different handlings
are more explicit.

I've tried this on GCP e2-standard-8 machines and these changes make
Compute a Merkle Tree for an entire chrome build take about ~50% of
the time (~50s instead of ~100s), working in a single thread. Weirdly,
changing pd-standard vs pd-ssd doesn't seem to make a difference.

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -4,7 +4,7 @@ package client
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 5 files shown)


rubensf:
I haven't had the chance to properly perf test this in isolation unfortunately, but I find it very unlikely it'd get any slower.

atetubou:
Could you at least provide benchmark test?
I don't think it is good idea to have additional dependency if it improvement is not visible.

atetubou (changes requested):
Could you at least provide benchmark test?
I don't think it is good idea to have additional dependency if it improvement is not visible.

rubensf:
Thanks, I've ran some benchmarks on linux and this does substantially improve performance. I noticed I didn't need the additional dependency.

go/pkg/client/tree.go
```diff
@@ -87,115 +87,120 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
...
-func getTargetRelPath(execRoot, symlinkNormDir, target string) (string, error) {
-	symlinkAbsDir := filepath.Clean(filepath.Join(execRoot, symlinkNormDir))
-	if !filepath.IsAbs(target) {
-		target = filepath.Clean(filepath.Join(symlinkAbsDir, target))
-	}
-	if _, err := getRelPath(execRoot, target); err != nil {
-		return "", err
+func getRelPath(base, path string) (string, error) {
+	rel, err := filepath.Rel(base, path)
+	if err != nil || strings.HasPrefix(rel, "..") {
```
atetubou, line 92: better to return different error if err != nil?

atetubou (changes requested):
I meant to ask write benchmark using
https://golang.org/pkg/testing/#hdr-Benchmarks

go/pkg/client/tree.go
```diff
@@ -226,13 +231,12 @@ func (c *Client) ComputeMerkleTree(execRoot string, is *command.InputSpec, cache
...
 	}
-	for _, i := range is.Inputs {
-		if i == "" {
-			return digest.Empty, nil, nil, errors.New("empty Input, use \".\" for entire exec root")
-		}
-		if e := loadFiles(execRoot, is.InputExclusions, i, fs, cache, c.TreeSymlinkOpts); e != nil {
-			return digest.Empty, nil, nil, e
-		}
+
+	if len(is.Inputs) == 1 && is.Inputs[0] == "" {
```
atetubou, line 235: is it fine to check this only when `len(is.Inputs) == 1`?

rubensf (commented):
Thanks! I didn't know about golang benchmarks. 

This sample seems to show an improvement of from ~100k ns/op to ~80k ns/op.

go/pkg/client/tree_test.go
```diff
@@ -1573,3 +1575,47 @@ func TestComputeOutputsToUploadDirectories(t *testing.T) {
...
 }
+
+func randomBytes(n int) []byte {
+	b := make([]byte, n)
+	rand.Read(b)
+	return b
+}
+
+func BenchmarkComputeMerkleTree(b *testing.B) {
+	rand.Seed(time.Now().UnixNano())
```
atetubou, line 1586: better to use fixed seed?

go/pkg/client/tree_test.go
```diff
@@ -1573,3 +1575,47 @@ func TestComputeOutputsToUploadDirectories(t *testing.T) {
...
+	for i := 0; i < b.N; i++ {
+		fmc := filemetadata.NewSingleFlightCache()
+		_, _, _, err := e.Client.GrpcClient.ComputeMerkleTree(e.ExecRoot, inputSpec, fmc)
+		if err != nil {
+			b.Errorf("Failed to compute merkle tree: %v", err)
+		}
+	}
+	b.StopTimer()
+
+	for _, f := range inputSpec.Inputs {
```
atetubou, line 1618: remove this and write files under b.TempDir()?
https://golang.org/pkg/testing/#B.TempDir

atetubou (changes requested):
Could you use
https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
to show improvement with and without change?

go/pkg/client/tree_test.go
rubensf: Done

go/pkg/client/tree_test.go
rubensf: Ah, `NewTestEnv` already uses a TempDir. Done.

rubensf (commented):
Did the benchstat:
```
$ benchstat old.txt new.txt
name                  old time/op  new time/op  delta
ComputeMerkleTree-12   101µs ± 2%    85µs ± 3%  -15.80%  (p=0.000 n=10+10)
```

go/pkg/client/tree_test.go
atetubou: not yet removed?

rubensf: My bad, missed the commit when testing the benchmark before & after. Fixed.

go/pkg/client/tree.go
```diff
@@ -87,115 +87,127 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
...
+				filesToProcess = append(filesToProcess, targetExecRoot)
+			}
+		case meta.IsDirectory:
+			if shouldIgnore(absPath, command.DirectoryInputType, excl) {
+				continue
+			} else if meta.Err != nil {
+				return meta.Err
+			}
+
+			f, err := os.Open(absPath)
```
atetubou, line 181: close this?

rubensf: done

atetubou: I think it is better to close earlier.
This may cause 'too many open file' error.

Also why not use ioutil.ReadDir?

go/pkg/client/tree_test.go
```diff
@@ -1573,3 +1574,40 @@ func TestComputeOutputsToUploadDirectories(t *testing.T) {
 		})
 	}
 }
+
+func randomBytes(n int) []byte {
```
atetubou, line 1578: nit: I think it is better to use function local rand state in benchmark.

atetubou (commented):
Could you run bench with -count 5 and include summary from benchstats in commit message?

go/pkg/client/tree.go
rubensf: Fixed to close it earlier.

See my PR description - ioutil.ReadDir isn't efficient as it'll call Stat for files and then unnecessarily sort the results. Besides, ioutil is generally being discouraged from usage: https://golang.org/pkg/io/ioutil/

go/pkg/client/tree_test.go
rubensf: I'm not sure what you mean?

rubensf (commented):
Done

go/pkg/client/tree_test.go
atetubou: I meant to use https://golang.org/pkg/math/rand/#Rand

go/pkg/client/tree.go
atetubou: Stat is not used in ReadDir?
https://golang.org/src/io/ioutil/ioutil.go?s=2125:2176#L48

And overhead of sort is ignorable here?

rubensf: If you follow ReadDir's code, `f.Readdir` then calls `readdir(-1, readdirFileInfo)`. The code then calls `lstat`. `readdirname` follows a similar path but doesn't call stat. The sort cost should be minimal but it's still unnecessary wasteful while easy to skip. I'm not sure what's the issue using `ReadDirnames`?

atetubou: Ah, I see. You are right.

go/pkg/client/tree_test.go
rubensf: Done

go/pkg/client/tree_test.go
```diff
@@ -1573,3 +1575,41 @@ func TestComputeOutputsToUploadDirectories(t *testing.T) {
...
+	b := make([]byte, n)
+	randGen.Read(b)
+	return b
+}
+
+func BenchmarkComputeMerkleTree(b *testing.B) {
+	e, cleanup := fakes.NewTestEnv(b)
+	defer cleanup()
+
+	randGen := rand.New(rand.NewSource(time.Now().Unix()))
```
atetubou, line 1589: better to use fixed seed

go/pkg/client/tree.go
```diff
@@ -87,115 +87,128 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
...
 
-// getTargetRelPath returns the part of the target relative to the symlink's
-// directory, iff the target is under execRoot. Otherwise it returns an error.
-func getTargetRelPath(execRoot, symlinkNormDir, target string) (string, error) {
-	symlinkAbsDir := filepath.Clean(filepath.Join(execRoot, symlinkNormDir))
-	if !filepath.IsAbs(target) {
-		target = filepath.Clean(filepath.Join(symlinkAbsDir, target))
-	}
-	if _, err := getRelPath(execRoot, target); err != nil {
+func getRelPath(base, path string) (string, error) {
```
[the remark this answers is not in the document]
rubensf: That's a good point -- I think we could save some time by eg explicitly stating that the inputs should be relative to exec root (`command.InputSpec` isn't explicit). 

Note that, at this point I've already trimmed all the explicit `Clean` calls, but filepath will `Clean` most of its other calls (we do a nontrivial amount of `Rel` and `Join`) that I'm hoping to also trim eventually.

In any case, should we punt that to a later CL? This one is already reasonably large.

go/pkg/client/tree_test.go
rubensf: Done

go/pkg/client/tree.go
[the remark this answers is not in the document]
ramymedhat: sgtm

ramymedhat (approved):

atetubou (approved):

atetubou:
It seems that this caused regression in upload performance on linux builder used for chromium.
See comments around https://crbug.com/1193375#c47
Could you revert this?

rubensf:
Is it possible there are other culprits here? Can we do experiments with Merkle Tree calculated?

I'm confused on how this change would have all the impact you see on the crbug. This change is for the most part removing quite a few unnecessary calls. I tested this quite a bit on the chrome source tree and with the go benchmarks to see noticeable improvements (~2x on chrome source). If this change makes some tree computations worse, I want to investigate why.

atetubou:
I still don't know there are other culprits, but we'll use nodir's implementation soon. So we'll leave this as-is.

PR #299, opened 2021-04-20
[cas] Add cas package skeleton

nodirg:
[cas] Add cas package skeleton

Add cas.Client with a structure, but without actual implementation.
The TODOs will be implemented in subsequent CLs and in parallel.

Unlike client package, NewClient() function in this package does not
option. Rationale:
- the "options" approach pollutes the godoc page: a new type,
  its Apply() function, the DefaultXXX constant.
  In contrast, a simple Config struct is much more compact and has
  minimal footprint on the number of symbols in the godoc page.
- GCP libraries use the approach used in this package
  - https://pkg.go.dev/cloud.google.com/go/spanner#NewClientWithConfig
  - https://pkg.go.dev/cloud.google.com/go/bigtable#NewClientWithConfig
  - the options approach is used only for common options

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -1,4 +1,4 @@
-// Package cas implements efficient client for Content Addressable Storage.
+// Package cas implements an efficient client for Content Addressable Storage.
 package cas
 
 import (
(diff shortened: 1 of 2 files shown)


nodirg:
created with wrong email address again

PR #300, opened 2021-04-20
[cas] Add cas package

nodirg:
Add cas.Client with a structure, but without actual implementation.
The TODOs will be implemented in subsequence CLs and in parallel.

Unlike client package, NewClient() function in this package does not
option. Rationale:
- the "options" approach pollutes the godoc page: a new type,
  its Apply() function, the DefaultXXX constant.
  In contrast, a simple Config struct is much more compact and has
  minimal footprint on the number of symbols in the godoc page.
- GCP libraries use the approach used in this package
  - https://pkg.go.dev/cloud.google.com/go/spanner#NewClientWithConfig
  - https://pkg.go.dev/cloud.google.com/go/bigtable#NewClientWithConfig
  - the options approach is used only for common options

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -9,7 +9,6 @@ import (
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 	"google.golang.org/grpc"
 
-	"github.com/bazelbuild/remote-apis-sdks/go/pkg/client"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 )
 
@@ -20,13 +19,13 @@ import (
 //
 // All fields are considered immutable, and should not be changed.
 type Client struct {
+	conn *grpc.ClientConn
 	// The full name of the RBE instance.
 	InstanceName string
 
 	// The configuration that the client was created with.
 	ClientConfig
 
-	conn       *grpc.ClientConn
 	byteStream bspb.ByteStreamClient
 	cas        repb.ContentAddressableStorageClient
 }
@@ -34,16 +33,6 @@ type Client struct {
 // ClientConfig is a config for Client.
 // See DefaultClientConfig() for the default values.
 type ClientConfig struct {
-	// The connection to RBE to use.
-	//
-	// WARNING: setting this field overrides DialParams completely - DialParams
-	// will be ignored.
-	Conn *grpc.ClientConn
-
-	// Parameters to create a connection.
-	// Ignored if Conn is not nill.
-	DialParams *client.DialParams
-
 	// TODO(nodir): add conifg values here.
 	// TODO(nodir): add per-RPC timeouts.
 	// TODO(nodir): add retries.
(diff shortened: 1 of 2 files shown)


nodirg:
@ramymedhat PTAL

go/pkg/cas/client.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+// NewClient creates a new client with the default configuration.
+func NewClient(ctx context.Context, instanceName string) (*Client, error) {
+	return NewClientWithConfig(ctx, instanceName, DefaultClientConfig())
+}
+
+// NewClientWithConfig creates a new client and accepts a configuration.
+func NewClientWithConfig(ctx context.Context, instanceName string, config ClientConfig) (*Client, error) {
```
rubensf, line 73: Should we also have a constructor that takes a `grpc.ClientConn`?

I imagine the long-term vision for this package is for `go/pkg/client` to use it as a dependency. In which case I think it may be better to just downstream that connection rather than create a new one.

nodirg: Why not accept the connection via Config, since it is an optional param?

rubensf: That also sounds good to me (and maybe add some comments and/or error checking that validates that setting ClientConfig.Conn is incompatible with setting `DialParams`).

go/pkg/cas/upload.go
```diff
@@ -0,0 +1,42 @@
...
+import "context"
+
+// UploadInput is one of inputs to Client.Upload function.
+//
+// It can be either a reference to a file/dir (see Path) or it can be an
+// in-memory blob (see Content).
+type UploadInput struct {
+	// Path to the file or a directory to upload.
+	// If empty, the Content is uploaded instead.
+	Path string
```
ramymedhat, line 12: Can we set expectations around what type of path this will be? abs or relative to cwd?

ramymedhat (commented):
I agree with using Config structs as a more go idiomatic approach, but would like this to be consistent across the different client packages. Not for this PR, but once the exec client is migrated to use the new cas package, would be nice to also migrate it to use config structs.

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/client.go
nodirg: Added Conn. Clarified in the comments that Conn overrides DialParams.

Making the two fields mutually exclusive is a bit awkward:
1. Callsite with an existing connection would have to write

  ```go
    cfg := cas.DefaultClientConfig()
    cfg.DialParams = nil  // the awkward part
    cfg.Conn = myConn
    client, err := cas.NewClientWithConfig(ctx, instance, cfg)
  ```

2. It requires `ClientConfig.DialParams` to be a pointer, but it is not a big deal. WDYT?

nodirg:
note: when the time comes to reuse `cas` package in `client`, we would need to move some common parts from `client` to a common package (to avoid circular dependency), for example `Dial`.

go/pkg/cas/client.go
ramymedhat: I would prefer the current version where Conn overrides DialParams. Maybe log a warning message if both are set?

nodirg: Ramy, just to make sure we are on the same page, do you prefer the callsite having to do both 1) set `Conn` and 2) unset `DialParams`; in order to avoid a warning? See my message right above for more detials.

ramymedhat: Ah sorry I missed the options above. Yes, 1 is awkward, perhaps we can have a DefaultWithDialParams or DefaultDialParams?

nodirg: Could you elaborate? Do you mean three symbols

- `DefaultClientConfigWithParams` which returns a fully-functioning config with `DialParams` and without `Conn`
- `DefaultDialParams` used by above
- `DefaultClientConfig` which returns a non-really-functioning client config because it has neither `DialParams` nor `Conn`? It is primarily for those who want to set `Conn`.

This seems a bit awkward to me as well, in particular the fact that a single attribute in the config causes symbol explosion at the package level.

Maybe it is worth to step back a bit to think about the problem we want to solve. FWIU it is that a programmer might get surprised that `DialParams` is ignored. I wonder if this can be addressed with a louder comment? Sort of "warning: if you override Conn, then DialParams is ignored" or something.

ramymedhat: I meant either  `DefaultDialParams` + `DefaultClientConfig` or `DefaultClientConfigWithParams` + `DefaultClientConfig`. In the first case one would:
```
cfg := cas.DefaultClientConfig()
cfg.DialPararms = cas.DefaultDialParams()
```

Yes that is the concern. The user could override some `DialParams` for CAS versus Exec, but accidentally sharing the connection would silently override the changes. I guess since overriding the connection will be explicit, the user should be aware of what they're doing. So a comment sounds sufficient to me.

ramymedhat:
> note: when the time comes to reuse `cas` package in `client`, we would need to move some common parts from `client` to a common package (to avoid circular dependency), for example `Dial`.

Ack.

go/pkg/cas/client.go
nodirg: > I guess since overriding the connection will be explicit, the user should be aware of what they're doing

yes, I implied this. Sorry for not making this explicit.

I took a stab at a louder version of the comment. PTAL.

ramymedhat (approved):
Thanks!

go/pkg/cas/client.go
```diff
@@ -0,0 +1,110 @@
...
+	if conn == nil {
+		var err error
+		if conn, err = client.Dial(ctx, config.DialParams.Service, config.DialParams); err != nil {
+			return nil, errors.Wrap(err, "failed to dial")
+		}
+	}
+
+	client := &Client{
+		InstanceName: instanceName,
+		ClientConfig: config,
```
rubensf, line 102: Reading this, we're exposing `DialParams` even if we set both `conn` and `DialParams`. Even if we don't yell a warning, we likely should unset `DialParams` if it's exported.

nodirg: PTAL. I had to make `DialParam` a pointer in order to be able to unset it.
It has caused additional complexity in `Validate` and `NewClientWithConfig` though.

Is this what you meant?

nodirg: Frankly, I am not sure it is better. Now the code reader might think that `DialParams` and `Conn` are mutually exclusive because it is possible to not have `DialParams`.

rubensf: Aren't `DialParams` and `Conn` mutually exclusive? Only one of them should ever be used. 

I'm not sure there's an option to improve both simplicity and correctness. And I'm not convinced that thorough checking of mixes of options that _do_ affect each other is bad. I think it's generally better to have an explicit error than a louder comment, particularly as developers work on more complicated codebases with dependency chains.

Granted, as you mentioned in the other thread, this set of configurations might be changed/migrated substantially (maybe we'll migrate to a hopefully tiny "ConnectionConfig" struct used by CAS and Exec?), so if you think the current setup is confusing/complex, I'm happy to look at this again once the structure is more stable.

rubensf (approved):

go/pkg/cas/client.go
nodirg: yes, `DialParams` and `Conn` are mutually exclusive.
yes, checking mixes of options that do affect each other *is* bad.
And I agree that an explicit error is better than a comment.
I think we all agree on these, the question was mostly about the chosen balance.

Regardless, I've removed `DialParams` and made `conn` a required param. I think it is fine both use cases:
- It is certainly easier to the `client` package to use `cas` package now (modulo the circular dep)
- Not a big deal for other packages. They just create a connection with `client.Dial` and then call `cas.NewClient` with the conneciton. Seems fine.

and it addresses all the contention points in this PR. PTAL.

rubensf (approved):

ramymedhat (approved):

PR #302, opened 2021-04-21
[cas] Implement file reading

nodirg:
This commit implements concurrent FS walk and reading of regular files
(except large files). After reading, the files are scheduled for a
presence check, which is not implemented in this CL.
Symlink support TBD.


go/pkg/cas/client.go
```diff
@@ -28,12 +28,32 @@ type Client struct {
...
+	// Mockable functions.
+
+	testScheduleCheck func(ctx context.Context, item *uploadItem) error
 }
 
 // ClientConfig is a config for Client.
 // See DefaultClientConfig() for the default values.
 type ClientConfig struct {
-	// TODO(nodir): add conifg values here.
+	// Maximum number of concurrent file system operations.
```
atetubou, line 40: better to follow
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences
?

go/pkg/cas/client.go
```diff
@@ -45,12 +65,39 @@ type ClientConfig struct {
 //   ... mutate cfg ...
 //   client, err := NewClientWithConfig(ctx, cfg)
 func DefaultClientConfig() ClientConfig {
-	return ClientConfig{}
+	return ClientConfig{
+		// GCE docs recommend at least 32 concurrent IOs.
+		// https://cloud.google.com/compute/docs/disks/optimizing-pd-performance#io-queue-depth
+		FSConcurrency: 64,
```
atetubou, line 71: why choose 64 then?

go/pkg/cas/upload.go
```diff
@@ -17,11 +36,18 @@ type UploadInput struct {
...
+	//
+	// The function participates in a cache key, so try to reuse the same functions.
+	Predicate FilePredicate
+
 	// TODO(nodir): add AllowDanglingSymlinks
 	// TODO(nodir): add PreserveSymlinks.
 }
 
+// FilePredicate is a condition for a file/dir.
+type FilePredicate func(absName string, mode os.FileMode) bool
```
atetubou, line 49:
```suggestion
type FilePredicate func(absPath string, mode os.FileMode) bool
```

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+	eg.Go(func() error {
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case in, ok := <-inputC:
+				if !ok {
+					return nil
+				}
+				u.eg.Go(func() error {
```
atetubou, line 89: Is it fine to use eg here?

```suggestion
				eg.Go(func() error {
```

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+// Special care is taken for large files: they are read sequentially, opened
+// only once per file, and read with large IO size.
+type uploader struct {
+	*Client
+	eg    *errgroup.Group
+	stats TransferStats
+
+	// already-processed files.
+	// Types fsCacheKey and fsCacheValue are used for keys/values respectively.
+	fsCache sync.Map
```
atetubou, line 114: https://golang.org/pkg/sync/#Map says
```
The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
```

Can we start from map with mutext?

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
+	if in.Path == "" {
+		// Simple case: the blob is in memory.
+		return u.scheduleCheck(ctx, uploadItemFromBlob("", in.Content))
+	}
+
+	// Compute the absolute path only once per directory tree.
+	absPath, err := filepath.Abs(in.Path)
+	if err != nil {
+		return err
```
atetubou, line 133: better to wrap error if it come from standard library?

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+	}
+
+	// Do not use os.Stat() here. We want to know if it is a symlink.
+	info, err := os.Lstat(absPath)
+	if err != nil {
+		return err
```
atetubou, line 139: ditto

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+		return nil
+	}
+
+	_, err = u.visitFile(ctx, absPath, info, in.Predicate)
+	return err
+}
+
+// fsCacheKey is the key type for uploader.fsCache.
+type fsCacheKey struct {
+	AbsName      string
```
atetubou, line 153:
```suggestion
	AbsPath      string
```

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+//    streamed via ByteStream.
+//    If the file is already present on the server, the ByteStream preempts
+//    the stream with EOF and WriteResponse.CommittedSize == Digest.Size.
+//    Rewinding helps locality: there is no delay between reading the file for
+//    the first and the second times.
+//    Only one large file is processed at a time because most GCE disks are
+//    network disks. Reading many large files concurrently appears to saturate
+//    the network and slows down the progress.
+//    See also ClientConfig.LargeFileThreshold.
+func (u *uploader) visitRegularFile(ctx context.Context, absName string, info os.FileInfo) (*repb.FileNode, error) {
```
atetubou, line 204:
```suggestion
func (u *uploader) visitRegularFile(ctx context.Context, absPath string, info os.FileInfo) (*repb.FileNode, error) {
```

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+		item.Open = func() (readSeekCloser, error) {
+			_, err := f.Seek(0, io.SeekStart)
+			return f, err
+		}
+		panic("not implemented")
+		// TODO(nodir): implement.
+	}
+
+	// Schedule a check and close the file (in defer).
+	// item.open will reopen the file.
```
atetubou, line 273: nit
```suggestion
	// item.Open will reopen the file.
```

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+	var mu sync.Mutex
+	dir := &repb.Directory{}
+	var subErr error
+	var wg sync.WaitGroup
+
+	// This sub-function exist to avoid holding the semaphore while waiting for
+	// children.
+	err := func() error {
+		if err := u.fsSem.Acquire(ctx, 1); err != nil {
+			return err
```
atetubou, line 294: better to wrap error from standard library?

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+		}
+		defer u.fsSem.Release(1)
+
+		f, err := os.Open(absName)
+		if err != nil {
+			return err
```
atetubou, line 300: ditto

go/pkg/cas/upload.go
```diff
@@ -17,11 +36,18 @@ type UploadInput struct {
 	// Ignored if Path is not empty.
 	Content []byte
 
-	// TODO(nodir): add Predicate.
+	// If not nil and returns false for a file, then it is not uploaded.
+	//
+	// The function participates in a cache key, so try to reuse the same functions.
```
atetubou, line 41: Why not have this in uploader instead?

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+// fsCacheKey is the key type for uploader.fsCache.
+type fsCacheKey struct {
+	AbsName      string
+	PredicatePtr uintptr
+}
+
+// fsCacheValue is the value type for uploader.fsCache.
+type fsCacheValue struct {
+	compute  sync.Once
+	dirEntry proto.Message // *repb.FileNode, *repb.DirectoryNode or *repb.SymlinkNode.
```
atetubou, line 160: Can we use node for this?
```suggestion
	node proto.Message // *repb.FileNode, *repb.DirectoryNode or *repb.SymlinkNode.
```

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+func uploadItemFromBlob(title string, blob []byte) *uploadItem {
+	item := &uploadItem{
+		Title:  title,
+		Digest: digest.NewFromBlob(blob).ToProto(),
+		Open: func() (readSeekCloser, error) {
+			return newByteReader(blob), nil
+		},
+	}
+	if item.Title == "" {
+		item.Title = fmt.Sprintf("blob %s", item.Digest.Hash)
```
atetubou, line 422: Better to show digest instead of hash?

go/pkg/cas/upload_test.go
```diff
@@ -0,0 +1,161 @@
...
+	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	"github.com/golang/protobuf/proto"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestFS(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+
+	absTestData, err := filepath.Abs(filepath.Join("testdata"))
```
atetubou, line 21:
```suggestion
	absTestData, err := filepath.Abs("testdata")
```

go/pkg/cas/upload_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	data, err := ioutil.ReadAll(r)
+	if err != nil {
+		panic(err)
+	}
+	return data
+}
+
+func inputChanFrom(inputs ...*UploadInput) chan *UploadInput {
+	inputC := make(chan *UploadInput, 1)
```
atetubou, line 155:
```suggestion
	inputC := make(chan *UploadInput, len(inputs))
```

go/pkg/cas/upload_test.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+		{
+			desc: "FilePredicate for root",
+			inputs: []*UploadInput{{
+				Path: filepath.Join("testdata", "root"),
+				Predicate: func(absName string, mode os.FileMode) bool {
+					return filepath.Base(absName) != "root"
+				},
+			}},
+			wantScheduledChecks: nil,
```
atetubou, line 101: no need to have this explicitly?

atetubou (changes requested):

atetubou:
cc @Jwata

go/pkg/cas/client.go
nodirg: Done

go/pkg/cas/client.go
nodirg: Changed to 32 and a TODO to tune it.

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: switched to mutex and add a TODO to consider sync.Mutex

ramymedhat: "The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow ...". I think syncMap is appropriate for this use case, but in general this seems very close to singleflightcache. Would it make sense to use the same package for the same functionality instead of having different implementations? We can address performance concerns in the singleflightcache itself and just use it here.

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload_test.go
nodirg: Done

go/pkg/cas/upload_test.go
nodirg: Done

go/pkg/cas/upload_test.go
nodirg: sure

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: > Does the use case of isolate have many cases of duplicate traversals of directories and use of input exclusions?

Yes. We often traverse the same directories, and there is almost always a generic regexp-based filter. 
I am no sure how to avoid this.

go/pkg/cas/upload.go
```diff
@@ -39,6 +65,361 @@ type DigestStat struct {
...
+		}
+		item := uploadItemFromBlob(absName, contents)
+		ret.Digest = item.Digest
+		return ret, u.scheduleCheck(ctx, item)
+	}
+
+	// It is a medium or large file.
+
+	// Compute the hash.
+	h := digest.HashFn.New()
```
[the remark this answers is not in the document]
nodirg: done

go/pkg/cas/upload.go
nodirg: @ramymedhat just to see if we are thinking about the same thing: do you mean that reusing a common functionality could save us 3 lines of code?

```go
	entryRaw, _ := u.fsCache.LoadOrStore(cacheKey, &fsCacheValue{})
	e := entryRaw.(*fsCacheValue)
	e.compute.Do(func() {
```

Is this over-generalization?

Also, FWIU the only usage of `singleflightcache` is for CAS, so if we switch to this package, do we still need `singleflightpackage`?

go/pkg/cas/upload.go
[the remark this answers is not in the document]
ramymedhat: Ah I see. Good to know.

I'm also concerned about the practicality of users using the same predicate function. Specially since the `client` package operates on the InputExclusions struct. So naturally it will be creating new functions when calling `cas` each time an execution comes in. Or it would maintain a cache of `InputExclusion` to predicate function.

Do you think it makes sense to defer the predicate cache until we start measuring performance? We may also get more insight once `client` starts using `cas`.  Alternatively we could put more restrictions on InputExclusions to  enforce there is at most one per input type, and then we can cache the actual regexs.

go/pkg/cas/upload.go
ramymedhat: My point is more about reusing the common lib to make sure whatever optimizations we do apply here and elsewhere. For instance, I think here you're doing one less lock which would be beneficial in the general implementation. Also with a central cache we can track its size and memory consumption, get cache hit rates and other stats.

ramymedhat: > Also, FWIU the only usage of singleflightcache is for CAS, so if we switch to this package, do we still need singleflightpackage?

We use the cache for other use cases besides file digest. We also use the file metadata cache outside of the SDK.

rubensf: +1 to what Ramy said. This can be useful outside of the SDK (to lookup metadata/digest checking). Enhancing `SingleFlightCache` (which likely can perform better) or eventually (in a later PR?) exporting this both sound good to me.

nodirg: > We also use the file metadata cache outside of the SDK.

Thanks, this didn't occur to me.

I am happy to improve singleflightcache in a separate CL (otherwise seems to be a scope creep). I just tried, but the semantics is not the same: `singleflightcache` avoids caching the error (if `valFn` errors out, then the value isn't stored), so the same trick doesn't work. `uploader` here doesn't need that, and thus can be simpler (just one map, with `once`s) and more efficient. Ideas?

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: Yes, it doesn't work on closures. I've removed predicate from this CL. We can revisit this in another CL.

For context, the performance [was already measured](https://bugs.chromium.org/p/chromium/issues/detail?id=1193375#c37).

`client` package isn't the only user of this package. In our package where we'd use `cas` package, we have a single regex per `UploadItem`, so it is easy to cache functions per regex. [Here is our code](https://chromium-review.googlesource.com/c/infra/luci/luci-go/+/2837288/1/client/cmd/isolate/lib/common.go#349).

Ramy, also could you clarify what exactly is the suggestion? If we don't have predicate in the cache key, then the code will become incorrect because one `UploadItem` may have one predicate, and another might have another. So their dir nodes shouldn't be the same. If you are suggesting to move the predicate from `UploadItem` to a parameter of `Upload()`, then it means separate uploads cannot reuse each other's cache.

> Do we have use cases besides file name regexes + file mode checks?
Chrome doesn't have use cases besides file name regexes.

go/pkg/cas/client.go
```diff
@@ -45,12 +66,40 @@ type ClientConfig struct {
...
 
 // Validate returns a non-nil error if the config is invalid.
 func (c *ClientConfig) Validate() error {
-	return nil
+	switch {
+	case c.FSConcurrency <= 0:
+		return fmt.Errorf("FSConcurrency must be positive")
+
+	case c.SmallFileThreshold <= 0:
+		return fmt.Errorf("SmallFileThreshold must not be negative")
```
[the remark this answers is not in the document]
nodirg: Updated this code. PTAL

go/pkg/cas/upload.go
```diff
@@ -39,6 +64,370 @@ type DigestStat struct {
...
+//
+// Special care is taken for large files: they are read sequentially, opened
+// only once per file, and read with large IO size.
+type uploader struct {
+	*Client
+	eg    *errgroup.Group
+	stats TransferStats
+
+	// muFsCache protects fsCache.
+	// TODO(nodir): consider sync.Map.
```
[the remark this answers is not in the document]
nodirg: added, although I don't see how it would help. There is only one place where we lock/unlock.

go/pkg/cas/upload.go
```diff
@@ -39,6 +64,370 @@ type DigestStat struct {
...
+	item := &uploadItem{
+		Title:  absPath,
+		Digest: ret.Digest,
+	}
+
+	if isLarge {
+		// Large files are special: locality is important - we want to re-read the
+		// file ASAP.
+		// Also we are not going to use BatchUploads anyway, so we can take
+		// advantage of ByteStream's built-in presence check.
```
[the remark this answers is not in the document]
nodirg: added a reference to the doc.

go/pkg/cas/ioutil.go
```diff
@@ -0,0 +1,39 @@
...
+	io.Reader
+	content []byte
+}
+
+func newByteReader(content []byte) *byteReader {
+	return &byteReader{Reader: bytes.NewReader(content), content: content}
+}
+
+func (r *byteReader) Seek(offset int64, whence int) (int64, error) {
+	if offset != 0 || whence != io.SeekCurrent {
```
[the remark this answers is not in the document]
nodirg: Done.

go/pkg/cas/upload.go
```diff
@@ -39,6 +64,370 @@ type DigestStat struct {
...
+
+					switch node := node.(type) {
+					case *repb.FileNode:
+						dir.Files = append(dir.Files, node)
+					case *repb.DirectoryNode:
+						dir.Directories = append(dir.Directories, node)
+					case *repb.SymlinkNode:
+						dir.Symlinks = append(dir.Symlinks, node)
+					default:
+						panic("impossible")
```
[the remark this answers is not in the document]
nodirg: added.

go/pkg/cas/client.go
```diff
@@ -20,20 +20,41 @@ import (
...
+
+	// SmallFileThreshold is a size threshold to categorize a file as small.
+	// Such files are buffered entirely (read only once).
+	SmallFileThreshold int64
+
+	// LargeFileThreshold is a size threshold to categorize a file as large. For
+	// such files, IO concurrency limits are much tighter and locality is
+	// prioritized: the file is read for the first and second times with minimal
+	// delay between
+	// the two.
```
rubensf, line 52: nit: no newline

go/pkg/cas/upload.go
```diff
@@ -39,6 +59,358 @@ type DigestStat struct {
...
+// visitation happens concurrently, using u.eg.
+func (u *uploader) visitDir(ctx context.Context, absPath string) (*repb.DirectoryNode, error) {
+	var mu sync.Mutex
+	dir := &repb.Directory{}
+	var subErr error
+	var wg sync.WaitGroup
+
+	// This sub-function exist to avoid holding the semaphore while waiting for
+	// children.
+	err := func() error {
```
rubensf, line 284: wdyt of moving this function into its own to reduce indentation? no strong feelings.

go/pkg/cas/upload.go
```diff
@@ -39,6 +59,358 @@ type DigestStat struct {
...
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return err
+			}
+
+			for _, info := range infos {
+				info := info
+				absChild := filepath.Join(absPath, info.Name())
```
rubensf, line 308: thinking: filepath functions _always_ call `Clean`, even when we know it isn't necessary (absPath should have been cleaned on `startProcessing`; AFAICT `info.Name()` should be as good as clean?). 

At the same time using `strings.Join(..., os.Separator)` feels awkward? But that's what filepath does: https://golang.org/src/path/filepath/path_unix.go. wdyt of using it directly?

go/pkg/cas/upload.go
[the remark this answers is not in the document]
rubensf: I meant on the error message :)

go/pkg/cas/upload.go
rubensf: I *think* those semantics aren't necessary. We might just need the stat metadata saved, which can be done at a later PR. @ramymedhat ?

rubensf (approved):
LGTM overall. Ramy and Takuto might have more thoughts.

go/pkg/cas/client.go
nodirg: done

go/pkg/cas/upload.go
nodirg: Added a new func. PTAL. `strings.Join` seems overkill here.

go/pkg/cas/upload.go
nodirg: I considered that, but it means I'd also have to carry all the state (mutex, dir, wg, etc) via params, or maybe a separate struct, so overall it gets more complicated.

go/pkg/cas/upload.go
```diff
@@ -22,6 +39,9 @@ type UploadInput struct {
 	// TODO(nodir): add PreserveSymlinks.
 }
 
+// FilePredicate is a condition for a file/dir.
```
atetubou, line 42: This is not used in this PR now.

nodirg: removed all predicate-related stuff from this CL

nodirg:
@rubensf idk how to respond to your resolved comment, but I've addressed it.

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: WDYM by this?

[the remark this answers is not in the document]
nodirg: Given that predicate support is removed from this CL, I assume this discussion is resolved. Unresolve otherwise.

atetubou (approved):

go/pkg/cas/upload.go
ramymedhat: Yeah I think we don't need that. Ok to defer to a later PR though.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
ramymedhat: Just responding to the last message:

> For context, the performance was already measured.
I meant overhead of applying the regex repeatedly on duplicate directories in practice. I couldn't find a specific test for that in the bug but I might've missed it.

My suggestions were to (1) make the cache contain an unfiltered list of items (files, subdirs, symlinks) and apply the predicate before uploading not before storing in cache. So the cache becomes effectively a ReadDir cache. Option (2) is to cache the regexes assuming we enforce there is a constant number of regexes in the InputExclusions slice. I agree we can't move the predicate to `Upload()`. But generally agree this makes sense in a separate PR.

ramymedhat (approved):

go/pkg/cas/upload.go
rubensf: I have a slight preference towards the many-arguments uglyness instead of 6 nesting levels uglyness. IMO it's easier to human parse the first than the second. But tbh this is pretty subjective, so I'm fine with whatever you think is best.

rubensf (approved):
FYI, if you want to reply to outdated comments, you'll have to go on commits -> click on hash of the commit where the comment was create -> add comment there.

Sadly, whenever you delete your branch after the PR is merged, checking those historical comments become pretty hard, since we squash the PRs and the intermediate comments aren't stored on github.

... github's interface is quite inferior to gerrit :(

go/pkg/cas/upload.go
nodirg: I agree that 6 sounds a bit too much, but in this case moving it out to a separate top-level function would save us only one level of indentation :)

go/pkg/cas/upload.go
nodirg: I sent out https://github.com/bazelbuild/remote-apis-sdks/pull/303, but I couldn't figure out how to rebase this PR on top of the new one (something that I do in gerrit all the time). So I am going to have to merge this PR, and then update https://github.com/bazelbuild/remote-apis-sdks/pull/303 to use `singleflightcache` in `cas` package.

PR #303, opened 2021-04-22
Simplify caching packages

nodirg:
Simplify and optimize singleflightcache implementation: replace two
sync.Maps and an N+1 sync.RWMutexes, with one sync.Map and N sync.Once.
Do other simplifications as well.

Remove dead code / impossible conditions, e.g. "unexpected type found in
lock map". Use that to simplify the function signatures, e.g. don't
return an error (and thus simplify callsites and tests) when an error
is impossible.

In filemetadata/cache.go, fix the bug that it resets all namespaces.

`cache.Cache.Reset` was inherently racy because it was overwriting a mutex (without holding another mutex).
Replace with safe ns-level resetting.

### DIFF
--- a/go/pkg/cache/cache.go
+++ b/go/pkg/cache/cache.go
@@ -25,11 +25,6 @@ func (c *Cache) Reset(ns string) {
 	c.caches.Delete(ns)
 }
 
-// ResetAll resets the cache.
-func (c *Cache) ResetAll() {
-	*c = Cache{}
-}
-
 func (c *Cache) getNSCache(ns string) *singleflightcache.Cache {
 	// Load first to avoid instantiating a new cache for LoadOrStore.
 	nsCache, ok := c.caches.Load(ns)
--- a/go/pkg/cache/cache.go
+++ b/go/pkg/cache/cache.go
@@ -2,7 +2,6 @@
 package cache
 
 import (
-	"fmt"
 	"sync"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/cache/singleflightcache"
@@ -10,31 +9,34 @@ import (
 
 // Cache is a cache backend.
 type Cache struct {
-	mu     *sync.RWMutex
-	caches *sync.Map
+	caches sync.Map
 }
 
-var (
-	instance *Cache
-	once     sync.Once
-)
+var instance Cache
 
 // GetInstance retrieves the singleton instance of the cache backend. This is useful in the
 // future to enforce memory bounds on the entire cache usage of a program.
 func GetInstance() *Cache {
-	once.Do(func() {
-		instance = &Cache{
-			mu:     &sync.RWMutex{},
-			caches: &sync.Map{},
-		}
-	})
-	return instance
+	return &instance
 }
 
-// Reset resets the cache.
-func (c *Cache) Reset() {
-	c.mu = &sync.RWMutex{}
-	c.caches = &sync.Map{}
+// Reset resets the cache in one namespace.
+func (c *Cache) Reset(ns string) {
+	c.caches.Delete(ns)
+}
+
+// ResetAll resets the cache.
+func (c *Cache) ResetAll() {
+	*c = Cache{}
+}
+
+func (c *Cache) getNSCache(ns string) *singleflightcache.Cache {
+	// Load first to avoid instantiating a new cache for LoadOrStore.
+	nsCache, ok := c.caches.Load(ns)
+	if !ok {
+		nsCache, _ = c.caches.LoadOrStore(ns, &singleflightcache.Cache{})
+	}
+	return nsCache.(*singleflightcache.Cache)
 }
 
 // LoadOrStore attempts to first read a value from the corresponding cache namespace. If no entry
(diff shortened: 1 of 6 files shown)


ramymedhat (approved):
Thanks for deleting a lot of code! :)

ramymedhat:
Did you run the tests with the race detector?

nodirg:
yep, all good

```
pkg $ (faster-cache)  go test -race ./...
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/actas     0.108s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/balancer  0.599s
?       github.com/bazelbuild/remote-apis-sdks/go/pkg/balancer/proto    [no test files]
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/cache     0.050s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/cache/singleflightcache   0.096s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/cas       0.154s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/chunker   0.307s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/client    92.812s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/command   0.147s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/digest    0.114s
?       github.com/bazelbuild/remote-apis-sdks/go/pkg/fakes     [no test files]
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/filemetadata      0.108s
?       github.com/bazelbuild/remote-apis-sdks/go/pkg/flags     [no test files]
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/moreflag  0.074s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/outerr    0.037s
?       github.com/bazelbuild/remote-apis-sdks/go/pkg/portpicker        [no test files]
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/reader    0.844s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/retry     0.044s
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/rexec     0.372s
?       github.com/bazelbuild/remote-apis-sdks/go/pkg/testutil  [no test files]
ok      github.com/bazelbuild/remote-apis-sdks/go/pkg/tool      0.337s
?       github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo        [no test files]
```

go/pkg/cache/singleflightcache/singleflightcache.go
```diff
@@ -5,86 +5,42 @@
...
 	store sync.Map
-	// locks is a map of locks per key. This map is used to ensure only one reader/writer of
-	// the key can have write access.
-	locks sync.Map
-	// mu is a mutex to ensure exclusive access to the locks map in the case of Delete.
-	mu sync.RWMutex
+}
+
+type entry struct {
+	compute sync.Once
```
atetubou, line 17: Is it possible to use https://pkg.go.dev/golang.org/x/sync/singleflight in Cache instead of having sync.Once here?

nodirg: https://github.com/golang/sync/blob/036812b2e83c/singleflight/singleflight.go#L46-L71 seems overkill?

atetubou: ack

ramymedhat, line 17: I wonder if sync.Once is actually better than RWMutex. In the [implementation](https://golang.org/src/sync/once.go), seems like if there are multiple concurrent readers trying to access the same cache key (which is common in the execution case as the first batch of actions attempt to digest common toolchain files), readers might all enter `doSlow()`, and serialize on a lock of the internal mutex. Whereas with an [RWMutex](https://golang.org/src/sync/rwmutex.go), once the writer finishes writing, a semaphore is released with the reader count, unblocking all readers at once.

ramymedhat: Also with `singleflight` we would need to ensure we don't call the `compute` function if the value is loaded from the cache, since `singleflight` only dedups in flight computations and not all computations.

nodirg: I never compared their performance, but `sync.Once` was created exactly for this use case (ensure that the func runs only once). `RWMutex` is a more general thing, i.e. supports more use cases, so (hand wavy) it _should_ be slower.

nodirg: more concretely, there is only one place here where we lock, so we'd have to do `Lock()` (as opposed to `RLock()`), which is the same thing as not using `RWMutex`.

go/pkg/cas/upload.go
```diff
@@ -64,8 +65,7 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats
 		Client: c,
 		eg:     eg,
 
-		fsCache: map[string]*fsCacheValue{},
-		fsSem:   semaphore.NewWeighted(int64(c.FSConcurrency)),
+		fsSem: semaphore.NewWeighted(int64(c.FSConcurrency)),
```
rubensf, line 68: initialize `fsCache: cache.getNSCache("filemetadata")`?

go/pkg/cache/cache.go
```diff
@@ -2,85 +2,54 @@
...
+func (c *Cache) Reset(ns string) {
+	c.caches.Delete(ns)
 }
 
-// LoadOrStore attempts to first read a value from the corresponding cache namespace. If no entry
-// is found, it will use the return value of the passed fn to store in the cache. Concurrent
-// callers of LoadOrStore on the same namespace and key will execute fn once. This is to avoid
-// costly redundant work to compute the value to store in cache.
-func (c *Cache) LoadOrStore(ns string, key interface{}, fn func() (interface{}, error)) (interface{}, error) {
+func (c *Cache) getNSCache(ns string) *singleflightcache.Cache {
```
rubensf, line 28: Maybe out of scope for your PR (though I'd love it if you do it here):

At this point we don't really need all the extra accessors (eg LoadOrStore). I think both `cache` and `singleflightcache` can be merged in the same file, and the only `cache` operations we need to keep are `Reset` and `GetNSCache`. Users can then run `myCache = caches.GetNSCache(ns).LoadOrStore`, or save the NS cache for efficiency.

At that point we can rename these `Cache` -> `CacheBag/CacheBundle/CacheSingleton/wtv`, and `SingleFlightCache` -> `Cache`.

rubensf (approved):

go/pkg/cas/upload.go
nodirg: Do you mean a process-global cache? This would mean that if new files were written and Upload is called again, those files might not be uploaded.

rubensf: Ah, I see. For our usage, the SDK code isn't long running - there's generally the assumption that files don't change mid build, so we do need some way of at least passing in a `fsCache` (and `Upload` can create a new one if none is provided?)

nodirg: I do think that we need `(absPath, mTime, size) => digest` cache, but it will come later.

ramymedhat: FYI the file metadata cache used to support invalidation, but we removed it due to the assumption Rubens mentioned.

https://github.com/bazelbuild/remote-apis-sdks/commit/6e74127154b6047b16f708f06ccfc46bb2a4172b#diff-a297d8d639b8a097cebe69978c28397a77fde5b4a4b6887205d0a12858bb2e8c

go/pkg/cache/cache.go
nodirg: I started doing it in this PR, but it is getting large (would have to restart the review), so I will do it in a separate PR.

rubensf (commented):
FYI you might want to `bazelisk run //:gazelle`

nodirg:
@rubensf https://github.com/bazelbuild/remote-apis-sdks/pull/308

but it updated only cas' BUILD.bazel. Your comment is in a PR about caching packages. Did you expect cache packages' BUILD files to change too? I am not sure how to achieve that.

rubensf:
My bad, I meant to put that comment on #306 which is failing CI.

We do also should update the CAS BUILD.bazel files, which - plot twist - probably weren't being CI'd yet, so #308 is welcome.

PR #304, opened 2021-04-22
[cas] Implement presence check

nodirg:
Implement batched checker that checks if a digest is present on the
server. Use https://pkg.go.dev/google.golang.org/api/support/bundler
to bundle digests together, while respecting bundle size limits.

Add google.golang.org/api dependency.


go/pkg/cas/client.go
```diff
@@ -54,6 +55,9 @@ type ClientConfig struct {
 	// FileIOSize is the size of file reads.
 	FileIOSize int64
 
+	// Maximum number of digests to check in a single FindMissingBlobs RPC.
```
atetubou, line 58: follow https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences ?

atetubou (approved):

go/pkg/cas/client.go
nodirg: done

go/pkg/cas/upload.go
```diff
@@ -373,6 +408,54 @@ func (u *uploader) scheduleCheck(ctx context.Context, item *uploadItem) error {
...
+}
+
+// check checks which items are present on the server, and schedules upload for
+// the missing ones.
+func (u *uploader) check(ctx context.Context, items []*uploadItem) error {
+	req := &repb.FindMissingBlobsRequest{
+		InstanceName: u.InstanceName,
+		BlobDigests:  make([]*repb.Digest, len(items)),
+	}
+	byDigest := make(map[string]*uploadItem, len(items))
```
rubensf, line 425: `byDigest` should key'd on `Digest`

rubensf (commented):
Also: there's an argument to make the `LargeFileThreshold` to actually be ~4MiB (or have a medium file threshold?). The maximum gRPC message size is ~4MiB, so any files larger than that (or bigger than what fits in a [batchupload](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L1701)) don't need to be checked as they'll be checked over the ByteStream call.

(github won't let me comment on lines far from the ones you're touching).

go/pkg/cas/upload.go
```diff
@@ -113,6 +136,11 @@ type uploader struct {
 	// muLargeFile ensures only one large file is read at a time.
 	// TODO(nodir): ensure this doesn't hurt performance on SSDs.
 	muLargeFile sync.Mutex
+
+	// checkBundler bundles digests that need to be checked for presence on the
+	// server.
+	checkBundler *bundler.Bundler
+	seenHashes   sync.Map
```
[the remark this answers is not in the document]
nodirg: I like the idea, so I've added a TODO because it is more than just making it a global variable:
- we use isolate client on long-running servers, so having a global cache without at least LRU sounds like a memory leak.
- once we make it more global, instance name and RBE service name must be incorporated into the cache key

go/pkg/cas/upload.go
```diff
@@ -68,8 +70,27 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats
...
 	}
 
+	// Initialize checkBundler, which checks if a blob is present on the server.
+	u.checkBundler = bundler.NewBundler(&uploadItem{}, func(items interface{}) {
+		// Handle errors and context cancelation via errgroup.
+		eg.Go(func() error {
+			return u.check(ctx, items.([]*uploadItem))
+		})
+	})
+	u.checkBundler.BundleCountThreshold = u.FindMissingBlobsBatchSize
```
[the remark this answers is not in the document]
nodirg: I've added comments why this can't happen. Basically digests are fairly small. We can hit this limit with over 100K digests, but that's too much for other reasons.

go/pkg/cas/upload.go
```diff
@@ -68,8 +70,27 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats
 		fsSem:   semaphore.NewWeighted(int64(c.FSConcurrency)),
 	}
 
+	// Initialize checkBundler, which checks if a blob is present on the server.
+	u.checkBundler = bundler.NewBundler(&uploadItem{}, func(items interface{}) {
+		// Handle errors and context cancelation via errgroup.
+		eg.Go(func() error {
```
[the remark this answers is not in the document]
nodirg: yes, the goal is stop everything when error happens. Also stop checking if another goroutine exited.

yeah, handling two contexts sounds fairly complicated. I am not sure how that would work.

I don't think we'd need two contexts. We can make a cancel func before creating errgroup, but then it would be more code, and it would be easier to breed a bug.

btw, it doesn't spin a new thread (at least in "OS thread" sense). Goroutines are pretty lightweight and Golang encourages to create them when needed https://screenshot.googleplex.com/5BcdW6eFhWkMYqs
We shouldn't shy away from creating goroutines when it helps.

go/pkg/cas/upload.go

nodirg:
@rubensf we already have small and large file thresholds. Any file between the two is medium, so we don't need one more threshold?

If `LargeFileThreshold` is only 4MiB, then all files of that size or more will be read and checked sequentially. Sequential reading is mandatory for large files. In my case, it is ~1GB files, but 4MiB is much less than that. Somewhere there is a threshold where sequential reading is not longer mandatory. It don't know what that number is (250MiB is made up. ), but it is more about disk reading, than about the size of the CAS request.

rubensf (commented):
Sorry, I think I was confusing. Large files should stay as they are, but there's one more condition that we can cover - I think there are four possible handlings depending on the file size (x):

* x < 1MiB -> load in memory, schedule check, schedule BatchUpdate
* 1MiB < x < ~4 MiB -> schedule check, schedule BatchUpdate
* **~4 MiB < x < 256 MiB (large)** -> this is bigger than a regular gRPC message maximum and/or bigger than the maximum that fits in a BatchUpload, so we _don't need_ to schedule a check. We can either schedule or immediatelly run a ByteStream Write.
* 256 MiB < x -> upload immediately, for locality.

The ~4 MiB threshold is min(gRPC max size, [RE Max batch size](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L1707)). If we just set the `LargeFileThreshold` to that, the code is simpler. But I'm also fine with scheduling an upload. But doing a check is just wasteful.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: not sure what happened. Added TODO.

nodirg:
@rubensf in my local implementation there are essentially 4 cases, but organized in the code differently. In the code, reading from disk is decoupled from uploading. The three categories of file sizes are about reading. The "above or below BatchUpload request limit" is about uploading and it is orthogonal to file reading ([code](https://chromium-review.googlesource.com/c/infra/luci/luci-go/+/2837288/1/client/rbe/cas/client.go#579)).

My initial reaction to your suggestion is that coupling the two might degrade performance (e.g. we'd be doing presence check sequentially in more cases than we have to), but I like the simplicity, so let's do it the way you suggest, and *then* see if decoupling the improved performance.

I'd prefer to do in a separate CL because it requires removing `ClientConfig.LargeFileThreshold`, hardcoding it to 4MiB, etc. It seems out of scope.

nodirg:
@rubensf I started working on a separate CL, but looks it is not going to work: the client might pass an `UploadItem` with `Contents` over 4MB, so we cannot avoid this [conditional piece of code](https://chromium-review.googlesource.com/c/infra/luci/luci-go/+/2837288/1/client/rbe/cas/client.go#577).

rubensf:
I confused by what you mean by coupling - I don't think that's what I'm suggesting? I'm suggesting either:
* One more condition at the end of `visitRegularFile` to call `scheduleUpload` directly (bypassing `scheduleCheck`); or
* Lowering `LargeFileThreshold` to catch anything over the Batch limit.

The conditional check on your linked `scheduleUpload` is a different point, I think you can move the check to `startProcessing`: 
```
if in.Path == "" {
  if len(in.Content) < BatchThresh { 
    scheduleCheck(in) 
  } else {
    scheduleUpload(in) 
  } 
}
```

rubensf:
Ah, I see what you mean by coupling now. The current state of the Github PR has comments on `isLarge` about bypassing Batch checks, while [client#L329](https://chromium-review.googlesource.com/c/infra/luci/luci-go/+/2837288/1/client/rbe/cas/client.go#329) does the check anyway. 

I _think_ that coupling is fine, unless we explicitly expose some "upload directly" interface, which doesn't seem like you're planning to do?

I'm fine doing that in a separate cl - the Large files check bypass isn't implemented in this one yet anyway.

nodirg:
re coupling: there are two things about large files
1) skipping FindMissingDigests, which is what you seem to be talking about
2) tight limits on concurrency, see `muLargeFile`. We read only one large file at a time.

I am talking about (2). We currently limit concurrency to 1-file-at-a-time only for 250MB+ files. Changing the threshold to 4MB would mean that all 4MB+ files are read sequentially. That's potentially *many more files to read sequentially* - and that's my concern.

nodirg:
FWIU your main idea is to skip `FindMissingBlobs` RPC for files over 4MB. The rationale is that such files wouldn't fit into `UploadBlobs` RPC anyway, so we could save one RPC by trying to stream the blob right away.

I think this works well if the files aren't present on the server: then we indeed save one FindMissingBlobs RPC per N digests that we check (up to 10k digests). 

However, it might hurt performance if the files are already present. In our case, smaller files are more likely to be already present on the server, and large files are more likely to be missing (because usually they are combined from smaller files, e.g. binaries). So I *think* that smaller files have a good chance to be already present on the server. 

Suppose we are uploading 100 5MB files, all of which are already present on the server. With the current impl, it would be a single RPC `FindMissingBlobs`. In the proposed implementation, it would be 100 RPCs. Also note that in the bytestream-based check, the first request usually contains the first chunk, so besides sending the digest, we would also send data unnecessarily. This is not a big deal for large (250MB+) files because one chunk is a drop in the bucket compared to the file size, but with 100 files, IOSize=1MB and threshold 4MB, it means we send 100MB unnecessarily. We can reduce the size of the initial chunk of course.

The latest comment proposes to add more conditional code. I think this complexity is justified if it improves performance. Whether it improves performance is currently unclear to me. Would you mind if land the current implementation, and then we can test if this additional complexity actually improves performance?

go/pkg/cas/upload_test.go
```diff
@@ -103,6 +105,76 @@ func TestFS(t *testing.T) {
...
+		{Hash: "18ac3e7343f016890c510e93f935261169d9e3f565436429830faf0934f4f8e4", SizeBytes: 1},
+		{Hash: "2e7d2c03a9507ae265ecf5b5356885a53393a2029d241394997265a1a25aefc6", SizeBytes: 1},
+		{Hash: "3e23e8160039594a33894f6564e1b1348bbd7a0088d42c4acb73eeaed59c009d", SizeBytes: 1},
+		{Hash: "ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb", SizeBytes: 1},
+	}
+	sort.Slice(gotDigestChecks, func(i, j int) bool {
+		return gotDigestChecks[i].Hash < gotDigestChecks[j].Hash
+	})
+	if diff := cmp.Diff(wantDigestChecks, gotDigestChecks); diff != "" {
+		t.Fatal(diff)
```
rubensf, line 158: Errorf? We technically can continue testing (or at least check if multiple things are wrong at the same time).

rubensf (approved):
Ah, I completely missed `muLargeFile`. My bad completely. Sorry!

You also have good points regarding performance.

go/pkg/cas/upload_test.go
nodirg: good point. done

nodirg:
Thanks for the review!

PR #305, opened 2021-04-23
client: deprecate DefaultRPCTimeout

atetubou:
Appropriate DefaultRPCTimeout depends how user uses this SDK.
So instead of provide that here, recommend user to specify their own
timeout if it is necessary.

context: http://b/183669037 and https://crbug.com/1189349#c19

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -618,7 +618,6 @@ func NewClientFromConnection(ctx context.Context, instanceName string, conn, cas
 		cas:                           regrpc.NewContentAddressableStorageClient(casConn),
 		execution:                     regrpc.NewExecutionClient(conn),
 		operations:                    opgrpc.NewOperationsClient(conn),
-		rpcTimeouts:                   DefaultRPCTimeouts,
 		Connection:                    conn,
 		CASConnection:                 casConn,
 		CompressedBytestreamThreshold: DefaultCompressedBytestreamThreshold,
@@ -665,6 +664,9 @@ func (d RPCTimeouts) Apply(c *Client) {
 	c.rpcTimeouts = map[string]time.Duration(d)
 }
 
+// DefaultRPCTimeouts are timeouts for each RPCs.
+//
+// Deprecated: specify appropriate timeout in each application.
 var DefaultRPCTimeouts = map[string]time.Duration{
 	"default":          20 * time.Second,
 	"GetCapabilities":  5 * time.Second,


go/pkg/client/client.go
```diff
@@ -618,7 +618,6 @@ func NewClientFromConnection(ctx context.Context, instanceName string, conn, cas
 		cas:                           regrpc.NewContentAddressableStorageClient(casConn),
 		execution:                     regrpc.NewExecutionClient(conn),
 		operations:                    opgrpc.NewOperationsClient(conn),
-		rpcTimeouts:                   DefaultRPCTimeouts,
```
Jwata, line 621: Why do you need to remove the default?  
The caller still can give RPCTimeouts an option. 
https://github.com/bazelbuild/remote-apis-sdks/blob/b63d37ddf3155e5a9fcbdc1ed642bf65d325758d/go/pkg/client/client.go#L659

rubensf: +1. I don't see why you'd want to remove the timeout when the user can already override it. Maybe you'd want to change the default, but I don't think we should remove it entirely.

rubensf (changes requested):

go/pkg/client/client.go
ramymedhat: +1, specially with the new `cas` package, makes sense for these defaults to stay put for the execution case.

atetubou: I rather want to know why you need to have its own timeout here.
This seems there from https://github.com/bazelbuild/remote-apis-sdks/blame/4df05f3a8575af96d1ece221e2222599174d1f5f/go/client/client.go#L65, but I was not able to get further context.

In which case retry due to timeout becomes success and it is valuable rather than using default timeout from user?
I think it is fine to have internal retry for some errors returned from server, but not sure for timeout.

rubensf: > This seems there from https://github.com/bazelbuild/remote-apis-sdks/blame/4df05f3a8575af96d1ece221e2222599174d1f5f/go/client/client.go#L65, but I was not able to get further context.

I'm not sure what you mean? `rpcTimeout` is clearly used on `callWithTimeout`, which is just good gRPC practice. 

Notice that there are multiple timeouts here, each for a different type of call.

> In which case retry due to timeout becomes success

That is correct. There are many moving pieces from client to fetching/writing data to a remote repository, many of which are beyond RBE's control. Occasional blips in performance should be normal.

Notice that if you don't use any timeout, chances are that your connection could in theory be hanging forever without the user knowing why. I don't think you want the user to be staring at a no-progress screen.

> it is valuable rather than using default timeout from user?

It is valuable to have a default timeout. The user can _also_ specify a timeout.

> I think it is fine to have internal retry for some errors returned from server, but not sure for timeout.

Are you trying to say we should _fail_ on timeouts? Because if so, that's not what you'd be doing here by deleting the default. 

Can you mention why can't you overwrite the timeout in your code? I still don't understand that.

ramymedhat: More context: we noticed flakily hanging connections when developers build from home. Adding rpc timeouts allowed us to retry and recover from the flake in most cases. If the connection issue is consistent, we would exhaust retries and then fallback locally (or racing would take effect). The defaults seemed reasonable for the execution traffic we were generating but makes sense that they may not work for the isolate use case, in which case they could be overridden.

atetubou:
I'll remove timeout from our client instead.

PR #306, opened 2021-04-23
Simplify caching packages further

nodirg:
Instead of using a global hashtable of caches with a single key, use
a global variable of the that cache.

Remove contents of "cache" package, while leaving the
"singleflightcache" subpackage.

Fix a bug: fmCache.Reset() cleared caches in all instances of fmCache
instead of just that one. Make the function global and clealry document
that it invalidates the cache globally.

Fix a couple of bugs around usage of atomic package along the way.

Remove some dead code / impossible conditions (`fmCache.check()`) and simplify code accordingly.

Increase consistency by renaming a couple of variables.


nodirg:
In the next CL I think we could move/rename `singleflightcache.Cache` to `cache.SingleFlight`. WDYT?

go/pkg/cache/singleflightcache/singleflightcache.go
```diff
@@ -35,12 +38,23 @@ func (c *Cache) LoadOrStore(key interface{}, valFn func() (interface{}, error))
...
 
 // Delete removes a key from the cache.
 func (c *Cache) Delete(key interface{}) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 	c.store.Delete(key)
 }
+
+// Reset invalidates all cache entries.
+func (c *Cache) Reset() {
```
rubensf, line 56: @ramymedhat do we need this Reset somewhere? I couldn't find non-test use cases that we could probably tweak in a simpler manner.

rubensf (approved):
Renaming is a great idea to me! Thanks a lot.

PR LGTM overall.

rubensf:
FYI you might want to `bazelisk run //:gazelle` here :)

nodirg:
> FYI you might want to bazelisk run //:gazelle here :)

I did and nothing changed (other than `cas`, but that's a separate CL).
What do you expect to change? `go/pkg/cache/BUILD.bazel` is removed in this CL.

rubensf:
CI is erroring with `(16:41:08) ERROR: /workdir/go/pkg/filemetadata/BUILD.bazel:6:11: no such package 'go/pkg/cache': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.`, which filemetadata is depending on. Filemetadata shouldn't be depending on an deleted package... You did change the import on `cache.go#L8`.

Gazelle should be changing that, but I missed that filematadata/BUILD.bazel has a #gazelle:ignore tag, ugh.

Can you remove that line and run gazelle? `cache_posix_test.go` already has a `# +build !windows`.

nodirg:
@rubensf fixed.

rubensf (approved):

go/pkg/cache/singleflightcache/singleflightcache.go
nodirg: Happy to remove in another CL

PR #307, opened 2021-04-23
[cas] Implement batch upload

nodirg:
Partially implement scheduleUpload - for small blobs.
Use bundler.Bundler to bundle blobs together.

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"golang.org/x/sync/semaphore"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 	"google.golang.org/grpc"
 
@@ -28,13 +29,13 @@ type Client struct {
 	// ClientConfig is the configuration that the client was created with.
 	ClientConfig
 
-	byteStream bspb.ByteStreamClient
-	cas        repb.ContentAddressableStorageClient
+	byteStream          bspb.ByteStreamClient
+	cas                 repb.ContentAddressableStorageClient
+	semBatchUpdateBlobs *semaphore.Weighted
 
 	// Mockable functions.
 
-	testScheduleCheck  func(ctx context.Context, item *uploadItem) error
-	testScheduleUpload func(ctx context.Context, item *uploadItem) error
+	testScheduleCheck func(ctx context.Context, item *uploadItem) error
 }
 
 // ClientConfig is a config for Client.
@@ -60,6 +61,14 @@ type ClientConfig struct {
 	// FindMissingBlobsBatchSize is the maximum number of digests to check in a
 	// single FindMissingBlobs RPC.
 	FindMissingBlobsBatchSize int
+
+	// BatchUpdateBlobsConcurrency is the maximum number of BatchUpdateBlobs RPCs
+	// at the same time.
+	BatchUpdateBlobsConcurrency int
+
+	// BatchBlobsBatchSize is the maximum number of blobs to upload/download
+	// in a single BatchUpdateBlobs/BatchReadBlobs RPC.
+	BatchBlobsBatchSize int
 
 	// RetryPolicy specifies how to retry requests on transient errors.
 	RetryPolicy retry.BackoffPolicy
@@ -98,6 +107,13 @@ func DefaultClientConfig() ClientConfig {
 		FileIOSize: 4 * 1024 * 1024, // 4MiB
 
 		FindMissingBlobsBatchSize: 1000,
+
+		BatchUpdateBlobsConcurrency: 256, // arbitrary
+
+		// This is a suggested approximate limit based on current RBE implementation for writes.
+		// Above that BatchUpdateBlobs calls start to exceed a typical minute timeout.
+		// This default might not be best for reads though.
+		BatchBlobsBatchSize: 4000,
 
 		RetryPolicy: retry.ExponentialBackoff(225*time.Millisecond, 2*time.Second, retry.Attempts(6)),
 	}
@@ -152,9 +168,17 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
 		byteStream:   bspb.NewByteStreamClient(conn),
 		cas:          repb.NewContentAddressableStorageClient(conn),
 	}
+	client.init()
 
-	// TODO(nodir): Check capabilities.
+	// TODO(nodir): Check capabilities, optionally.
 	return client, nil
+}
+
+// init is a part of NewClientWithConfig that can be done in tests without
+// creating a real gRPC connection. This function exists purely to aid testing,
+// and is tightly coupled with NewClientWithConfig.
+func (c *Client) init() {
+	c.semBatchUpdateBlobs = semaphore.NewWeighted(int64(c.BatchUpdateBlobsConcurrency))
 }
 
 func (c *Client) withRetries(ctx context.Context, f func() error) error {
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -16,9 +16,11 @@ import (
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 	"google.golang.org/api/support/bundler"
+	"google.golang.org/grpc/status"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/cache"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 )
 
@@ -69,9 +71,12 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats 
 	}
 
 	// Initialize checkBundler, which checks if a blob is present on the server.
+	var wgChecks sync.WaitGroup
 	u.checkBundler = bundler.NewBundler(&uploadItem{}, func(items interface{}) {
+		wgChecks.Add(1)
 		// Handle errors and context cancelation via errgroup.
 		eg.Go(func() error {
+			defer wgChecks.Done()
 			return u.check(ctx, items.([]*uploadItem))
 		})
 	})
@@ -80,6 +85,20 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats 
 	// We might run into the request size limits only if we have >100K digests.
 	u.checkBundler.BundleCountThreshold = u.FindMissingBlobsBatchSize
 
+	// Initialize batchBundler, which uploads blobs in batches.
+	u.batchBundler = bundler.NewBundler(&repb.BatchUpdateBlobsRequest_Request{}, func(subReq interface{}) {
+		// Handle errors and context cancelation via errgroup.
+		eg.Go(func() error {
+			return u.uploadBatch(ctx, subReq.([]*repb.BatchUpdateBlobsRequest_Request))
+		})
+	})
+	// Limit the sum of sub-request sizes to (maxRequestSize - requestOverhead).
+	// 4MiB is the default gRPC request size limit.
+	// TODO: retrieve it from server's capabilities instead.
+	// Subtract 1KB to be on the safe side.
+	u.batchBundler.BundleByteLimit = 4*1024*1024 - int(marshalledFieldSize(int64(len(c.InstanceName)))) - 1000
+	u.batchBundler.BundleCountThreshold = c.BatchBlobsBatchSize
+
 	// Start processing input.
 	eg.Go(func() error {
 		// Before exiting this main goroutine, ensure all the work has been completed.
@@ -87,9 +106,9 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats 
 		// in a bundler.
 		defer func() {
 			u.wgFS.Wait()
-
-			// checkBundler can be flushed only after FS walk is done.
-			u.checkBundler.Flush()
+			u.checkBundler.Flush() // only after FS walk is done.
+			wgChecks.Wait()        // only after checkBundler is flushed
+			u.batchBundler.Flush() // only after wgChecks is done.
 		}()
 
 		for {
@@ -130,6 +149,7 @@ type uploader struct {
 
 	// fsSem controls concurrency of file IO.
 	// TODO(nodir): ensure it does not hurt streaming.
+	// TODO(nodir): move to Client struct.
 	fsSem *semaphore.Weighted
 	// muLargeFile ensures only one large file is read at a time.
 	// TODO(nodir): ensure this doesn't hurt performance on SSDs.
@@ -139,6 +159,9 @@ type uploader struct {
 	// server.
 	checkBundler *bundler.Bundler
 	seenDigests  sync.Map // TODO: consider making it more global
+
+	// batchBundler bundles blobs that can be uploaded using UploadBlobs RPC.
+	batchBundler *bundler.Bundler
 }
 
 // startProcessing adds the item to the appropriate stage depending on its type.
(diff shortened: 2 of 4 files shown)


go/pkg/cas/upload.go
```diff
@@ -441,12 +476,67 @@ func (u *uploader) check(ctx context.Context, items []*uploadItem) error {
...
+		bytesTransferred := int64(0)
+		digestsTransferred := int64(0)
+		var retriableErr error
+		req.Requests = req.Requests[:0] // reset for the next attempt
+		for _, r := range res.Responses {
+			if err := status.FromProto(r.Status).Err(); err != nil {
+				if !retry.TransientOnly(err) {
+					return err
+				}
+				// This error is retrieble. Save it to return later, and
```
atetubou, line 527:
```suggestion
				// This error is retriable. Save it to return later, and
```

go/pkg/cas/upload_test.go
```diff
@@ -105,33 +106,46 @@ func TestFS(t *testing.T) {
...
 	t.Parallel()
 	ctx := context.Background()
 
 	var mu sync.Mutex
 	var gotDigestChecks []*repb.Digest
-	var gotRequestSizes []int
-	var gotScheduleUploadCalls []*uploadItem
+	var gotDigestCheckRequestSizes []int
+	var gotUploadBlobReqs []*repb.BatchUpdateBlobsRequest_Request
+	failCBlob := true
```
atetubou, line 117: What is 'C'?

go/pkg/cas/upload_test.go
```diff
@@ -205,8 +224,13 @@ func inputChanFrom(inputs ...*UploadInput) chan *UploadInput {
 type fakeCAS struct {
```
atetubou, line 224: Can we use https://pkg.go.dev/github.com/bazelbuild/remote-apis-sdks@v0.0.0-20210423213014-847bca232f88/go/pkg/fakes#CAS instead?

go/pkg/cas/client.go
```diff
@@ -89,6 +96,12 @@ func DefaultClientConfig() ClientConfig {
 
 		FindMissingBlobsBatchSize: 1000,
 
+		BatchUpdateBlobsConcurrency: 256, // arbitrary
+
+		// This is a suggested approximate limit based on current RBE implementation.
```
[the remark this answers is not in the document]
nodirg: this was copied from https://github.com/bazelbuild/remote-apis-sdks/blob/2dacc28301fb5eb8af27adf40456cac474f6f0e6/go/pkg/client/client.go#L122.

I prefer not link that line because that code is likely to be replaced with this new code.

Also I suspect the limit is somewhere in RBE server impl, which is closed-source, so I guess that's why that line isn't referring to anything either.

go/pkg/cas/upload_test.go
nodirg: added a comment.

go/pkg/cas/upload_test.go
nodirg: I considered that, but I am not sure it would save a lot of code, and I am not sure how would I check exactly what requests were sent, on how would I return transient values. On the other hand, `fastCAS` seems pretty small.

go/pkg/cas/client.go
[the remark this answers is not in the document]
atetubou: Ack.

go/pkg/cas/upload_test.go
atetubou: I see.

go/pkg/cas/upload.go
```diff
@@ -441,12 +476,67 @@ func (u *uploader) check(ctx context.Context, items []*uploadItem) error {
...
+				// This error is retriable. Save it to return later, and
+				// save the failed sub-request for the next attempt.
+				retriableErr = err
+				req.Requests = append(req.Requests, reqMap[digest.NewFromProtoUnvalidated(r.Digest)])
+				continue
+			}
+			bytesTransferred += r.Digest.SizeBytes
+			digestsTransferred++
+		}
+		atomic.AddInt64(&u.stats.Batched.Bytes, bytesTransferred)
```
atetubou, line 536: If we use atomic package to update this value, better to make this private and provide accessor using atomic.Load*?
Or use package like https://pkg.go.dev/go.uber.org/atomic ?

go/pkg/cas/upload.go
```diff
@@ -441,12 +476,67 @@ func (u *uploader) check(ctx context.Context, items []*uploadItem) error {
...
+		reqMap[digest.NewFromProtoUnvalidated(r.Digest)] = r
+	}
+
+	req := &repb.BatchUpdateBlobsRequest{
+		InstanceName: u.InstanceName,
+		Requests:     reqs,
+	}
+	return u.retry(ctx, func() error {
+		// TODO(nodir): add per-RPC timeouts.
+		res, err := u.cas.BatchUpdateBlobs(ctx, req)
```
ramymedhat, line 513: Why not use https://github.com/bazelbuild/remote-apis-sdks/blob/fd877b05ba2ed611f6d34e711da05917b729eff9/go/pkg/client/client.go#L821? Or is this plan once we move things out of the client package?

nodirg: Because `client` package will eventually depend on `cas` package, so we cannot have `cas` package heavily depending on `client` package. Besides, `client` package is scoped to execution - will we still want `BatchUploadBlobs` in the client package?

go/pkg/cas/upload.go
```diff
@@ -68,12 +70,17 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats
 		eg:     eg,
 
 		fsSem: semaphore.NewWeighted(int64(c.FSConcurrency)),
+
+		semBatches: semaphore.NewWeighted(int64(c.BatchUpdateBlobsConcurrency)),
```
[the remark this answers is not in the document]
nodirg: yes, it is. If there are multiple callers to `Upload`, then they would have separate semaphores. Are you suggesting to move `semBatches` to `Client`?

go/pkg/cas/upload.go
```diff
@@ -82,16 +89,31 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats
...
+		// Handle errors and context cancelation via errgroup.
+		eg.Go(func() error {
+			return u.uploadBatch(ctx, subReq.([]*repb.BatchUpdateBlobsRequest_Request))
+		})
+	})
+	// Limit the sum of sub-request sizes to (maxRequestSize - requestOverhead).
+	// 4MiB is the default gRPC request size limit.
+	// TODO: retrieve it using u.conn.GetMethodConfig instead. Currently it would break tests.
+	// Subtract 1KB to be on the safe side.
+	u.batchBundler.BundleByteLimit = 4*1024*1024 - int(marshalledFieldSize(int64(len(c.InstanceName)))) - 1000
```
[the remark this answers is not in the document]
nodirg: Updated the TODO to retrieve from capabilities instead. Is this what you meant?

go/pkg/cas/upload.go
nodirg: This field is private until it is done updating. We don't update it after we publish it.

go/pkg/cas/upload.go
ramymedhat: I see. So I guess this question is premature, until we start using `cas` in `client` and then we will figure out how the retry policy is passed through and how RPC timeouts are set.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
ramymedhat: Thanks. Just to make sure we're on the same page, we should load from Capabilities if the `startup_capabilities` flag is set: https://github.com/bazelbuild/remote-apis-sdks/blob/2dacc28301fb5eb8af27adf40456cac474f6f0e6/go/pkg/flags/flags.go#L60

go/pkg/cas/upload.go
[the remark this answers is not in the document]
ramymedhat: +1, the current `client` package sets a client wide batch concurrency limit as well.

go/pkg/cas/upload.go
atetubou: I think that won't be matter in practice, but https://github.com/golang/go/issues/5045#issuecomment-252730563 says
> you shouldn't mix atomic and non-atomic accesses for a given memory word.

nodirg: yes, it says that and we indeed must follow those rules for concurrent reads and writes. But this is not the case here. 

To be be clear: uploader is a private struct. Other packages do not have access to it. Once we return the stat, we no longer write.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: It looks like that package defines a global flag. I don't think that would be OK for many applications that might want to use this package. For example, this would be surprising to see in our servers - seems invasive. In our CLI, we use a specific subcommand package, and we very rarely use global flags.

Would it be OK to have a boolean field in `ClientConfig` instead? The code already has a TODO on capabilities, so I it would be done as a part of that TODO

go/pkg/cas/upload.go
nodirg: the retry policy can be passed by setting `ClientConfig.RetryPolicy` today.
I expect RPC Timeouts to be similar, but it is not done yet.

go/pkg/cas/client.go
```diff
@@ -59,6 +58,14 @@ type ClientConfig struct {
 	// single FindMissingBlobs RPC.
 	FindMissingBlobsBatchSize int
 
+	// BatchUpdateBlobsConcurrency is the maximum number of BatchUpdateBlobs RPCs
+	// at the same time.
+	BatchUpdateBlobsConcurrency int
```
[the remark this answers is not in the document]
nodirg: I think we need to distinguish concurrency limits for batch and stream uploads. They are different. So how about `BatchUploadConcurrency`? 

> CASConcurrency

I am not sure what this means. This entire package is CAS. I think it should be a bit more specific.

---

generally, I tried to follow the existing pattern: the `FindMissingBlobsBatchSize` starts with the RPC name, and so does this field. It is not critical though.

go/pkg/cas/upload.go
```diff
@@ -82,16 +88,31 @@ func (c *Client) Upload(ctx context.Context, inputC <-chan *UploadInput) (stats
...
+	u.batchBundler.BundleByteThreshold = u.batchBundler.BundleByteLimit
```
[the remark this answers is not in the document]
nodirg: no need, removed.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: I've updated that TODO

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: done, and renamed to `semBatchUpdateBlobs` since now it is at the `Client` level and "batches" can mean anything.

Also add a TODO to move `semFS` for the same reasons.

go/pkg/cas/client.go
```diff
@@ -59,6 +58,14 @@ type ClientConfig struct {
...
+
+	// BatchUpdateBlobsBatchSize is the maximum number of blobs to upload in a
+	// single BatchUpdateBlobs RPC.
+	BatchUpdateBlobsBatchSize int
```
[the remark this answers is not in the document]
nodirg: Done, but should they have the same defaults though. The default value for this field was designed for writes. Now the same value is used for reads. This seems a bit arbitrary?

go/pkg/cas/upload.go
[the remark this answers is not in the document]
ramymedhat: Sure the `cas` package should not rely on a global flag, the current client package doesn't either. Adding a booolean and a TODO sgtm. Thanks!

go/pkg/cas/upload.go
ramymedhat: Ack.

go/pkg/cas/client.go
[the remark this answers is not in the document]
ramymedhat: +1 to separating stream and non-stream concurrency, they have material differences on a grpc client's performance. We currently separate them on a per connection [basis](https://github.com/bazelbuild/remote-apis-sdks/blob/fd877b05ba2ed611f6d34e711da05917b729eff9/go/pkg/client/client.go#L449), but having a knob here doesn't hurt and we can always abstract into one knob in the client package.

go/pkg/cas/upload.go
atetubou: I think copying value by return is not guaranteed to be valid operation too if atomic is used there?

ramymedhat (approved):

go/pkg/cas/client.go
```diff
@@ -142,11 +158,19 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
...
 
-	// TODO(nodir): Check capabilities.
+	// TODO(nodir): Check capabilities, optionally.
 	return client, nil
 }
 
+// init is a part of NewClientWithConfig that can be done in tests without
+// creating a real gRPC connection. This function exists purely to aid testing,
+// and is tightly coupled with NewClientWithConfig.
+func (c *Client) init() {
```
rubensf, line 170: This could be in the tests package?

go/pkg/cas/upload_test.go
```diff
@@ -105,35 +107,49 @@ func TestFS(t *testing.T) {
...
 	cas := &fakeCAS{
 		findMissingBlobs: func(ctx context.Context, in *repb.FindMissingBlobsRequest, opts ...grpc.CallOption) (*repb.FindMissingBlobsResponse, error) {
 			mu.Lock()
 			defer mu.Unlock()
 			gotDigestChecks = append(gotDigestChecks, in.BlobDigests...)
-			gotRequestSizes = append(gotRequestSizes, len(in.BlobDigests))
+			gotDigestCheckRequestSizes = append(gotDigestCheckRequestSizes, len(in.BlobDigests))
 			return &repb.FindMissingBlobsResponse{MissingBlobDigests: in.BlobDigests[:1]}, nil
 		},
+		batchUpdateBlobs: func(ctx context.Context, in *repb.BatchUpdateBlobsRequest, opts ...grpc.CallOption) (*repb.BatchUpdateBlobsResponse, error) {
```
rubensf, line 127: FWIW I think we might actually want to converge this with fakeCAS at some point (no need for now) - otherwise we'll get quite a bit of repeated code as you write more tests.

go/pkg/cas/upload_test.go
```diff
@@ -105,35 +107,49 @@ func TestFS(t *testing.T) {
...
 	t.Parallel()
 	ctx := context.Background()
 
 	var mu sync.Mutex
 	var gotDigestChecks []*repb.Digest
-	var gotRequestSizes []int
-	var gotScheduleUploadCalls []*uploadItem
+	var gotDigestCheckRequestSizes []int
+	var gotUploadBlobReqs []*repb.BatchUpdateBlobsRequest_Request
+	failCBlob := true // blob "c" is uploaded below.
```
rubensf, line 118: what about `failBlob := map[dg]error`?

go/pkg/cas/client.go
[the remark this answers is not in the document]
rubensf: I'm fine with one knob per RPC here :)

go/pkg/cas/client.go
nodirg: it is used in `NewClient`. Could you elaborate?

go/pkg/cas/upload_test.go
nodirg: see also the previous discussion with Takuto in this PR.

go/pkg/cas/upload_test.go
nodirg: could you elaborate? what do we gain from the additional complexity?

go/pkg/cas/upload_test.go
rubensf: Right - doesn't seem like it's important now. Let's see if we end up having to duplicate it.

go/pkg/cas/upload_test.go
rubensf: flexibility?
Tbh I'm thinking of the case where fakeCAS is reused. I guess as long as it's test specific, that doesn't matter.

go/pkg/cas/client.go
rubensf: ah, at first glance I thought this would be specific for tests. Nvm.

go/pkg/cas/upload.go
nodirg: I don't think so. Do you have links to some documentation to backup this up?

rubensf (approved):

go/pkg/cas/upload.go
nodirg: Seems like this is the only outstanding comment and I am pretty sure this code is correct. I am happy to fix it in another CL if it is not.

PR #308, opened 2021-04-23
[cas] Add BUILD.bazel

nodirg:
Add pkg/cas/BUILD.bazel

Also add bazelisk to the deps to ensure all developers have the same
toolchain.


nodirg:
Warning: I don't know what I am doing.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	cloud.google.com/go v0.65.0 // indirect
+	github.com/bazelbuild/bazelisk v1.8.0 // indirect
```
rubensf, line 7: Bazelisk is a simple wrapper to ensure users have the same _bazel_ version. While having the same `bazelisk` version might be useful, that shouldn't affect build performance/correctness/determinism. Usually there's a `.bazelrc` at the workspace's root specifying the .bazelversion that apparently we haven't bothered including on the SDK.

Bazelisk isn't an actual build dependency, so we shouldn't be adding it to `go.mod` :)

rubensf (changes requested):

go.mod
nodirg: got it. removed. ptal

rubensf:
Ah, see the buildkite error message (at the bottom of the PR).

Gazelle picked up that you're trying to use the bundler but we didn't add the new dependency to the bazel workspace. You'll want to add it to the [re-sdk-deps.bzl](https://github.com/bazelbuild/remote-apis-sdks/blob/master/remote-apis-sdks-deps.bzl#L9).

Also run `go mod tidy` - it'll likely tweak go.sum/mod to show it as a dependency.

nodirg:
It appears that bazel assumes its OK to symlink testData files. So tests now panic because symlink support isn't there (even if it was there, the output would be different). After googling for 30min, I couldn't figure out how to force Bazel to copy files. Do you know? This is my first time using Blaze.

Otherwise I might have to change the tests to create files with Go code.

rubensf:
It's... Not easy :) :sweat: 

There's https://pkg.go.dev/github.com/bazelbuild/rules_go/go/tools/bazel, but I think that only gives you the bazel path (with symlinks). You may have to follow each individual one.

Also keep in mind that on Windows, bazel doesn't create symlinks as by default only adm can create symlinks on windows, so you'll have to follow symlinks conditionally.

nodirg:
I've changed the tests to create new files in temp dir. Checks pass now.

go/pkg/cas/upload_test.go
```diff
@@ -19,22 +20,28 @@ func TestFS(t *testing.T) {
 	t.Parallel()
 	ctx := context.Background()
 
-	absTestData, err := filepath.Abs("testdata")
+	tmpDir, err := ioutil.TempDir("", "test")
```
atetubou, line 23: use https://golang.org/pkg/testing/#T.TempDir ?

nodirg: TIL. Nice. Done.

atetubou (approved):

rubensf (approved):

PR #309, opened 2021-04-23
[cas] Add support for Symlinks.

nodirg:
Add support for symlinks in the file traversal. Add options
PreserveSymlinks and AllowDanglingSymlinks.

Incorporate the options into the cache key, since they affect the
resulting repb.Directory objects.


go/pkg/cas/upload.go
```diff
@@ -163,26 +172,43 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
 		// Do not use os.Stat() here. We want to know if it is a symlink.
 		info, err := os.Lstat(absPath)
 		if err != nil {
-			return errors.Wrapf(err, "lstat failed")
```
atetubou, line 166: why remove this wrap?

go/pkg/cas/upload.go
```diff
@@ -358,24 +385,67 @@ func (u *uploader) visitDir(ctx context.Context, absPath string) (*repb.Director
...
 
-// visitSymlink converts a symlink to a SymlinkNode and schedules visitation
+// visitSymlink converts a symlink to a directory node and schedules visitation
 // of the target file.
-func (u *uploader) visitSymlink(ctx context.Context, absPath string) (*repb.SymlinkNode, error) {
-	// TODO(nodir): implement
-	panic("not supported")
+// If key.PreserveSymlinks is true, then returns a SymlinkNode, otherwise
+// returns the directory node of the target file.
+func (u *uploader) visitSymlink(ctx context.Context, key fsCacheKey) (interface{}, error) {
```
atetubou, line 405:
```suggestion
func (u *uploader) visitSymlink(ctx context.Context, key fsCacheKey) (proto.Message, error) {
```

go/pkg/cas/upload.go
nodirg: turns out lstat errors start with "lstat: ", so this didn't add much.

go/pkg/cas/upload.go
nodirg: this doesn't compile. I could fix those compilation errors by adding more code though if you prefer.

go/pkg/cas/upload.go
atetubou: I think stacktrace is helpful when showing this error?

go/pkg/cas/upload.go
atetubou: I prefer to have nallow scoped types if it can be done with few simple line changes.

go/pkg/cas/upload.go
```diff
@@ -37,9 +37,17 @@ type UploadInput struct {
 	// Ignored if Path is not empty.
 	Content []byte
 
+	// PreserveSymlinks specifies whether to preserve symlinks or convert them
+	// to regular files.
```
rubensf, line 41:
```suggestion
	// their underlying files.
```
?

go/pkg/cas/upload.go
rubensf: +1, trace is useful (or even more verbose wording to make it easier to find, eg "error while stating uploadinput")

nodirg: Added `errors.WithStack`

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
```diff
@@ -358,24 +388,67 @@ func (u *uploader) visitDir(ctx context.Context, absPath string) (*repb.Director
...
+// If key.PreserveSymlinks is true, then returns a SymlinkNode, otherwise
+// returns the directory node of the target file.
+func (u *uploader) visitSymlink(ctx context.Context, key fsCacheKey) (proto.Message, error) {
+	target, err := os.Readlink(key.AbsPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "os.ReadLink")
+	}
+
+	// Determine absolute and relative paths of the target.
+	var absTarget, relTarget string
```
rubensf, line 415: I just remembered when checking this `relTarget`: generally action inputs are restricted under the execution root (or input root) of an action. If we're preserving symlinks, and their targets are outside of the input root, they'll be effectively dangling. When creating the input (the `Tree`), the paths also have to be relative to the execution root.

I think we should have a ClientConfig that specifies an "ExecRoot", and if specified, we should be err-ing UploadInputs outside of the root on `startProcessing`, and failing on symlinks that target outside the root if danglings are not allowed. We should maybe err even if danglings are allowed, to prevent user error?

Note that technically the API usually refers to it as "Input Root", but we use "Exec Root" throughout the codebase... Which means we likely want to stick to that naming, even if "Exec" isn't necessarily pertinent to "CAS".

rubensf: Doing in a separate CL is probably a better idea, but leave a TODO.

nodirg: added a TODO.

go/pkg/cas/upload.go
atetubou: seems not added?

atetubou (commented):
fix presubmit?

go/pkg/cas/upload.go
nodirg: ah, I've added it to a different `Lstat` call. Now both have `WithStack`.

go/pkg/cas/upload.go
nodirg: ah, I've added it to a different Lstat call. Now both have WithStac

nodirg:
@atetubou fixed

go/pkg/cas/upload.go
rubensf: ?

go/pkg/cas/upload.go
```diff
@@ -37,9 +37,17 @@ type UploadInput struct {
 	// Ignored if Path is not empty.
 	Content []byte
 
+	// PreserveSymlinks specifies whether to preserve symlinks or convert them
+	// to regular files.
+	PreserveSymlinks bool
+
+	// AllowDanglingSymlinks specifies whether to upload dangling links or halt
+	// the upload.
```
rubensf, line 45:
```suggestion
	// the upload with an error.
```
?

go/pkg/cas/upload.go
```diff
@@ -163,30 +171,51 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
...
 }
 
+// fsCacheKey is a key in u.fsCache.
+//
+// The fsCacheKey is primarily the absolute path, but also configuration of
+// file traversal.
+type fsCacheKey struct {
+	AbsPath            string
+	PreserveSymlinks   bool
+	AllowDanglingLinks bool
```
rubensf, line 195: Do you guys have uses where you change these settings? If not we could make these client settings for simplicity? AFAIK we don't use these options, and I can't actually envision use cases where a client will choose different options for different uploads.

go/pkg/cas/upload.go
```diff
@@ -358,24 +388,69 @@ func (u *uploader) visitDir(ctx context.Context, absPath string) (*repb.Director
...
+	symlinkDir := filepath.Dir(key.AbsPath)
+	target = filepath.Clean(target) // target may end with slash
+	if filepath.IsAbs(target) {
+		absTarget = target
+		if relTarget, err = filepath.Rel(symlinkDir, absTarget); err != nil {
+			return nil, err
+		}
+	} else {
+		relTarget = target
+		absTarget = filepath.Join(symlinkDir, relTarget)
```
rubensf, line 425: `joinFilePathsFast`? `Dir` is already cleaned, and you `Clean` target on L417.

go/pkg/cas/upload.go
```diff
@@ -358,24 +388,69 @@ func (u *uploader) visitDir(ctx context.Context, absPath string) (*repb.Director
...
+	}
+	// TODO(nodir): add an option to return an error if a symlink taret outside of
```
rubensf, line 427:
```suggestion
	// TODO(nodir): add an option to return an error if a symlink target outside of
```

go/pkg/cas/upload.go
```diff
@@ -358,24 +388,69 @@ func (u *uploader) visitDir(ctx context.Context, absPath string) (*repb.Director
...
+	targetKey := key
+	targetKey.AbsPath = absTarget
+	if !key.PreserveSymlinks {
+		return u.visitFile(ctx, targetKey, targetInfo)
+	}
+
+	// Even though we return a SymlinkNode, we still must visit the target file
+	// to ensure it is uploaded.
+	_, err = u.visitFile(ctx, targetKey, targetInfo)
+	return symlinkNode, err
```
rubensf, line 453:
```suggestion
      node, err := u.visitFile(ctx, targetKey, targetInfo)
	if key.PreserveSymlinks {
		return symlinkNode, err
	}
	return node, err
```
wdyt?

go/pkg/cas/upload_test.go
```diff
@@ -75,33 +108,63 @@ func TestFS(t *testing.T) {
...
 					return nil
 				},
 			}
 			client.SmallFileThreshold = 5
 			client.LargeFileThreshold = 10
 
-			if _, err := client.Upload(ctx, inputChanFrom(tc.inputs...)); err != nil {
-				t.Fatalf("failed to upload: %s", err)
+			_, err := client.Upload(ctx, inputChanFrom(tc.inputs...))
+			if tc.expectedErrContains != "" {
```
rubensf, line 151: Use an actual `error` and check for `os.ErrNotExist` instead of string check -- see comment on https://golang.org/pkg/os/#IsNotExist.

This will also simplify this check:

```
if err != tc.expectedErr { t.Fatalf("error mismatch") }
if err != nil { continue } // Happy path where error matches expected non-nil error
```

go/pkg/cas/upload.go
nodirg: this is exactly the code I had before, but I figured that a comment is warranted because otherwise it is super obvious why do read the node and throw it away. I'll commit your change and restore the comment.

go/pkg/cas/upload.go
nodirg: `relTarget` may have `../`. We don't want `a/../b`

go/pkg/cas/upload_test.go
nodirg: I like the idea, but it doesn't work because today we are using a mix of `errors` and `github.com/pkg/errors` packages. They aren't really compatible. Also we are using an old version of the latter which doesn't have the `Is` function. We should probably update, but this is out of scope of this CL. Added a TODO.

go/pkg/cas/upload.go
nodirg: we don't. done

go/pkg/cas/upload_test.go
nodirg: nvm, updated the errors package.

go/pkg/cas/upload.go
rubensf: Sg

rubensf (approved):

go/pkg/cas/upload.go
nodirg: One concern here is that now the Client is not reusable for different parts of the code that need different Upload-specific settings. It is not very nice, not very generic. The prev implementation didn't have this problem, but it had the problem of additional complexity.

I am currently working on file predicate, and this problem seems very similar. I think we need to add a predicate that applies to all `UploadInput`s, but doing it at the client level seems wrong for the same reasons. Therefore I am thinking of adding `UploadOptions`, and change the `Upload` func signature to

```
  func (...) Upload(ctx context, opt UploadOptions, items <- chan *UploadItem) ...
```

The `UploadOptions` would contain `PreserveSymlinks` and `AllowDanglingSymlinks`, therefore allowing reusing the client for different settings, but still not complicated the cache key. I'd probably do it in a separate CL, since this one is reviewed already.

After that, I'd add file predicate to `UploadOptions` too.

How does this sound?

ramymedhat: I'm not sure I follow how `UploadOptions` would not complicate the cache key. You plan to include it in the cache key right? Do you mean its one struct versus adding a list of different flags and knobs to the cache key explicitly?

nodirg: let's continue in https://github.com/bazelbuild/remote-apis-sdks/pull/314

nodirg: @ramymedhat didn't see your comment earlier. I've responded in https://github.com/bazelbuild/remote-apis-sdks/pull/314#issuecomment-827768249

PR #311, opened 2021-04-23
[cas] Add retries

nodirg:
Move ShouldRetry impl from client package to retry package, since it is
generic. Reuse it in the cas package.

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -147,6 +147,6 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
 	return client, nil
 }
 
-func (c *Client) retry(ctx context.Context, f func() error) error {
+func (c *Client) withRetries(ctx context.Context, f func() error) error {
 	return retry.WithPolicy(ctx, retry.TransientOnly, c.RetryPolicy, f)
 }
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -4,11 +4,13 @@ package cas
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/pkg/errors"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 	"google.golang.org/grpc"
 
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 )
 
@@ -59,8 +61,10 @@ type ClientConfig struct {
 	// single FindMissingBlobs RPC.
 	FindMissingBlobsBatchSize int
 
+	// RetryPolicy specifies how to retry requests on transient errors.
+	RetryPolicy retry.BackoffPolicy
+
 	// TODO(nodir): add per-RPC timeouts.
-	// TODO(nodir): add retries.
 }
 
 // DefaultClientConfig returns the default config.
@@ -84,6 +88,8 @@ func DefaultClientConfig() ClientConfig {
 		FileIOSize: 4 * 1024 * 1024, // 4MiB
 
 		FindMissingBlobsBatchSize: 1000,
+
+		RetryPolicy: retry.ExponentialBackoff(225*time.Millisecond, 2*time.Second, retry.Attempts(6)),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


go/pkg/retry/retry.go
```diff
@@ -60,6 +61,27 @@ const UnlimitedAttempts = Attempts(0)
 // Always always retries, regardless of error.
 func Always(error) bool { return true }
 
+// TransientOnly returns true if the error is transient.
+// It implements ShouldRetry type.
+func TransientOnly(err error) bool {
+	// Retry RPC timeouts. Note that we do *not* retry context cancellations (context.Cancelled);
+	// if the user wants to back out of the call we should let them.
+	if err == context.DeadlineExceeded {
```
atetubou, line 69: better to use https://golang.org/pkg/errors/#Is ?

go/pkg/retry/retry.go
```diff
@@ -60,6 +61,27 @@ const UnlimitedAttempts = Attempts(0)
...
+		return true
+	}
+	s, ok := status.FromError(err)
+	if !ok {
+		return false
+	}
+	switch s.Code() {
+	case codes.Canceled, codes.Unknown, codes.DeadlineExceeded, codes.Aborted,
+		codes.Internal, codes.Unavailable, codes.Unauthenticated, codes.ResourceExhausted:
```
atetubou, line 78: Unauthenticated is not transient?

atetubou (approved):

go/pkg/retry/retry.go
nodirg: yeah, this code should probably use this func, but this func is a part of "errors" package, while this repo seems to be using `github.com/pkg/errors` (i.e. it uses `errors.Wrapf` instead of `fmt.Errorf("foobar: %w", err)`).

Maybe this repo should switch from `github.com/pkg/errors` to `errors`, but this seems out of scope, and probably warrants a discussion with the team.

go/pkg/retry/retry.go
nodirg: agree. removed.

go/pkg/retry/retry.go
atetubou: I think it is fine to use the function as `github.com/pkg/errors` supports that like
https://github.com/pkg/errors/blob/614d223910a179a466c1767a985424175c39b465/go113.go#L16

nodirg: sorry, i meant this file is already using a different `errors` package.
But sure, I've added `errors.Is`, with a package alias. Looks awkward though. Hopefully it is temporary.

atetubou: Ah, I see.

go/pkg/retry/retry.go
```diff
@@ -60,6 +62,27 @@ const UnlimitedAttempts = Attempts(0)
 // Always always retries, regardless of error.
 func Always(error) bool { return true }
```
rubensf, line 63: Can you move `ShouldRetry` after `UnlimitedAttempts`? To group it closer with the retry check funcs.

go/pkg/cas/client.go
```diff
@@ -140,3 +146,7 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
 	// TODO(nodir): Check capabilities.
 	return client, nil
 }
+
+func (c *Client) retry(ctx context.Context, f func() error) error {
```
rubensf, line 150: `tryWithRetries`? `doWithRetries`? wdyt?

go/pkg/retry/retry.go
rubensf: FWIW I like `errors.Is`. We should converge the errors package to fmt.Errorf and errors.Is. Realistically, we're probably not going to touch it much :)

rubensf (approved):

go/pkg/cas/client.go
nodirg: how about `withRetries`?

go/pkg/retry/retry.go
nodirg: done

rubensf (approved):

go/pkg/cas/client.go
rubensf: sgtm!

go/pkg/retry/retry.go
nodirg: Turns out `github.com/pkg/errors` also has `Is`: https://pkg.go.dev/github.com/pkg/errors#Is
maybe we need to update our pin.

PR #313, opened 2021-04-27
[cas] Move file IO semaphore/mutex to Client

nodirg:
Move file IO semaphore from uploader struct to Client, such that
it is shared across all uploaders. Add a comment to uploader to prevent
this in the future.


rubensf (commented):
should `muLargeFile` also be moved? IIUC that's a global system resource.

nodirg:
@rubensf good point, done.

rubensf (approved):

PR #314, opened 2021-04-27
[cas] Add UploadOptions

nodirg:
Add UploadOptions which is optional configuration for Upload() func.
Move PreserveSymlinks and AllowDanglingSymlinks there. This way the
same Client can be reused with different options, i.e. the callsite
can reuse caches.

This slightly simplifies tests too.

In the next CL, predicate will be added to UploadOptions.

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -79,16 +79,6 @@ type ClientConfig struct {
 	// RetryPolicy specifies how to retry requests on transient errors.
 	RetryPolicy retry.BackoffPolicy
 
-	// PreserveSymlinks specifies whether to preserve symlinks or convert them
-	// to regular files.
-	PreserveSymlinks bool
-
-	// AllowDanglingSymlinks specifies whether to upload dangling links or halt
-	// the upload with an error.
-	//
-	// This field is ignored if PreserveSymlinks is false, which is the default.
-	AllowDanglingSymlinks bool
-
 	// TODO(nodir): add per-RPC timeouts.
 }
 
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -56,15 +56,30 @@ type DigestStat struct {
 	Bytes   int64 // total sum of of digest sizes
 }
 
+// UploadOptions is optional configuration for Upload function.
+// The default options are the zero value of this struct.
+type UploadOptions struct {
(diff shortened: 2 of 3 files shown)


nodirg:
in https://github.com/bazelbuild/remote-apis-sdks/pull/309 @ramymedhat wrote:

> I'm not sure I follow how UploadOptions would not complicate the cache key. You plan to include it in the cache key right? Do you mean its one struct versus adding a list of different flags and knobs to the cache key explicitly?

the cache is at the uploader level and so are options. Before adding this to the cache key was necessary because these options were at the `UploadInput` level.

nodirg:
further, I would like to reuse `UploadOptions` to propagate internal settings used only in tests. E.g. size of the disk read buffer.

ramymedhat:
> in #309 @ramymedhat wrote:
> 
> > I'm not sure I follow how UploadOptions would not complicate the cache key. You plan to include it in the cache key right? Do you mean its one struct versus adding a list of different flags and knobs to the cache key explicitly?
> 
> the cache is at the uploader level and so are options. Before adding this to the cache key was necessary because these options were at the `UploadInput` level.

I think having the cache at the client level may be beneficial for the execution use case, where there are multiple uploaders potentially in flight. Granted that makes having the predicate and symlink flags required for correctness. In practice, these fields will most likely be consistent for the same client. But in general the cache is an improvement over the current state so tackling this optimization later makes sense.

ramymedhat (approved):

nodirg:
I agree that it is better to reuse cache when we can, but it is dangerous to assume that the file system didn't change since last time we uploaded something. It might be true for reclient, but not in the general case (I treat this repository as a generic library, so assumptions should be kept minimal).

To accommodate the reclient use case, I think we could provide an option for the reclient to say "I know that it won't change. Just reuse the cache" in some way, e.g. by letting the callsite to provide their own cache, like https://github.com/bazelbuild/remote-apis-sdks/blob/144126c43e73aeeabdbeb8bc4ef0290fa7a91ba7/go/pkg/client/tree.go#L122 accepts cache as param.

rubensf (approved):
Adding a cache to the function signature sgtm (presumably in a later cl?)

nodirg:
@rubensf in another CL, yes. Also in practice it will probably be a field in the `ClientConfig`, not a function param.

PR #315, opened 2021-04-27
[cas] Implement ServerCapabilities check

nodirg:
Move common digest function checking code to digest package.

For consistency, rename `ClientConfig.BatchBlobsBatchSize` to
`MaxBatchTotalSizeBytes` which is the name used in proto.

### DIFF
--- a/go/pkg/digest/digest.go
+++ b/go/pkg/digest/digest.go
@@ -166,6 +166,34 @@ func NewFromReaderWithBuffer(r io.Reader, buf []byte) (Digest, error) {
 	}, nil
 }
 
+// CheckCapabilities returns an error if the digest function is not supported
+// by the server.
+func CheckCapabilities(caps *repb.ServerCapabilities) error {
+	fn := GetDigestFunction()
+
+	if caps.ExecutionCapabilities != nil {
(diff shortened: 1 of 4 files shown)


ramymedhat (approved):

go/pkg/digest/digest.go
```diff
@@ -166,6 +166,34 @@ func NewFromReaderWithBuffer(r io.Reader, buf []byte) (Digest, error) {
...
+			return fmt.Errorf("server requires %v, client uses %v", serverFn, fn)
+		}
+	}
+
+	if caps.CacheCapabilities != nil {
+		cc := caps.CacheCapabilities
+		found := false
+		for _, serverFn := range cc.DigestFunction {
+			if serverFn == fn {
+				found = true
```
atetubou, line 185: it is fine to return from here?

nodirg: i think this makes the code a bit less robust. If another programmer doesn't read the code carefully and adds another check in the end of this func, it would be ignored.

atetubou: I see.

PR #316, opened 2021-04-27
[cas] Add filtering/callback

nodirg:
Use cases:
1) Chrome needs to have different regexp-based path filters in different
UploadItems.
2) reclient needs to ensure that all files are under the exec root
3) reclient needs a non-trivial filter (more than one regexp) per Upload()
  call. It doesn't have to be at UploadInput level.

Add UploadInput.Path{Include,Exclude} of type *regexp.Regexp to support
Chrome's use case. Incroporate the regexps into the FS cache key.

Add UploadOptions.Callback that is invoked for each file and may return
ErrSkip if the file should be skipped. This covers both reclient use
cases.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"sync"
 	"sync/atomic"
@@ -23,6 +24,10 @@ import (
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 )
 
+// ErrFilteredSymlinkTarget is returned when a symlink's target was filtered out
+// via UploadInput.PathExclude or ErrSkip, while the symlink itself wasn't.
+var ErrFilteredSymlinkTarget = errors.New("symlink's target was filtered out")
+
 // UploadInput is one of inputs to Client.Upload function.
 //
 // It can be either a reference to a file/dir (see Path) or it can be an
@@ -38,7 +43,14 @@ type UploadInput struct {
 	// Ignored if Path is not empty.
 	Content []byte
 
-	// TODO(nodir): add Predicate.
+	// PathExclude is a file/dir filter. If PathExclude is not nil and the
+	// absolute path of a file/dir match this regexp, then the file/dir is skipped.
+	// If the Path is a directory, then the filter is evaluated against each file
+	// in the subtree.
+	// See ErrSkip comments for more details on semantics regarding excluding symlinks .
+	//
+	// This field has no effect if Path is empty.
+	PathExclude *regexp.Regexp
 }
 
 // TransferStats is upload/download statistics.
@@ -68,7 +80,27 @@ type UploadOptions struct {
 	//
 	// This field is ignored if PreserveSymlinks is false, which is the default.
 	AllowDanglingSymlinks bool
+
+	// Callback is called for each file/dir to be uploaded.
+	// If it returns an error which is ErrSkip according to errors.Is, then the
+	// file/dir is not processed.
+	// If it returns another error, then the upload is halted with that error.
+	//
+	// Callback might be called multiple times for the same file if different
+	// UploadInputs directly/indirectly refer to the same file, but with different
+	// PathExclude.
+	//
+	// Callback is called from different goroutines.
+	Callback func(absPath string, mode os.FileMode) error
 }
+
+// ErrSkip, when returned by UploadOptions.Callback, means the file/dir must be
+// not be uploaded.
+//
+// Note that if UploadOptions.PreserveSymlinks is true and the ErrSkip is
+// returned for a symlink target, but not the symlink itself, then it may
+// result in a dangling symlink.
+var ErrSkip = errors.New("skip file")
 
 // Upload uploads all inputs. It exits when inputC is closed or ctx is canceled.
 func (c *Client) Upload(ctx context.Context, opt UploadOptions, inputC <-chan *UploadInput) (stats *TransferStats, err error) {
@@ -158,6 +190,7 @@ type uploader struct {
 
 	// wgFS is used to wait for all FS walking to finish.
 	wgFS sync.WaitGroup
+
 	// fsCache contains already-processed files.
 	fsCache cache.SingleFlight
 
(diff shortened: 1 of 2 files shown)


rubensf (commented):
fyi differently from gerrit, you have to rebase this `predicate` branch on the `upload-opt` branch and force push it, or git will generally be confused about the technically-equal-diff-but-different-commits :)

nodirg:
thanks. Normally I do that, but clearly something went wrong in this CL. Fixed.

go/pkg/cas/upload.go
```diff
@@ -38,6 +39,20 @@ type UploadInput struct {
...
 	Content []byte
 
+	// PathInclude is a file/dir filter. If PathInclude is not nil and the absolute
+	// path of a file/dir does not match this regexp, then the file/dir is skipped.
+	// If the Path is a directory, then the filter is evaluated against each file
+	// in the subtree.
+	// See ErrSkip comments for more details on semantics.
+	//
+	// This field has no effect if Path is empty.
+	PathInclude *regexp.Regexp
```
atetubou, line 49: I think we don't use this?

nodirg: we don't, but having PathExclude without PathInclude seems asymmetric. 
Since this is a general-purpose lib, I figured why not implement both right away.

atetubou: I see, but my preference is to follow https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it
Resolving asymmetric feeling is not important compared to not having unnecessary code and tests for that?

nodirg: @ramymedhat to decide

ramymedhat: The command proto also only defines input [exclusions](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/api/command/command.proto#L70), so I think just having exclusions is sufficient for now.

nodirg: Done.

go/pkg/cas/upload.go
```diff
@@ -38,7 +39,14 @@ type UploadInput struct {
 	// Ignored if Path is not empty.
 	Content []byte
 
-	// TODO(nodir): add Predicate.
+	// PathExclude is a file/dir filter. If PathExclude is not nil and the
+	// absolute path of a file/dir match this regexp, then the file/dir is skipped.
+	// If the Path is a directory, then the filter is evaluated against each file
+	// in the subtree.
+	// See ErrSkip comments for more details on semantics.
```
rubensf, line 46: I'm not sure how ErrSkip is being used along PathExclude? Reading the code it seems that ErrSkip is exclusively being used with callback, which is (afaict) a completely separate thing from PathExclude.

go/pkg/cas/upload.go
```diff
@@ -68,8 +76,24 @@ type UploadOptions struct {
...
 	// This field is ignored if PreserveSymlinks is false, which is the default.
 	AllowDanglingSymlinks bool
+
+	// Callback is called for each file/dir to be uploaded.
+	// If it returns an error which is ErrSkip according to errors.Is, then the
+	// file/dir is not not uploaded.
+	// If it returns another error, then the upload is halted with that error.
+	//
+	// Callback is called from different goroutines.
+	Callback func(absPath string, mode os.FileMode) error
```
rubensf, line 86: We should document that the Callback is not reflected at on the cache.

Particularly, to allow for skipping processing files, Callback can be called repeatedly for the same file, and the files processed by callback might still not be included/uploaded since this happens before processing the file.

go/pkg/cas/upload.go
rubensf: Actually, we could use `PathInclude` to filter files under the input root for reclient, instead of the callback - and that gets saved on the cache key.

Wdyt?

go/pkg/cas/upload.go
nodirg: ErrSkip itself isn't used in PathExclude, but the semantics documented in ErrSkip is common to both. I figured I could "reuse" docs, but I can copy them too if you prefer.

Specifically I am referring to semantics around symlinks.

go/pkg/cas/upload.go
```diff
@@ -195,21 +220,47 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
...
+		case err != nil:
+			return nil, err
+		}
+	}
+	if pathExclude != nil && pathExclude.MatchString(absPath) {
+		return nil, nil
+	}
+
+	// Make a cache key.
+	cacheKey := struct {
```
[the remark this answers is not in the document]
nodirg: made its own type, but still in the same func. It communicates to the reader that it is not used in any other func, which I find nice.

go/pkg/cas/upload.go
```diff
@@ -195,21 +220,47 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
...
 	return nil
 }
 
 // visitFile visits the file/dir depending on its type (regular, dir, symlink).
 // Visits each file only once.
-func (u *uploader) visitFile(ctx context.Context, absPath string, info os.FileInfo) (dirEntry proto.Message, err error) {
-	node, err := u.fsCache.LoadOrStore(absPath, func() (interface{}, error) {
+//
+// If the file should be skipped, then returns (nil, nil).
+func (u *uploader) visitFile(ctx context.Context, absPath string, info os.FileInfo, pathExclude *regexp.Regexp) (dirEntry proto.Message, err error) {
```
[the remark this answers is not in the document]
nodirg: done

go/pkg/cas/upload.go
nodirg: that's a great observation. Thank you!

I've changed the order of Callback/regexp checks, so now Callback func doesn't have to bother with PathExclude.

I've documented that Callback might be called twice for the same file.

go/pkg/cas/upload.go
nodirg: - reclient has one execRoot for the entire invocation, so not sure it would help reclient
- it would have a different semantics from now. Currently it returns an error https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/client/tree.go#L96, hence the Callbacks func signature that let's you return an error

go/pkg/cas/upload.go
```diff
@@ -195,21 +224,49 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
...
+	if pathExclude != nil && pathExclude.MatchString(absPath) {
+		return nil, nil
+	}
+	// Call the callback only after checking the pathExclude.
+	if u.Callback != nil {
+		switch err := u.Callback(absPath, info.Mode()); {
+		case errors.Is(err, ErrSkip):
+			return nil, nil
+		case err != nil:
+			return nil, err
```
rubensf, line 248: I think you can remove the TODO on L432 original file?

go/pkg/cas/upload_test.go
```diff
@@ -94,13 +96,81 @@ func TestFS(t *testing.T) {
 		inputs              []*UploadInput
 		wantScheduledChecks []*uploadItem
 		wantErr             error
+		wantErrContain      string
```
rubensf, line 99: instead of a `wantErrContain`, we can make a `var ErrDanglingSymlink = errors.New("symlink's target was filtered out")` and return `errors.Wrap(ErrDanglingSymlink, "path: %v, target: %v")`.

I generally think sentinels are nicer than relying on strings, as to not allow eg the same string to match multiple errors.

go/pkg/cas/upload.go
rubensf: I'm hoping to avoid callbacks since they don't get cached, but I agree that error-ing at a source outside the root is essential for reclient. But the best option I can think of is naming a `InputRoot` (instead of `PathInclude`, as to not confuse users for the different semantics) that is just a prefix + error-ing.

This is by far less powerful than a callback, but I _think_ from an RE perspective that's good enough, since the concept of an input/exec root is quite central to the API. Wdyt?

I don't think the performance impact would be substantial (besides eg potentially caching errors, and avoiding one callstack that the go compiler might as well inline) to block on this though.

rubensf (approved):
just nits

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload_test.go
nodirg: I had to name it `ErrFilteredSymlinkTarget` though, because `ErrDanglingSymlink` looks like result of `!AllowDanglingSymlink`, in which case we actually don't return it.

rubensf: lgtm

rubensf (approved):

PR #318, opened 2021-04-28
[cas] Add RPCConfig

nodirg:
First of all, this CL fixes a bug that max-bytes was confused with
max-count, in (*Client).Upload().

Stepping back, the reason is that the RPC config fields are long and
confusing. Earlier version of the code attempted to deal with this by
partially following a rule that a per-RPC config fields starts with the RPC
name, but clearly this wasn't enough.

Introduce RPCConfig struct which contains RPC-specific configuration,
namely concurrency, max-size, max-count and timeout. Use the struct
as a type of two new fields: ClientConfig.FindMissingBlobs and
ClientConfig.BatchUpdateBlobs, which match RPC names exactly.

This is different from client package's approach, which is having a
map[string]RPCConfig. It was considered, but it causes non-type-safe
code, e.g. using string literals to refer to configuration of a
particular RPC, in multiple places. This could be mitigated by adding constant strings, but
this is no better than having strongly-typed fields instead of the map.

Also it seems like the main reason to use map in the client package is
to have defaults that apply to all RPCs, e.g. it is reasonable to have
1s as the default timeout. In contrast, the cas package defines more
than timeout, namely max-size and max-count, and those values
are rather different from RPC to RPC. So the wildcard approach used
in the client package provides less value to the cas package.

Along all of this, this CL implements per-RPC timeouts, but it is a
rather small part of this CL.

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -123,7 +123,7 @@ func DefaultClientConfig() ClientConfig {
 		FindMissingBlobs: RPCConfig{
 			Concurrency: 64,
 			MaxItems:    1000,
-			Timeout:     time.Second,
+			Timeout:     time.Minute,
 		},
 		BatchUpdateBlobs: RPCConfig{
 			Concurrency: 256,
@@ -134,7 +134,7 @@ func DefaultClientConfig() ClientConfig {
 			MaxItems: 4000,
 			// 4MiB is the default gRPC request size limit.
 			MaxSizeBytes: 4 * 1024 * 1024,
-			Timeout:      time.Second,
+			Timeout:      time.Minute,
 		},
 
 		RetryPolicy: retry.ExponentialBackoff(225*time.Millisecond, 2*time.Second, retry.Attempts(6)),
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -31,9 +31,8 @@ type Client struct {
 	// ClientConfig is the configuration that the client was created with.
 	ClientConfig
 
-	byteStream          bspb.ByteStreamClient
-	cas                 repb.ContentAddressableStorageClient
-	semBatchUpdateBlobs *semaphore.Weighted
+	byteStream bspb.ByteStreamClient
+	cas        repb.ContentAddressableStorageClient
 
 	// TODO(nodir): ensure it does not hurt streaming.
 	semFileIO *semaphore.Weighted
@@ -98,6 +97,12 @@ type RPCConfig struct {
 
 	// Timeout is the maximum duration of the RPC.
 	Timeout time.Duration
+
+	sem *semaphore.Weighted
+}
+
+func (c *RPCConfig) init() {
+	c.sem = semaphore.NewWeighted(int64(c.Concurrency))
 }
 
 // DefaultClientConfig returns the default config.
@@ -226,11 +231,13 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
 // creating a real gRPC connection. This function exists purely to aid testing,
 // and is tightly coupled with NewClientWithConfig.
 func (c *Client) init() {
-	c.semBatchUpdateBlobs = semaphore.NewWeighted(int64(c.BatchUpdateBlobs.Concurrency))
 	c.semFileIO = semaphore.NewWeighted(int64(c.FSConcurrency))
 	c.fileIOBufs.New = func() interface{} {
 		return make([]byte, c.FileIOSize)
 	}
+
+	c.FindMissingBlobs.init()
+	c.BatchUpdateBlobs.init()
 }
 
 // unaryRPC calls f with retries, and with per-RPC timeouts.
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -86,7 +86,7 @@ type ClientConfig struct {
 // RPCConfig is configuration for a particular CAS RPC.
 // Some of the fields might not apply to certain RPCs.
 type RPCConfig struct {
-	// Concurrency is the maximum number of RPCs at a time.
+	// Concurrency is the maximum number of RPCs in flight.
 	Concurrency int
 
 	// MaxSizeBytes is the maximum size of the request/response, in bytes.
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -31,8 +31,9 @@ type Client struct {
 	// ClientConfig is the configuration that the client was created with.
 	ClientConfig
 
-	byteStream bspb.ByteStreamClient
-	cas        repb.ContentAddressableStorageClient
+	byteStream          bspb.ByteStreamClient
+	cas                 repb.ContentAddressableStorageClient
+	semBatchUpdateBlobs *semaphore.Weighted
 
 	// TODO(nodir): ensure it does not hurt streaming.
 	semFileIO *semaphore.Weighted
@@ -97,12 +98,6 @@ type RPCConfig struct {
 
 	// Timeout is the maximum duration of the RPC.
 	Timeout time.Duration
-
-	sem *semaphore.Weighted
-}
-
-func (c *RPCConfig) init() {
-	c.sem = semaphore.NewWeighted(int64(c.Concurrency))
 }
 
 // DefaultClientConfig returns the default config.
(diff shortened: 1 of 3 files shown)


go/pkg/cas/client.go
```diff
@@ -115,14 +120,22 @@ func DefaultClientConfig() ClientConfig {
...
-		BatchUpdateBlobsConcurrency: 256, // arbitrary
-
-		// This is a suggested approximate limit based on current RBE implementation for writes.
-		// Above that BatchUpdateBlobs calls start to exceed a typical minute timeout.
-		// This default might not be best for reads though.
-		MaxBatchTotalSizeBytes: 4000,
+		FindMissingBlobs: RPCConfig{
+			Concurrency: 64,
+			MaxItems:    1000,
+			Timeout:     time.Second,
```
atetubou, line 126: This may be too strict if we need to send RPC to opposite side of the earth?

go/pkg/cas/client.go
```diff
@@ -115,14 +120,22 @@ func DefaultClientConfig() ClientConfig {
...
+		BatchUpdateBlobs: RPCConfig{
+			Concurrency: 256,
+
+			// This is a suggested approximate limit based on current RBE implementation for writes.
+			// Above that BatchUpdateBlobs calls start to exceed a typical minute timeout.
+			// This default might not be best for reads though.
+			MaxItems: 4000,
+			// 4MiB is the default gRPC request size limit.
+			MaxSizeBytes: 4 * 1024 * 1024,
+			Timeout:      time.Second,
```
atetubou, line 137: ditto

go/pkg/cas/client.go
nodirg: Changed to minute. The `client` package uses a minute, but for some reason I thought it is a second.

go/pkg/cas/client.go
```diff
@@ -192,15 +231,24 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
...
+	c.FindMissingBlobs.init()
+	c.BatchUpdateBlobs.init()
 }
 
-func (c *Client) withRetries(ctx context.Context, f func() error) error {
-	return retry.WithPolicy(ctx, retry.TransientOnly, c.RetryPolicy, f)
+// unaryRPC calls f with retries, and with per-RPC timeouts.
+// Does not limit concurrency.
+// It is useful when f calls an unary RPC.
+func (c *Client) unaryRPC(ctx context.Context, cfg *RPCConfig, f func(context.Context) error) error {
```
ramymedhat, line 246: Will this eventually be moved to a separate package so `client` can also use it?

go/pkg/cas/client.go
```diff
@@ -68,31 +67,42 @@ type ClientConfig struct {
...
+
+	// MaxItems is the maximum number of blobs/digests per RPC.
+	// Applies to batch RPCs.
+	MaxItems int
 
-	// TODO(nodir): add per-RPC timeouts.
+	// Timeout is the maximum duration of the RPC.
+	Timeout time.Duration
+
+	sem *semaphore.Weighted
```
ramymedhat, line 101: It's slightly awkward that there's a semaphore in a config struct? Would moving it to the client make sense?

go/pkg/cas/client.go
nodirg: Maybe? I thought of this too, but it is a bit hard for me to tell right now because this is specific to `RPCConfig` which is a bit specific CAS (it has `MaxItems` for example).

In Go's author own words, "little copying is better than little dependency" https://youtu.be/PAAkCSZUG1c?t=567
We are trying to save 5 lines of code here, at least for now.

go/pkg/cas/client.go
nodirg: This was intentional, but undone now.

ramymedhat: Sorry not following, what is undone? My point is that it seems like an antipattern that the config struct would need to be initialized and would have state.

nodirg: sorry, just noticed that `git push` was still waiting for me to enter my ssh key password.. Still getting used to GitHub.
PTAL again.

I said "undone" because I moved semaphores to RPCConfig initially, and the undid that after your request.

ramymedhat: Ah now it makes sense! Thanks.

ramymedhat (approved):

go/pkg/cas/client.go
```diff
@@ -68,31 +68,36 @@ type ClientConfig struct {
...
+type RPCConfig struct {
+	// Concurrency is the maximum number of RPCs at a time.
+	Concurrency int
+
+	// MaxSizeBytes is the maximum size of the request/response, in bytes.
+	MaxSizeBytes int
 
-	// TODO(nodir): add per-RPC timeouts.
+	// MaxItems is the maximum number of blobs/digests per RPC.
+	// Applies to batch RPCs.
```
rubensf, line 96: `Applies only to batch RPCs` or `Ignored for Stream RPCs`.

go/pkg/cas/client.go
```diff
@@ -68,31 +68,36 @@ type ClientConfig struct {
...
 
 	// IgnoreCapabilities specifies whether to ignore server-provided capabilities.
 	// Capabilities are consulted by default.
 	IgnoreCapabilities bool
+}
+
+// RPCConfig is configuration for a particular CAS RPC.
+// Some of the fields might not apply to certain RPCs.
+type RPCConfig struct {
+	// Concurrency is the maximum number of RPCs at a time.
```
rubensf, line 89:
```suggestion
	// Concurrency is the maximum number of RPCs in flight.
```
?

go/pkg/cas/client.go
```diff
@@ -145,9 +158,30 @@ func (c *ClientConfig) Validate() error {
 		return fmt.Errorf("FileIOSize must be positive")
 
 	// Do not allow more than 100K, otherwise we might run into the request size limits.
-	case c.FindMissingBlobsBatchSize <= 0 || c.FindMissingBlobsBatchSize > 10000:
-		return fmt.Errorf("FindMissingBlobsBatchSize must be in [1, 10000]")
+	case c.FindMissingBlobs.MaxItems > 10000:
```
rubensf, line 161: Comment says 100k but code says 10k?

go/pkg/cas/client.go
```diff
@@ -145,9 +158,30 @@ func (c *ClientConfig) Validate() error {
...
+		return errors.Wrap(err, "FindMissingBlobs")
+	}
+	if err := c.BatchUpdateBlobs.validate(true); err != nil {
+		return errors.Wrap(err, "BatchUpdateBlobs")
+	}
+	return nil
+}
+
+// validate returns an error if the config is invalid.
+func (c *RPCConfig) validate(validateMaxSizeBytes bool) error {
```
rubensf, line 175: I'm not sure if validating MaxSizeBytes makes sense. If we're validating that, we might as well validate MaxItems for streams... And adding one option per RPC sounds complex.

go/pkg/cas/client.go
```diff
@@ -68,31 +68,36 @@ type ClientConfig struct {
...
+}
+
+// RPCConfig is configuration for a particular CAS RPC.
+// Some of the fields might not apply to certain RPCs.
+type RPCConfig struct {
+	// Concurrency is the maximum number of RPCs at a time.
+	Concurrency int
+
+	// MaxSizeBytes is the maximum size of the request/response, in bytes.
+	MaxSizeBytes int
```
rubensf, line 93: Add: `Applies only to batch RPCs.`

rubensf (approved):
bunch of nits.

Also - cl description says `Now that we have per-RPC chunk of memory, move per-RPC semaphores there, and limit concurrency in the uploader.check(), as it should.`, but I think you reverted from that?

go/pkg/cas/client.go
nodirg: This is not about stream vs non-stream. 
This is about batch vs non-batch. An unary RPC can be non-batch too.

How can I say it more clearly, while saying concise?

go/pkg/cas/client.go
nodirg: clarified the comment.

go/pkg/cas/client.go
nodirg: removed.

go/pkg/cas/client.go
nodirg: See another comment. Clarified though. PTAL.

nodirg:
@rubensf still learning GitHub: is the first comment treated as CL description? I wasn't sure if it is, or it is something that the CL author at the time.

go/pkg/cas/client.go
nodirg: took a stab. PTAL.

go/pkg/cas/client.go
rubensf: LGTM

rubensf (approved):

rubensf:
Technically, when you try "squash and merging" a PR, all the commits descriptions are put on the message and you have to manually edit them to keep them sane (ie usually just pick the first commit's message).

Github defaults your initial PR comment to the first commit's message, but IIRC they're technically separate things that can be edited separately.

PR #319, opened 2021-04-29
[cas] Limit FindMissingBlobs concurrency

nodirg:
Respect ClientConfig.FindMissingBlobs.Concurrency by adding a semaphore.


rubensf (approved):
Should FindMissingBlob's concurrency be higher? I'd expect it to be cheaper than uploading, but it's set to a lower concurrency.

nodirg:
sorry, didn't see your message earlier. How does 256 sound?

rubensf:
256 looks good to me

PR #320, opened 2021-04-29
[cas] Implement streaming

nodirg:
This CL implements streaming of files/blobs using ByteStream service
with compression. This is the last major part of the uploader
implementation.

Update `RPCConfig.Maxitems` docs: do not specify "unary". It is possible to have a streaming RPC where
each request deals with multiple items.

This CL implements retries of the stream, but not per-RPC timeout. In particular, it is unclear how to implement
per-request-in-stream timeout, which is what we need.

This CL does not use https://github.com/mostynb/zstdpool-syncpool - no need. Its main value seems to be finalizer-based
guarantee that Close() method will be called.
This code guarantees to close the encoder without a need for a finalizer.


go/pkg/cas/upload.go
```diff
@@ -21,8 +23,16 @@ import (
...
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
 )
 
+var zstdEncoders = sync.Pool{
+	New: func() interface{} {
+		enc, _ := zstd.NewWriter(nil)
+		return enc
+	},
```
mostynb, line 33: If you're not going to use zstdpool here I would recommend leaving a comment about not forgetting to call Close on encoders, it's a subtle gotcha waiting to happen.

go/pkg/cas/upload.go
```diff
@@ -589,6 +601,143 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+			switch _, err := io.CopyBuffer(enc, r, fileBuf); {
+			case err == io.ErrClosedPipe:
+				// The other goroutine exited before we finished encoding.
+				// Might be a cache hit or context cancelation.
+				// In any case, the other goroutine has the actual error, so return nil
+				// here.
+				return nil
+			case err != nil:
+				return errors.Wrapf(err, "failed to read the file/blob")
+			}
```
mostynb, line 666: It looks like `enc.Close()` will not be called if we return in either of these two cases.

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
```diff
@@ -589,6 +604,144 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+		return err
+	}
+	defer r.Close()
+
+	// TODO(nodir): implement per-RPC timeouts. No nice way to do it.
+	rewind := false
+	return u.withRetries(ctx, func(ctx context.Context) error {
+		// Do not rewind if this is the first attempt.
+		if rewind {
+			if _, err := r.Seek(0, io.SeekStart); err != nil {
```
rubensf, line 640: I think these retries have to start at the `WriteResponse.CommitedSize` -- or you'll be writing from zero on top of partially written data on the server. See https://github.com/googleapis/googleapis/blob/master/google/bytestream/bytestream.proto#L140

go/pkg/cas/upload.go
```diff
@@ -589,6 +604,144 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+			case err != nil:
+				return err
+			case res.CommittedSize != item.Digest.SizeBytes:
+				return fmt.Errorf("unexpected commitSize: got %d, want %d", res.CommittedSize, item.Digest.SizeBytes)
+			}
+
+			// Update stats.
+			cacheHit := !req.FinishWrite
+			if !cacheHit {
+				atomic.AddInt64(&u.stats.Streamed.Bytes, item.Digest.SizeBytes)
```
rubensf, line 728: This way we have no way of measuring the impact of compression on the network. 

Particularly, we should also be able to count how often we're retrying (it'd also be nice to be able to infer how many bytes we're retrying, but I think realistically it'd either be too few to matter, or so much that we'd have to fix other network/backend load issues).

go/pkg/cas/upload.go
nodirg: While I agree that consistency would be nice, my main concern, which I didn't mention anywhere, is the use of finalizers in zstd-pool. They feel like a bit overkill. Dave Cheney recommends against them and provides some history around finalizers here: https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_finalisation
His main point is in the end: 

> Do not write programs who’s correctness depends on finalisation, instead associate the lifetime of a resource to the lifetime of a goroutine.

which is what this code is doing (except, this is not about correctness).

Thoughts?

go/pkg/cas/upload.go
```diff
@@ -21,8 +23,19 @@ import (
...
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
 )
 
+// zstdEncoders is a pool of ZStd encoders.
+// Clients of this pool must call Close() on the encoder after using the
+// encoder.
+var zstdEncoders = sync.Pool{
+	New: func() interface{} {
+		enc, _ := zstd.NewWriter(nil)
```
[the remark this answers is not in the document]
nodirg: could you elaborate what do we gain from this? 

In my experiments, most of the time is spent on uploading large files which are uploaded sequentially (most small files are already present). If we use concurrency=1, then we would be using only one CPU per `cas.Client` most of the time.

go/pkg/cas/upload.go
mostynb: I agree in principle that using finalizers isn't great, but fixing that properly would require significant changes to the zstd library because it breaks this other one of Dave Cheney's tips:
https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_leave_concurrency_to_the_caller

go/pkg/cas/upload.go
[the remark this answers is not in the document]
mostynb: `zstd.NewWriter()` spawns goroutines internally, IIRC it defauts to the number of virtual CPU cores so `Encoder.EncodeAll()` can process that many simultaneous encodes. But if you're using a `sync.Pool` you're probably only ever doing one encode per `Encoder`. If you create these with `WithEncoderConcurrency(1)` then it creates fewer goroutines.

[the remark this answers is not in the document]
rubensf: @nodirg I'm not sure of the specifics of your use case, but unless you have a large tail end of files < ncores, you'll have enough parallelism controlled by `semByteStreamWrite`.

Few tail ends would be problematic with one concurrency though.

go/pkg/cas/upload.go
nodirg: I agree that it would be hard to fix it at the zstd package level.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: see `muLargeFiles`: we upload at most one large file at a time (because it results in higher readers from `pd-standard` disks), and this is where most of the upload time goes (in my case, 22min to upload ~50GB, where vasy majority of bytes reside in few large files). So we have 8 cores (e2-standard-8 machine) with one upload at a time.

to be super clear: `muLargeFiles` limits concurrency much more than `semByteStramWrite` - we can ignore the latter in this case.

go/pkg/cas/upload.go
```diff
@@ -589,6 +604,144 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+		// Compress using an in-memory pipe. This is mostly to accomodate the fact
+		// that zstd package expects a writer.
+		// Note that using io.Pipe() means we buffer only bytes that were not uploaded yet.
+		pr, pw := io.Pipe()
+		enc.Reset(pw)
+
+		// Read from disk and make RPCs concurrently.
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error {
+			switch _, err := io.CopyBuffer(enc, r, fileBuf); {
```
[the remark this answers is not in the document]
nodirg: Great catch! Thanks. I sent out https://github.com/bazelbuild/remote-apis-sdks/pull/322 and will rebase this PR later.

go/pkg/cas/upload.go
```diff
@@ -589,6 +604,144 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+
+		// Compress using an in-memory pipe. This is mostly to accomodate the fact
+		// that zstd package expects a writer.
+		// Note that using io.Pipe() means we buffer only bytes that were not uploaded yet.
+		pr, pw := io.Pipe()
+		enc.Reset(pw)
+
+		// Read from disk and make RPCs concurrently.
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error {
```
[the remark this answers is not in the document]
nodirg: your suggestion becomes mandatory with conditional compression. Will do.

go/pkg/cas/upload.go
nodirg: I agree. In my original version I was actually adding `WriteOffset` and unconditionally, but then realized that this is inconsistent with stats.Batched. Also it is weird to see streamed.Bytes higher than cacheMisses - it may happen if all streams resulted in cache hits.

WDYT about a new stats field dedicated specifically to the network? It would probably combine batched and streamed writes?

go/pkg/cas/upload.go
```diff
@@ -589,6 +604,144 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+			}
+
+			if err := enc.Close(); err != nil {
+				return errors.Wrapf(err, "failed to close the zstd encoder")
+			}
+			return pw.Close()
+		})
+		eg.Go(func() error {
+			defer pr.Close()
+			req := &bspb.WriteRequest{
```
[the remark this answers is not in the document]
nodirg: Made it optoinal/configurable and a TODO for capabilities (I was relying on my code-completion to tell me if we have a capability about that, but turns out we have a stale version of the remove-apis proto).

go/pkg/cas/upload.go
nodirg: Added a TODO.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: I've extracted streaming part, which is the bulk of the code. `eg.Go` stayed somewhat the same, i.e. I didn't create sub-functions for them. partially because those functions aren't large, partially because moving them our would require carrying lots of state in params causing actually more code. Most of it is comments anyway.

BTW, I think GitHub is doing a poor job here. In my editor tab is configured to two spaces, while on GitHub one tab looks like 6 spaces? So indentation appears much larger than it actually is. In Gerrit it looks like 4 spaces. So if we were using Gerrit, I would look better :)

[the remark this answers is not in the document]
nodirg: Looks much better with ?ts=2

https://github.com/bazelbuild/remote-apis-sdks/pull/320/files/a5351ae12a5c3384b5d0296658af924210cb4811#diff-bb4a3d7f90b86fd97be5e865f3f0a58dfc1ed61e1f77ac269d3b36fad7bdd238

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: using `ReadFrom` onw.

go/pkg/cas/client.go
```diff
@@ -238,24 +267,39 @@ var emptyReader = bytes.NewReader(nil)
...
 	c.semFindMissingBlobs = semaphore.NewWeighted(int64(c.FindMissingBlobs.Concurrency))
 	c.semBatchUpdateBlobs = semaphore.NewWeighted(int64(c.BatchUpdateBlobs.Concurrency))
+	c.semByteStreamWrite = semaphore.NewWeighted(int64(c.ByteStreamWrite.Concurrency))
 
 	c.semFileIO = semaphore.NewWeighted(int64(c.FSConcurrency))
 	c.fileBufReaders.New = func() interface{} {
 		return bufio.NewReaderSize(emptyReader, int(c.FileIOSize))
 	}
+
+	streamBufSize := 32 * 1024 // by default, send 4KiB chunks.
```
atetubou, line 277:
```suggestion
	streamBufSize := 32 * 1024 // by default, send 32KiB chunks.
```

go/pkg/cas/upload.go
```diff
@@ -662,6 +677,160 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+	if err := u.semByteStreamWrite.Acquire(ctx, 1); err != nil {
+		return err
+	}
+	defer u.semByteStreamWrite.Release(1)
+
+	// Prepare compressor, if needed.
+	enableCompression := item.Digest.SizeBytes >= u.CompressedBytestreamThreshold
+	var enc *zstd.Encoder
+	if enableCompression {
+		enc = zstdEncoders.Get().(*zstd.Encoder)
```
rubensf, line 694: I think we can move this within `withRetries`, after we check if we're not enabling compression.

Pretty much any error leading to retries will requiring flushing the encoder and `Resetting` it. We'd lose the time to put/retrieve an encoder from the pool, but that shouldn't happen often enough to be significant.

go/pkg/cas/upload.go
```diff
@@ -662,6 +677,160 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+		defer func() {
+			enc.Close()
+			zstdEncoders.Put(enc)
+		}()
+	}
+
+	streamBuf := u.streamBufs.Get().([]byte)
```
rubensf, line 701: Move this to `streamFromReader`? I don't expect retries to happen often enough to actually impact performance.

go/pkg/cas/upload.go
```diff
@@ -662,6 +677,160 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+	case err != nil:
+		return err
+	case res.CommittedSize != digest.SizeBytes:
+		return fmt.Errorf("unexpected commitSize: got %d, want %d", res.CommittedSize, digest.SizeBytes)
+	}
+
+	// Update stats.
+	cacheHit := !req.FinishWrite
+	if !cacheHit {
+		atomic.AddInt64(&u.stats.Streamed.Bytes, digest.SizeBytes)
```
rubensf, line 821: _github won't let me reply on the original thread_

I think adding a "RealMoved" or "NetworkBytes" to `DigestStats`, so that we have more granularity between batch and stream, is better. It gives more granularity on the impact of compression, particularly when we get BatchCompression on the API (it'll happen at some point, just not in the near future). For now, `NetworkBytes` will be the same to `Bytes` for batches, not accounting for retries.

go/pkg/cas/upload_test.go
```diff
@@ -354,6 +356,69 @@ func TestSmallBlobs(t *testing.T) {
 	}
 }
 
+func TestStreaming(t *testing.T) {
```
rubensf, line 359: FYI if you leave the retries here, we should have tests with retries.

go/pkg/cas/upload.go
```diff
@@ -662,6 +677,160 @@ func (u *uploader) uploadBatch(ctx context.Context, reqs []*repb.BatchUpdateBlob
...
+	r, err := item.Open()
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+
+	// TODO(nodir): implement per-RPC timeouts. No nice way to do it.
+	rewind := false
+	return u.withRetries(ctx, func(ctx context.Context) error {
+		// TODO(nodir): check QueryWriteStatus.
```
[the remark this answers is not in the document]
rubensf: FWIW, I just sanity checked here - RBE doesn't currently support resumable uploads (ie we never "commit" non-full blobs). In case of eg local timeouts, multiple in flight uploads are valid but only one gets committed. In the current implementation we always reset to zero, and that happens to work.

I'm ok if you want to leave the TODO, just mention this won't work for backends that implement resumable uplodas.

go/pkg/cas/client.go
```diff
@@ -271,5 +311,8 @@ func (c *Client) checkCapabilities(ctx context.Context) error {
 	if c.BatchUpdateBlobs.MaxSizeBytes > int(caps.CacheCapabilities.MaxBatchTotalSizeBytes) {
 		c.BatchUpdateBlobs.MaxSizeBytes = int(caps.CacheCapabilities.MaxBatchTotalSizeBytes)
 	}
+
+	// TODO(nodir): check compression capabilities.
```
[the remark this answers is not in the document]
nodirg: looks like `.pb.go` is stale: https://github.com/bazelbuild/remote-apis/blob/0943dc4e70e1414735a85a3167557392c177ff45/build/bazel/remote/execution/v2/remote_execution.pb.go#L3458

go/pkg/cas/client.go
```diff
@@ -147,6 +164,15 @@ func DefaultClientConfig() ClientConfig {
 			MaxSizeBytes: 4 * 1024 * 1024,
 			Timeout:      time.Minute,
 		},
+		ByteStreamWrite: RPCConfig{
+			Concurrency: 256,
+			// 4MiB is the default gRPC request size limit.
+			MaxSizeBytes: 4 * 1024 * 1024,
+			Timeout:      time.Minute,
```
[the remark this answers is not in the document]
nodirg: note: the comment above says that these fields apply to each request in a stream (i.e. not entire stream). From that perspective, 1m per chunk is more than enough (The point of timeout is to prevent hanging, but if we are making progress - so I think it should be per request, not stream).

go/pkg/cas/upload.go
nodirg: sgtm, let's punt

go/pkg/cas/upload.go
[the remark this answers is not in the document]
nodirg: > You don't need to check QueryWriteStatus - that'd be an unnecessary RPC call. While not explicitly stated, I can't foresee a sane backend implementation query QueryWriteStatus.CommittedSize is diff from WriteResponse.CommittedSize. You can check the response on stream.CloseAndRecv();.

Sorry, I am not sure I understood this. Do you mean that instead of making `QueryWriteStatus` RPC, we can opportunistically start streaming assuming `CommittedSize=0`, but then check `WriteResponse.CommittedSize`. We don't get `WriteResponse.CommittedSize` until we start streaming, and we need to decide on the offset before we start streaming.

> I also don't think you can leave the withRetries with SeekStart(0) here. That'd make the retries useless (ie they'd never succeed. Either remove the retrying for now or fix the seek.

Could you elaborate on why this wouldn't work? Suppose we started streaming, something went wrong and we closed the stream in defer. I assume the server will throw the results and reset to its original state. On the second attempt we start on a clean slate.

I've changed TODO to "add support for resumable uploads", since it is conceivable to implement it on the server (at least without compression)

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
nodirg: added a todo

go/pkg/cas/upload_test.go
nodirg: `fakes.CAS` currently doesn't support transient errors. Adding it there seems like a large feature (ideally done in all RPCs), so I've added a TODO in this CL

go/pkg/cas/client.go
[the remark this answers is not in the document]
rubensf: Ah, I missed your TODO comment regarding per-RPC timeouts - which means so far this Timeout is being ignored? sgtm

go/pkg/cas/client.go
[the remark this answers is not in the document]
rubensf: FYI I just updated the library. Sg to do this in a diff PR.

go/pkg/cas/upload.go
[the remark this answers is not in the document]
rubensf: FWIW I think extracting the streaming part is already good enough - and I wish Github would let me use 2/4 by default, but I couldn't figure a way that doesn't involve me tweaking the URL everything, *sigh*. (thanks for the tip!)

rubensf (approved):
> Sorry, I am not sure I understood this. Do you mean that instead of making QueryWriteStatus RPC, we can opportunistically start streaming assuming CommittedSize=0, but then check WriteResponse.CommittedSize. We don't get WriteResponse.CommittedSize until we start streaming, and we need to decide on the offset before we start streaming.

Ah, you're right that for the first call we'd have to check `QueryWriteStatus`, my bad. On subsequent ones though, we could opportunistically check `WriteResponse.CommittedSize`

> I assume the server will throw the results and reset to its original state.

IIUC there's no expectation by ByteStream that the server would throw away or save partial work. RBE's always throws away the partial work, hence Seeking to 0 without checking the CommittedSize works. Leaving the TODO with "Support resumable uploads" sounds good to me as it's a more accurate description of the motivation to tweak the Seeking :)

go/pkg/cas/upload.go
nodirg: a TODO was added.

PR #322, opened 2021-04-29
[cas] Move file IO buffering deeper

nodirg:
Passing explicit []byte buffers to various functions does not guarantee
buffered reads from an io.Reader. In particular. io.CopyBuffer may
ignore the passed buffer. As a result, the file IO size might get
ignored.

Move buffering deeper by using bufio.Reader. Use it throughout.
Replace a sync.Pool of []byte with a pool of bufio.Readers.

Ungeneralize readSeekCloser since we no longer read from *os.File
directly. Still, preserve seeking (only relative to start) because streaming files must start
with the current-offset-on-the-server.

### DIFF
--- a/go/pkg/cas/client.go
+++ b/go/pkg/cas/client.go
@@ -2,6 +2,8 @@
 package cas
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"sync"
@@ -44,8 +46,12 @@ type Client struct {
 	// muLargeFile ensures only one large file is read/written at a time.
 	// TODO(nodir): ensure this doesn't hurt performance on SSDs.
 	muLargeFile sync.Mutex
-	// Pools of []byte slices with the length of ClientConfig.FileIOSize.
-	fileIOBufs sync.Pool
+
+	// fileBufReaders is a pool of reusable *bufio.Readers
+	// with buffer size = ClientConfig.FileIOSize, so e.g. 4MiB.
+	// Use fileBufReaders.Get(), then reset the reader with bufio.Reader.Reset,
+	// and put back to the when done.
+	fileBufReaders sync.Pool
 
 	// Mockable functions.
 
@@ -224,6 +230,8 @@ func NewClientWithConfig(ctx context.Context, conn *grpc.ClientConn, instanceNam
 	return client, nil
 }
 
+var emptyReader = bytes.NewReader(nil)
+
 // init is a part of NewClientWithConfig that can be done in tests without
 // creating a real gRPC connection. This function exists purely to aid testing,
 // and is tightly coupled with NewClientWithConfig.
@@ -232,8 +240,8 @@ func (c *Client) init() {
 	c.semBatchUpdateBlobs = semaphore.NewWeighted(int64(c.BatchUpdateBlobs.Concurrency))
 
 	c.semFileIO = semaphore.NewWeighted(int64(c.FSConcurrency))
-	c.fileIOBufs.New = func() interface{} {
-		return make([]byte, c.FileIOSize)
+	c.fileBufReaders.New = func() interface{} {
+		return bufio.NewReaderSize(emptyReader, int(c.FileIOSize))
 	}
 }
 
--- a/go/pkg/cas/ioutil.go
+++ b/go/pkg/cas/ioutil.go
@@ -18,35 +18,54 @@ type byteSource interface {
 	Rewind() error
 }
 
+// byteSliceSource implements byteSource on top of []byte.
+type byteSliceSource struct {
+	io.Reader
+	content []byte
+}
+
+func newByteSliceSource(content []byte) *byteSliceSource {
+	return &byteSliceSource{Reader: bytes.NewReader(content), content: content}
+}
+
+func (s *byteSliceSource) Rewind() error {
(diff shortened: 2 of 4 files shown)


go/pkg/digest/digest.go
```diff
@@ -147,16 +147,8 @@ func NewFromFile(path string) (Digest, error) {
...
-	return NewFromReaderWithBuffer(r, make([]byte, 32*1024))
-}
-
-// NewFromReaderWithBuffer computes a file digest from a reader, using the
-// given buffer.
-// It returns an error if there was a problem reading the file.
-func NewFromReaderWithBuffer(r io.Reader, buf []byte) (Digest, error) {
 	h := HashFn.New()
-	size, err := io.CopyBuffer(h, r, buf)
+	size, err := io.Copy(h, r)
```
rubensf, line 151: Why did you rever these?

HashFn doesn't implement `ReadFrom` [1], and your implemented fileSource doesn't implement `WriteTo`.

1: https://golang.org/pkg/crypto/#Hash

go/pkg/cas/ioutil.go
```diff
@@ -1,39 +1,85 @@
...
 
 // This file contains various IO utility types and functions.
 
-// readSeekCloser can Read, Seek and Close.
-// TODO(nodir): use io.ReadSeekCloser after dropping Go 1.15 support.
-type readSeekCloser interface {
-	io.ReadSeeker
+// byteSource is a common interface to in-memory blobs and files.
+// SeekStart is like io.Seeker.Seek, but only supports io.SeekStart.
+type byteSource interface {
```
rubensf, line 16: `source` instead of `byteSource`? I think `byteSource` can imply bytes in memory.

go/pkg/cas/ioutil.go
```diff
@@ -1,39 +1,85 @@
...
+	io.Reader
 	io.Closer
+	SeekStart(offset int64) error
```
rubensf, line 19: We might need to preserve offsets for retries?

rubensf (commented):
I'm not sure if there's many benefits from this?

The only `CopyBuffer` you changed was the one in the digest, but that was actually using the buffer since hashFn doesn't implement ReadFrom, and os.File doesn't implement WriteTo.

The `ReadFrom` from #320 doesn't do internal allocations - it already has some internal buffering, so we don't need to try to optimize it.

go/pkg/cas/client.go
```diff
@@ -232,8 +240,8 @@ func (c *Client) init() {
 	c.semBatchUpdateBlobs = semaphore.NewWeighted(int64(c.BatchUpdateBlobs.Concurrency))
 
 	c.semFileIO = semaphore.NewWeighted(int64(c.FSConcurrency))
-	c.fileIOBufs.New = func() interface{} {
-		return make([]byte, c.FileIOSize)
+	c.fileBufReaders.New = func() interface{} {
+		return bufio.NewReaderSize(emptyReader, int(c.FileIOSize))
```
rubensf, line 244: I think this can be `NewReaderSize(nil, ...)`?

If so, add comments explaining this is meant to always be reset with an actual reader.

rubensf (commented):
Ah, thanks! I missed (2) entirely. Should have read your description closer.

I was trying to convince myself that writing (copying?) the `copyBuffer` code would be simpler/easier to read than using bufio, but after some mulling I think your solution is cleaner (by a small margin). Also some googling shows me that copyBuffer using WriteTo/ReadFrom is a common concern :)

I think my individual comments are all valid still though.

go/pkg/cas/client.go
nodirg: it happens to work (based on my reading of the code), but it is documented. I am slightly hesitant to rely on undocumented behavior, but I can change it to nil if yo prefer.

go/pkg/digest/digest.go
nodirg: Updated this code to use a global pool of buffers. Now packages other than `cas` also benefit from this.

go/pkg/cas/ioutil.go
nodirg: Sorry, could you elaborate? Isn't preserving offsets the job of the caller?

go/pkg/cas/ioutil.go
nodirg: but it is a source of bytes :) How does `uploadSource` sound? `source` is a bit too generic.

go/pkg/cas/client.go
rubensf: emptyReader sgtm. I agree it seems better to not accidentally crash an user because of internal/golang changes handling nils.

go/pkg/cas/ioutil.go
rubensf: My bad again - I read `SeakStart()` without an offset, oops.
Nvm my comment.

go/pkg/cas/ioutil.go
rubensf: heh, `uploadSource` sgtm. Maybe it's just me but `byte` reminds me of `[]byte` :)

go/pkg/digest/digest.go
rubensf: Sgtm

rubensf (approved):

PR #325, opened 2021-04-30
[cas] Implement per-request timeouts in a stream

nodirg:
grpc package does not have native support for per-request timeouts for
streaming RPC. Add a helper function to accomodate that.

The func is not specific to CAS, but it wasn't obvious which package
is a better place for it.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -11,7 +11,6 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/google/uuid"
@@ -899,22 +898,4 @@ func marshalledRequestSize(d *repb.Digest) int64 {
 		reqSize += marshalledFieldSize(int64(d.SizeBytes))
 	}
 	return marshalledFieldSize(reqSize)
-}
-
-// withPerCallTimeout returns a function wrapper that cancels the context if
-// fn does not return within the timeout.
-func withPerCallTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc, func(fn func())) {
-	ctx, cancel := context.WithCancel(ctx)
-	return ctx, cancel, func(fn func()) {
-		stop := make(chan struct{})
-		defer close(stop)
-		go func() {
-			select {
-			case <-time.After(timeout):
-				cancel()
-			case <-stop:
-			}
-		}()
-		fn()
-	}
 }
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/google/uuid"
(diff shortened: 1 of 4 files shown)


go/pkg/cas/upload.go
```diff
@@ -894,3 +900,21 @@ func marshalledRequestSize(d *repb.Digest) int64 {
 	}
 	return marshalledFieldSize(reqSize)
 }
+
+// withPerCallTimeout returns a function wrapper that cancels the context if
+// fn does not return within the timeout.
+func withPerCallTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc, func(fn func())) {
```
rubensf, line 906: I think this should go either on `cas/client.go` for now, or `client/client.go` where there's already something similar: https://github.com/bazelbuild/remote-apis-sdks/blob/5121020d2a436fe8ced44ab65be31d5f0c4a1276/go/pkg/client/client.go#L694

If you put it on `client/client.go`, it probably should wrap a `func(ctx context.Context)` and use a `context.WithTimeout`, as to be more generic (like the existing `CallWithTimeout`).

atetubou:
test is flaky?

go/pkg/cas/upload.go
nodirg: I've moved it to `cas/client.go`. 

I am still trying to avoid a dependency on `client` package, so that `client` package can use `cas` package. But I agree that this is pretty generic, and we should probably add a new package, at some point, with these generic helper funcs, perhaps an internal package.

`(*Client) CallWithTimeout()` is a bit different though. Yes, both are about timeouts, but this func is specifically a helper for funcs that don't accepted context (like `StreamServer.Send`).

nodirg:
@atetubou looks like the flaky test is in the `client` package, which is not changed in this PR.

atetubou (approved):

go/pkg/cas/upload.go
```diff
@@ -788,7 +790,10 @@ chunkLoop:
 		req.Data = buf[:n] // must limit by `:n` in ErrUnexpectedEOF case
 
 		// Send the chunk.
-		switch err = stream.Send(req); {
+		withTimeout(func() {
+			err = stream.Send(req)
```
rubensf, line 794: IIUC, even if this "times out", we'd still have to wait until the hanging request returns so we can stop streaming at the next loop iteration. 

Do you think we can handle this in an more aggressive way? Eg launch the `Send` in a separate goroutine, and moving on to an early return if the time ticks early. I _think_ that even though it doesn't seem like we can forcefully close the stream, letting its resources leak while early returning the error to the user provides a better UX. (FWIW, I'm not sure how complicated the implementation would end up ending up like).

rubensf (approved):
Preemptively approving in case you think the early return isn't worth the complexity.

go/pkg/cas/upload.go
nodirg: we do have to wait for `Send()` to return, but I think it will return quickly because `stream`'s context is canceled. I found a couple of discussions about this: https://github.com/grpc/grpc-go/issues/990 and https://github.com/grpc/grpc-go/issues/3230#issuecomment-562490274 that seem to suggest that `Send()` would be unblocked. FWIU, once the context is canceled, there is not much `StreamClient` can do, other than fail, which is what we are trying to achieve.

Also I am generally worried about leaks :)

rubensf: Ah, good to know that cancelling the ctx does close the stream!

rubensf (approved):

atetubou:
This breaks build in current HEAD?

atetubou:
Fix is easy, made https://github.com/bazelbuild/remote-apis-sdks/pull/329

PR #326, opened 2021-05-11
Add Mtime to the file metadata cache.

ramymedhat:


go/pkg/filemetadata/cache_posix_test.go
```diff
@@ -28,7 +29,7 @@ func TestExecutableCacheLoad(t *testing.T) {
 		Digest:       wantDg,
 		IsExecutable: true,
 	}
-	if diff := cmp.Diff(want, got); diff != "" {
+	if diff := cmp.Diff(want, got, cmpopts.IgnoreFields(Metadata{}, "MTime")); diff != "" {
```
gkousik, line 32: Declare "MTime" as a const and use everywhere?

gkousik (approved):

go/pkg/filemetadata/cache_posix_test.go
ramymedhat: Done.

PR #330, opened 2021-05-17
Create golangci-lint.yml

atetubou:
This is mostly copied from https://github.com/marketplace/actions/run-golangci-lint#how-to-use

Context: http://b/188387884#comment3

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,38 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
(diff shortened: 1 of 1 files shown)


rubensf:
This LGTM, but I don't see any extra checks running - is the check supposed to be running already?

atetubou:
I think we need to submit this first, let me see what happens after that.

rubensf (approved):

atetubou:
Thanks, github action started running in https://github.com/bazelbuild/remote-apis-sdks/runs/2601345692

PR #331, opened 2021-05-17
add .golangci.yml

atetubou:
This updates some default configs for golangci-lint.
https://golangci-lint.run/usage/configuration/#config-file

context: http://b/188387884#comment3


rubensf (approved):
Now I see linting happening! ... with many errors, heh

PR #332, opened 2021-05-18
remove unused variables

atetubou:


go/pkg/cas/client.go
```diff
@@ -44,9 +44,6 @@ type Client struct {
 
 	// TODO(nodir): ensure it does not hurt streaming.
 	semFileIO *semaphore.Weighted
-	// muLargeFile ensures only one large file is read/written at a time.
-	// TODO(nodir): ensure this doesn't hurt performance on SSDs.
-	muLargeFile sync.Mutex
```
rubensf, line 49: I believe @nodirg  still inteds to use this

atetubou: Sorry, I saw lint warning in wrong local branch.

rubensf (approved):

go/pkg/cas/client.go
nodirg: it is used in https://github.com/bazelbuild/remote-apis-sdks/blob/395c674af7a9cd696dfd1f2b4a950f6899ccb3a0/go/pkg/cas/upload.go#L324

PR #334, opened 2021-05-18
[cas] Rename UploadOptions.Callback to Prelude

nodirg:
"Callback" may mean many things. The name does not entail *when* it is
called. The next CL will introduce another callback which is called
*after* a blob is uploaded. Rename "Callback" to "Prelude" to avoid
ambiguity.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -96,20 +96,20 @@ type UploadOptions struct {
 	// This field is ignored if PreserveSymlinks is false, which is the default.
 	AllowDanglingSymlinks bool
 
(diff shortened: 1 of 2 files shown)


rubensf:
Preflight or Prelude?

(personally I like the name "prelude" better but I'm ambivalent at which one is best at conveying information)

nodirg:
Yeah, I was on the fence too. If you are OK with Prelude, I'd prefer that, so that the other one can be called Postlude. Done

rubensf (approved):

nodirg:
thanks for the quick reviews!

PR #335, opened 2021-05-18
[cas] Read files once

nodirg:
Currently PathExclude regexp is incorporated into the cache key
unconditionally. However, pathExclude isn't used if the path is a regular file.
Thus do not incorporate PathExclude into the cache key in this case.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -274,7 +274,9 @@ func (u *uploader) visitPath(ctx context.Context, absPath string, info os.FileIn
 	cacheKey := cacheKeyType{
 		AbsPath: absPath,
 	}
-	if pathExclude != nil {
+	// Incorporate the pathExclude only if the path is a directory.
+	// If it is a regular file or a symlink, then read it only once.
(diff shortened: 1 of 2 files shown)


rubensf:
Why would we not want to exclude regular files?

nodirg:
We do exclude regular files. The cache is checked after the regexp is evaluated.

go/pkg/cas/upload.go
```diff
@@ -274,7 +274,9 @@ func (u *uploader) visitPath(ctx context.Context, absPath string, info os.FileIn
 	cacheKey := cacheKeyType{
 		AbsPath: absPath,
 	}
-	if pathExclude != nil {
+	// Incorporate the pathExclude, unless it is a regular file.
+	// If it is a regular file, then pathExclude is not used below.
```
rubensf, line 278:
```suggestion
	// If it is a regular file, then there's no need to include pathExclude in the cache key, as we already know the
	// regex does not match the file and the exclusion isn't propagated.
```

rubensf (approved):
Ah, my bad!

PR #336, opened 2021-05-19
[cas] Use pathInfo struct

nodirg:
Refactor the code to pass new `pathInfo` struct around which is a tuple of
`absPath` and `os.FileInfo`. This is a preparation for another CL which actually
adds new functionality, and uses the new struct.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -241,24 +241,29 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
 			return errors.WithStack(err)
 		}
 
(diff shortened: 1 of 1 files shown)


atetubou (approved):

rubensf (approved):

nodirg:
so, I've forgot to rebase it on `master` before merging, but also it turned out that this is not really needed - my approach changed.
@rubensf do you still like this PR? Should I re-upload it for master?

nodirg:
I can do just fine without it. Maybe leave it as is for now?

rubensf:
hah, github makes it hard again!
imo just leave it for now.

PR #337, opened 2021-05-19
[cas] Refactor batchBundler

nodirg:
Refactor `batchBundler` to bundle `uploadItem`s instead of
`repb.BatchUpdateBlobsRequest_Request`.
This CL is a pure refactoring, and a preparation for another CL
which actually adds new functionality.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -624,9 +624,9 @@ func (u *uploader) check(ctx context.Context, items []*uploadItem) error {
 }
 
 func (u *uploader) scheduleUpload(ctx context.Context, item *uploadItem) error {
-	reqSize := int(marshalledRequestSize(item.Digest))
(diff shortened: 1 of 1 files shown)


go/pkg/cas/upload.go
```diff
@@ -624,40 +624,46 @@ func (u *uploader) check(ctx context.Context, items []*uploadItem) error {
 }
 
 func (u *uploader) scheduleUpload(ctx context.Context, item *uploadItem) error {
+	reqSize := int(marshalledRequestSize(item.Digest))
```
rubensf, line 627: nit: while not realistic that we'd ever have more than 2^31-1 items, it's technically safer to do `int` -> `int64` than `int64` -> `int`

rubensf (approved):

go/pkg/cas/upload.go

PR #338, opened 2021-05-19
Make TreeSymlinkOpts.Preserved configurable via flag.

yoshisatoyanagisawa:
Some compilers require symlinks are preserved because they use readlink to convert from wrapper names to real compiler names.
Let me make expose the option via flag so that a user of this library can easily configure it via flags.


rubensf:
I think we actually want this done on the internal wrapper so we can control this on an per action basis.

yoshisatoyanagisawa:
Sure.  Let me go with per command basis instead of configure this option globally.
https://github.com/bazelbuild/remote-apis-sdks/pull/343

PR #340, opened 2021-05-19
[cas] Add UploadResult.Digest

nodirg:
The callers of `Client.Upload` need the digest back in order to
pass it to another system which would fetch the uploaded files.
Add `cas.UploadInput.OnDigest` callback, called when a digest is computed.

Note that by the time `OnDigest` is called, the item is not fully uploaded yet.
It is generally hard to detect when a tree is fully uploaded because
file uploads happen in arbitrary order. In particular, a blob for a
parent directory may be uploaded before its children.
Still, this is enough for the purpose.

Returning digests from `Upload()` function was also considered, but it
is getting complicated quickly because the callsite needs to know
uploadItem->digest association.

Another considered alternative is to add something like
`UploadInput.Digest digest.Digest` which is populated by `Upload()`.
This is a bit awkward because Digest is output, while the struct is called
`UploadInput`.

### DIFF
--- a/go/pkg/cache/singleflightcache.go
+++ b/go/pkg/cache/singleflightcache.go
@@ -34,6 +34,20 @@ func (s *SingleFlight) LoadOrStore(key interface{}, valFn func() (interface{}, e
 	return e.val, e.err
 }
 
+// Load is similar to a sync.Map.Load.
+func (s *SingleFlight) Load(key interface{}) (val interface{}, err error, loaded bool) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	var eUntyped interface{}
+	eUntyped, loaded = s.store.Load(key)
+	if loaded {
+		e := eUntyped.(*entry)
+		val = e.val
+		err = e.err
+	}
+	return
+}
+
 // Store forcefully updates the given cache key with val. Note that unlike LoadOrStore,
 // Store accepts a value instead of a valFn since it is intended to be only used in
 // cases where updates are lightweight and do not involve computing the cache value.
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -100,6 +100,12 @@ type UploadOptions struct {
 	Prelude func(absPath string, mode os.FileMode) error
 }
 
+// digested is a result of preprocessing a file/dir.
+type digested struct {
+	dirEntry proto.Message // FileNode, DirectoryNode or SymlinkNode
+	digest   *repb.Digest  // may be nil, e.g. for dangling symlinks
+}
+
 // ErrSkip when returned by UploadOptions.Prelude, means the file/dir must be
 // not be uploaded.
 //
@@ -108,6 +114,46 @@ type UploadOptions struct {
 // result in a dangling symlink.
 var ErrSkip = errors.New("skip file")
 
+// UploadResult is the result of a Client.Upload call.
+// It provides file/dir digests and statistics.
+type UploadResult struct {
+	Stats TransferStats
+	u     *uploader
+}
+
+// Digest returns the digest computed for a file/dir at ps.Path.
+//
+// To retrieve a digest of a regular file, only ps.Path is checked - other
+// fields are ignored.
+//
+// To retrieve a digest of a directory or a symlink, ps.Exclude must match one
+// of the PathSpecs passed to Client.Upload earlier.
+//
+// If the digest is unknown, returns (nil, nil).
+// If the file is a danging symlink, then its digest is unknown.
+func (r *UploadResult) Digest(ps *PathSpec) (*digest.Digest, error) {
+	if !filepath.IsAbs(ps.Path) {
+		return nil, errors.Errorf("%q is not absolute", ps.Path)
+	}
+
+	// TODO(nodir): cache this syscall too.
+	info, err := os.Lstat(ps.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	key := makeFSCacheKey(ps.Path, info.Mode(), ps.Exclude)
+	switch val, err, loaded := r.u.fsCache.Load(key); {
+	case !loaded:
+		return nil, nil
+	case err != nil:
+		return nil, err
+	default:
+		dig := digest.NewFromProtoUnvalidated(val.(*digested).digest)
+		return &dig, nil
+	}
+}
+
 // Upload uploads all files/directories specified by pathC.
 //
 // Close pathC to indicate that there are no more files/dirs to upload.
@@ -115,7 +161,7 @@ var ErrSkip = errors.New("skip file")
 // exits successfully.
 //
 // If ctx is canceled, the Upload returns with an error.
-func (c *Client) Upload(ctx context.Context, opt UploadOptions, pathC <-chan *PathSpec) (stats *TransferStats, err error) {
+func (c *Client) Upload(ctx context.Context, opt UploadOptions, pathC <-chan *PathSpec) (*UploadResult, error) {
 	eg, ctx := errgroup.WithContext(ctx)
 	// Do not exit until all sub-goroutines exit, to prevent goroutine leaks.
 	defer eg.Wait()
@@ -179,7 +225,7 @@ func (c *Client) Upload(ctx context.Context, opt UploadOptions, pathC <-chan *Pa
 			}
 		}
 	})
-	return &u.stats, eg.Wait()
+	return &UploadResult{Stats: u.stats, u: u}, eg.Wait()
 }
 
 // uploader implements a concurrent multi-stage pipeline to read blobs from the
@@ -204,6 +250,8 @@ type uploader struct {
 	wgFS sync.WaitGroup
 
 	// fsCache contains already-processed files.
+	// A key can be produced by makeFSCacheKey.
+	// The values are of type *digested.
 	fsCache cache.SingleFlight
 
 	// checkBundler bundles digests that need to be checked for presence on the
@@ -240,11 +288,39 @@ func (u *uploader) startProcessing(ctx context.Context, ps *PathSpec) error {
 	return nil
 }
 
+// makeFSCacheKey returns a key for u.fsCache.
+func makeFSCacheKey(absPath string, mode os.FileMode, pathExclude *regexp.Regexp) interface{} {
+	// The structure of the cache key is incapsulated by this function.
+	type cacheKey struct {
+		AbsPath       string
+		ExcludeRegexp string
+	}
+
+	key := cacheKey{
+		AbsPath: absPath,
+	}
+
+	if mode.IsRegular() {
+		// This is a regular file.
+		// Its digest depends only on the file path (assuming content didn't change),
+		// so the cache key is complete. Just return it.
+		return key
+	}
+	// This is a directory and/or a symlink, so the digest also depends on fs-walk
+	// settings. Incroporate those too.
+
+	if pathExclude != nil {
+		key.ExcludeRegexp = pathExclude.String()
+	}
+	return key
+}
+
 // visitPath visits the file/dir depending on its type (regular, dir, symlink).
 // Visits each file only once.
 //
 // If the file should be skipped, then returns (nil, nil).
-func (u *uploader) visitPath(ctx context.Context, absPath string, info os.FileInfo, pathExclude *regexp.Regexp) (dirEntry proto.Message, err error) {
+// The returned digested.digest may also be nil if the symlink is dangling.
+func (u *uploader) visitPath(ctx context.Context, absPath string, info os.FileInfo, pathExclude *regexp.Regexp) (*digested, error) {
 	// First, check if the file passes all filters.
 	if pathExclude != nil && pathExclude.MatchString(filepath.ToSlash(absPath)) {
 		return nil, nil
@@ -259,31 +335,21 @@ func (u *uploader) visitPath(ctx context.Context, absPath string, info os.FileIn
 		}
 	}
 
-	// Make a cache key.
-	type cacheKeyType struct {
-		AbsPath       string
-		ExcludeRegexp string
-	}
-	cacheKey := cacheKeyType{
-		AbsPath: absPath,
-	}
-	// Incorporate the pathExclude, unless it is a regular file.
-	// If it is a regular file, then there's no need to include pathExclude in the
-	// cache key, as we already know the regex does not match the file and the
-	// exclusion isn't propagated.
-	if pathExclude != nil && !info.Mode().IsRegular() {
-		cacheKey.ExcludeRegexp = pathExclude.String()
-	}
-
-	node, err := u.fsCache.LoadOrStore(cacheKey, func() (interface{}, error) {
+	cacheKey := makeFSCacheKey(absPath, info.Mode(), pathExclude)
+	cached, err := u.fsCache.LoadOrStore(cacheKey, func() (interface{}, error) {
 		switch {
 		case info.Mode()&os.ModeSymlink == os.ModeSymlink:
 			return u.visitSymlink(ctx, absPath, pathExclude)
+
 		case info.Mode().IsDir():
-			return u.visitDir(ctx, absPath, pathExclude)
+			node, err := u.visitDir(ctx, absPath, pathExclude)
+			return &digested{dirEntry: node, digest: node.GetDigest()}, err
+
 		case info.Mode().IsRegular():
-			// Code above assumes that pathExclude is not used here.
-			return u.visitRegularFile(ctx, absPath, info)
+			// Note: makeFSCacheKey assumes that pathExclude is not used here.
+			node, err := u.visitRegularFile(ctx, absPath, info)
+			return &digested{dirEntry: node, digest: node.GetDigest()}, err
+
 		default:
 			return nil, fmt.Errorf("unexpected file mode %s", info.Mode())
 		}
@@ -291,7 +357,7 @@ func (u *uploader) visitPath(ctx context.Context, absPath string, info os.FileIn
 	if err != nil {
 		return nil, err
 	}
-	return node.(proto.Message), nil
+	return cached.(*digested), nil
 }
 
 // visitRegularFile computes the hash of a regular file and schedules a presence
@@ -434,23 +500,26 @@ func (u *uploader) visitDir(ctx context.Context, absPath string, pathExclude *re
 				u.eg.Go(func() error {
 					defer wgChildren.Done()
 					defer u.wgFS.Done()
-					node, err := u.visitPath(ctx, absChild, info, pathExclude)
+					digested, err := u.visitPath(ctx, absChild, info, pathExclude)
 					mu.Lock()
 					defer mu.Unlock()
-					if err != nil {
+
+					switch {
+					case err != nil:
 						subErr = err
 						return err
+					case digested == nil:
+						// This file should be ignored.
+						return nil
 					}
 
-					switch node := node.(type) {
+					switch node := digested.dirEntry.(type) {
 					case *repb.FileNode:
 						dir.Files = append(dir.Files, node)
 					case *repb.DirectoryNode:
 						dir.Directories = append(dir.Directories, node)
 					case *repb.SymlinkNode:
 						dir.Symlinks = append(dir.Symlinks, node)
-					case nil:
-						// This file should be ignored.
 					default:
 						// This condition is impossible because all functions in this file
 						// return one of the three types above.
@@ -485,7 +554,9 @@ func (u *uploader) visitDir(ctx context.Context, absPath string, pathExclude *re
 // of the target file.
 // If u.PreserveSymlinks is true, then returns a SymlinkNode, otherwise
 // returns the directory node of the target file.
-func (u *uploader) visitSymlink(ctx context.Context, absPath string, pathExclude *regexp.Regexp) (proto.Message, error) {
+//
+// The returned digested.digest is nil if the symlink is dangling.
+func (u *uploader) visitSymlink(ctx context.Context, absPath string, pathExclude *regexp.Regexp) (*digested, error) {
 	target, err := os.Readlink(absPath)
 	if err != nil {
 		return nil, errors.Wrapf(err, "os.ReadLink")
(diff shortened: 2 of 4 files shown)


ramymedhat:
Can't the caller use the file metadata cache to get the digest after uploading is done?

rubensf:
Have you considered using a filemetadata cache instead? Like [in the current loadFiles](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/client/tree.go#L122)

It'd be a bit more awkward/more work since you don't digest and stat all files at the same time, but we use it quite heavily internally.

go/pkg/cas/upload.go
```diff
@@ -64,6 +64,19 @@ type UploadInput struct {
...
+	// The item is not yet uploaded when OnDigest is called.
+	// It is guaranteed to be fully uploaded only after Client.Upload returns
+	// successfully.
+	//
+	// OnDigest is not guaranteed to be called, e.g.
+	//  - the item is skipped via PathExclude or UploadOptions.Prelude
+	//  - the file at Path is a dangling symlink.
+	//
+	// OnDigest functions are called concurrently.
+	OnDigest func(digest.Digest)
```
atetubou, line 79: better to make it possible to return error from this function?

nodirg:
@rubensf @ramymedhat I have a few questions.

`UploadItems` accepts an in-memory blob (i.e. no files). How can filemetadata cover that case? i.e. the key is not necessarily a path. It sounds like you use the filemetadata approach elsewhere - how do you solve this problem with filemetadata?

Do you mean use `filemetadata` package or the approach "accumulate what we've collected, and then make it available to the callsite"? For the former, I think we've discussed why `cas` doesn't use `filemetadata` package before, but I cannot find it. I looked at it again:
-  It does not actually compute digests for a directory, and most of our inputs are directories. Not sure how it would support directories - it does not know about `PathExclude` or callback-based exclusions, nor it incorporates `PathExclude` into the cache key.
- it is currently using a global cache without expirations - this wouldn't work for server-side apps, but it can be changed.
- it calls `filepath.Abs` unnecessarily, but it can be fixed
- Reusing it would require significant changes in the caching implementation of uploader

go/pkg/cas/upload.go

nodirg:
hmm, looking at isolate code, it looks like I actually need directory entry protos (e.g. `FileNode`), no just digest; but let's keep the discussion going before I upload a new CL (still thinking about it too).

ramymedhat:
> hmm, looking at isolate code, it looks like I actually need directory entry protos (e.g. `FileNode`), no just digest; but let's keep the discussion going before I upload a new CL (still thinking about it too).

We use the filemetadata cache in reproxy for use cases related to dependency validation, and since the SDK will compute digests for uploading to RBE, we use the same cache so the digest computation is done once. Iirc, we did not use the fmc in the Uploader because the cache implementation was simple enough, but we ended up using the singleflight cache which is essentially what I and I guess Rubens are proposing.

I don't think I understand the isolate use case fully. Some questions:
1. Do you need the digest of blobs that only exist in memory?
2. Based on your latest comment, would the fsCache work?
3. Do you need asynchronous callbacks, will there be work to do in `onDigest` that would benefit from concurrency?

I'm not sure what would happen at the callsite with the `onDigest` functions, but if its sufficient to get all the digests at the end, then seems like maintaining a map in the `uploader` and returning it in `Upload` would simplify the callsite as well?

rubensf:
@nodirg AFAIK in practice, we don't ever upload in memory blobs - while the functionality is there, it's not useful for the sake of remote compilation. For this Upload structure, we could just save that data in filemetadata?

In any case, we do need a way to make information on files available for other remote execution related purposes - correct me if I'm wrong, but stat calls are somewhat expensive, particularly on Windows. Digesting surely is particularly expensive as it requires reading the whole file. 

I vaguely remember talking about using that before, and I think you had some issue the assumption the a filepath always maps to the same digest. I think you proposed passing a filemetadata cache as an input so we could eg pass an empty one as needed.

I don't think we need to use the current filemetadata cache per se but we do need to cache stats and digests and other options would bean much more migration work.

nodirg:
> hmm, looking at isolate code, it looks like I actually need directory entry protos (e.g. FileNode), no just digest; but let's keep the discussion going before I upload a new CL (still thinking about it too).

ignore this. Just a digest is fine.

> Do you need the digest of blobs that only exist in memory?

No. Also, in hindsight, uploading in-memory blobs is simple enough and different enough that removing this feature from `Client.Upload` would fine with me.

> Based on your latest comment, would the fsCache work?

I assume you mean use existing fsCache to let the user retrieve digests *after* `Upload()` returns.
Yes. We could change `Client.Upload` return type from `TransferStats` to something like

```go
type UploadResults struct {
  Stats TransferStats
  // some internal fields here
}

// Digest returns the digest computed for the upload input.
func (r *UploadResults) Digest(*UploadInput) digest.Digest { ... }
```

Note that just `absPath` parameter wouldn't be enough because the digest depends on `PathExclude` and other filtering things we might add to `UploadInput` (and I need to add soon).

It is a bit weird to use `UploadInput` this way. If we ditch `UploadInput.Content`, perhaps we could rename `UploadInput` to something like `PathSpec`, and then it would be less awkward:

```go
func (r *UploadResults) Digest(*PathSpec) digest.Digest { ... }
```


WDYT?

> Do you need asynchronous callbacks, will there be work to do in onDigest that would benefit from concurrency?

No. The func signature would be fine too.

> In any case, we do need a way to make information on files available for other remote execution related purposes - correct me if I'm wrong, but stat calls are somewhat expensive, particularly on Windows. Digesting surely is particularly expensive as it requires reading the whole file.

Besides this CL, I also have another one locally, with `UploadOptions.Postlude` callback, which is called on each file once the digest is computed. But that would mean we have two ways of doing similar thing (`OnDigest` and `Postlude`). The above approach, `UploadResults.Digest(PathSpec)`, would cover both cases. WDYT?

Note: it doesn't preclude us from passing reusable cache later. It just decouples caching from retrieving digests.
Otherwise, the user is forced to pass a cache, even if they don't actually care about reusing the cache across `Upload` calls, just to be able to retrieve digests back.

I also like `UploadResults` because we can add more stuff there later, which is not necessarily related to stats. WDYT?

nodirg:
@rubensf I also noticed that `client` package supports "VirtualInputs", where one can provide a path to a non-existing file, and its content. This is currently not supported by `cas` package. Could you confirm that this is not actually used anywhere? 
If it is still used, then we might need to provide some sort of virtual file system, perhaps based on [fs package](https://pkg.go.dev/io/fs).

ramymedhat:
+1 to `UploadResults` and `PathSpec`. Seems generally useful and a net improvement.

Ack regarding the cache. We're currently adding features to reproxy that would benefit from not needing to recompute the digest in the sdk. So not needed now but once the client package uses the cas package, we would need the fmc to be reused as well.

We use virtual inputs in many places. Some actions require a directory to exist on the bot, even though we don't want to upload everything under it. So we add it as a virtual input. The fs package seems like overkill for this use case.

nodirg:
Thinking about this a bit more, `fs` wouldn't work well because we might need to different content for the same virtual files in different uploads => fs parameter cannot be specified in `UploadOptions`; it has to be in `PathSpec`, which will probably mean pathspecs wouldn't reuse each others' cache.

It is just felt a bit weird if `PathSpec` contains *content* of files. At least, in my mental model `PathSpec` only specifies a subset of the file system to read, whereas *the content* of files is on the file system; but I guess I will have to change my mental model. It feels like adding content to pathSpec makes it less generic, but we can deal with that.

TIL isolate needs partial uploads of a directory: we don't upload the entire directory, but only certain subdirectories/subfiles and they don't have to have common parent, e.g. it could be `./foo/dir1`, `./foo/dir2`, `./bar.exe`. So I was going to add optional `PathSpec.Allowlist` (note: regexp would work in this case).

nodirg:
Uploaded https://github.com/bazelbuild/remote-apis-sdks/pull/342 for the `UploadInput` -> `PathSpec` part, and removal of `UploadInput.Content`.

nodirg:
PTAL.

- I've replaced `OnDigest` with `UploadResults` which as `Digest(PathSpec) (*digest.Digest, error)` function
- Along the way, I also made the FS cache reusable across `Client.Upload` calls.
- Ideally we cache `os.FileInfo`, so that we don't another `LStat` in `FSCache.Digest()`. This doesn't matter for Chrome since we need digests of only roots and we have few of them, but it might matter for reclient (does it?).

rubensf:
> No. Also, in hindsight, uploading in-memory blobs is simple enough and different enough that removing this feature from Client.Upload would fine with me.

Ramy, can you confirm if that's something we use? Despite for eg tests. Otherwise I'm also fine without it. Certainly simplifies the code. Echo +1 for UploadResults, even if it only has TransferStats.

> Otherwise, the user is forced to pass a cache, even if they don't actually care about reusing the cache across Upload calls, just to be able to retrieve digests back.

Personally, I don't think that's bad practice - the user can pass `nil` if he doesn't want to reuse a cache. I've seen this strategy in quite a few golang codebases particularly for re-utilizing buffers.

> @rubensf I also noticed that client package supports "VirtualInputs", where one can provide a path to a non-existing file, and its content. This is currently not supported by cas package. Could you confirm that this is not actually used anywhere?

We actually do require virtual inputs and use extensively. The main use case I'm aware of are for tools either don't create directories for outputs or complain if they can't find a directory (even if it's empty). I don't think it requires something as involved io/fs.

> Ideally we cache os.FileInfo, so that we don't another LStat in FSCache.Digest(). This doesn't matter for Chrome since we need digests of only roots and we have few of them, but it might matter for reclient (does it?).

It does :)


(reading the code now)

rubensf (commented):
> Thinking about this a bit more, fs wouldn't work well because we might need to different content for the same virtual files in different uploads => fs parameter cannot be specified in UploadOptions; it has to be in PathSpec, which will probably mean pathspecs wouldn't reuse each others' cache.

I'm confused on what you mean here by virtual files - is that an isolate concept? I don't think you're referring to VirtualInputs.

Somehow we didn't mention this explicitly in earlier comments, but we do have to access file digests concurrently with uploads - reproxy still receives and process requests while uploads happen. Offering a `Digest` function after the end of execution wouldn't be enough for us.

It also went unsaid, but reclient may need/use the digest/stat of a file before it's sent for upload. It doesn't seem like the current approach is going in a direction that'd be able to make use of that.

I have some thoughts on how we could do this instead:
* The current `fsCache cache.SingleFlight` can store digests and stats. You'd have to move Stat'ing the file inside the locked access for the file; 
* Long running processes can call upload with new cache objects/`nil` to "invalidate" the cache, and/or add other methods to invalidate entries (TTL? Occasional mtime checks? Manual triggers?).
* Other code uses _always_ have to rely on this package to compute digests and stats, least be unable to get directory digests;
* We'd need to break down a way to _walk_ files to process them, and way to process (potentially cached) + upload them. 

Wdyt?

As a side note, I realized that the current approach to build `DirectoryNodes` might need to be tweaked for reclient usage. Particularly, we want the minimal set of files that affects an action. So we may send two different entries as `foo/baz.code` and `foo/bar.code`, while there may be any number of files under `foo/`. IIUC the current approach would try creating two directory nodes for `foo/`.

nodirg:
By VirtualFiles I meant VirtualInputs. I didn't realize `VirtualInput` also supports empty directories. Now I understand why it is called VirtualInput, and not more concrete VirtualFile (although `os` package uses "file" for both files and dirs).

> Somehow we didn't mention this explicitly in earlier comments, but we do have to access file digests concurrently with uploads.

This sounds like we should go back to the callback approach that this PR started with, right?

> It also went unsaid, but reclient may need/use the digest/stat of a file before it's sent for upload. It doesn't seem like the current approach is going in a direction that'd be able to make use of that.

This is partially supported today: Prelude is called before upload, and the caller has a chance to skip a file.

Could you elaborate on the use case? Why does it need the digest necessarily *before* the upload?

> The current fsCache cache.SingleFlight can store digests and stats. You'd have to move Stat'ing the file inside the locked access for the file;

stat depends only on the file path, but fsCache key currently contains more than the file path. I was actually planning to add another field into FSCache, that would map from the absolute file path to `os.FileInfo`. I was going to do it in order to avoid the `Lstat` in `Digest()` func, but it can be used for other purposes as well.

> Long running processes can call upload with new cache objects/nil to "invalidate" the cache, and/or add other methods to invalidate entries (TTL? Occasional mtime checks? Manual triggers?).

SGTM - this seems consistent/possible with the current approach, right? The caller can pass `UploadOptions.FSCache = nil` if they want to. We can add these methods/checks to `FSCache` in the future (there is a TODO about it).

> Other code uses always have to rely on this package to compute digests and stats, least be unable to get directory digests;

Yeah, it sounds like you want some sort of "dry upload", where we walk the FS and computed digests, but don't actually upload anything. It can be a separate method in `Client`, but with shared implementation. I like the idea, and it seems consistent with what we have now (except for the part that we need digests sooner, NOT in the very end).

> As a side note, I realized that the current approach to build DirectoryNodes might need to be tweaked for reclient usage. Particularly, we want the minimal set of files that affects an action. So we may send two different entries as foo/baz.code and foo/bar.code, while there may be any number of files under foo/. IIUC the current approach would try creating two directory nodes for foo/.

Yeah, isolate needs this too. See above, I plan to add `PathSpec.Allowlist` for that. It would equivalent to deleting all not-allowed files/dirs before uploading.

go/pkg/cas/fscache.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Digest returns the digest computed for a file/dir at ps.Path.
+//
+// In the context of upload, ps.Exclude must match one of the PathSpecs passed
+// to Client.Upload earlier.
+//
+// If the digest is unknown, returns (nil, nil).
+// If the file is a danging symlink, then its digest is unknown.
+func (c *FSCache) Digest(ps *PathSpec) (*digest.Digest, error) {
+	absPath, err := filepath.Abs(ps.Path)
```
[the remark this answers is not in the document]
nodirg: I've sent out https://github.com/bazelbuild/remote-apis-sdks/pull/345 to require `PathSpec.Path` to be absolute, but for a different reason.

I am not sure I understood your reasoning though. Most package I know, including the lowest-level `os` package, accept relatives. Does your argument apply to them too?

rubensf:
> This is partially supported today: Prelude is called before upload, and the caller has a chance to skip a file.
Could you elaborate on the use case? Why does it need the digest necessarily before the upload?

For instance, we currently are working on a dependencies cache across builds - before calculating dependencies, we check if the digests of all the file dependencies for the same compilation in a previous build are still the same. If any file changed, we have to re-calculate dependencies, which can be much more CPU intensive. Chances are that we'd still need to upload the current version of the file, but not necessarily. 

> This sounds like we should go back to the callback approach that this PR started with, right?
FWIW I also considered UploadOptions.Postlude of type func(absPath string, mode os.FileMode, digest Digest) which is called after each file is uploaded, along with digest.

By concurrent, I mean also independently. Many actions may be taking place that require the same dependency file.

> stat depends only on the file path, but fsCache key currently contains more than the file path. I was actually planning to add another field into FSCache, that would map from the absolute file path to os.FileInfo. I was going to do it in order to avoid the Lstat in Digest() func, but it can be used for other purposes as well.

We definitely also need access to Stats in the same way we need it for digests. FWIW we only ever need that for files (ie not directories), so sticking them in the same object would work for our use case. But alas, the entire point of introducing this package is that reclient isn't the only client :)

I generally liked using `fsCache` for simplicity, though I'm afraid you're correct that we do need to account for path exclusions when producing a directory proto.  I'm not opposed to having eg a second cache object  - I doubt we'd ever have enough files for the memory cost of duplicating the path name in the second cache to be substantial. 

> SGTM - this seems consistent/possible with the current approach, right? The caller can pass UploadOptions.FSCache = nil if they want to. We can add these methods/checks to FSCache in the future (there is a TODO about it).

Yeah, I think this is consistent with the current approach.

> Yeah, isolate needs this too. See above, I plan to add PathSpec.Allowlist for that. It would equivalent to deleting all not-allowed files/dirs before uploading.

Ah, sorry I missed that comment. A regex does technically work, but I wonder how that'd work exactly? We'd craft a massive regex to catch all of the specific 10-500 files in the entire source tree? I guess that'd technically work... But we'd _still_ need to do the work of "constructing" a tree of directories regardless, but have the overhead of matching regexes. 

I wonder if instead we could punt digesting the directory protos until after the inputs channel is closed? That'd mean that directory cache keys would depend on the entire inputs channel - a much weirder way to represent that key than a regex (though technically they'd just be different representations of the same information).

FWIW I think this was my main confusion with this PR - we don't ever need digests of _directories_ outside uploads, and we don't need path exclusions, so I struggled updating my mental model of requirements :sweat_smile:

go/pkg/cas/fscache.go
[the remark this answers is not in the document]
rubensf: My argument does apply to them too - relative paths are tricky because we have clear definitions of "input/exec root" and a "working directory", but, AFAIK, no expectations for the running directory of the binary using the SDK.

nodirg:
Rubens, I think it would help if both Chrome and reclient listed their requirements for getting the digest back. Then we can produce a design that satisfies them all. I have *some* understanding of reclient requirements based on your comments, but I'd prefer to avoid interpretation. Could you provide a concrete list of requirements?

Chrome's requirements are very simple: get the digest of the uploaded PathSpec after the upload is complete.

rubensf:
1. We need digests for files before, currently and after uploading them (that's what I meant by "concurrently"); 
2. We need `Stat` for files concurrently;
3. We should avoid to re-digest or re-stat files;
4. AFAIK, we need neither digests nor stats for directories on reclient.
    * Additionally, the majority of our uploads are single files that may or may not share path prefixes, that have to be "merged" in their directory prefixes at some point.

I'm only familiar insofar as building chrome with reclient though, so @ramymedhat might know more requirements, but he's OOO until Tue (Mon is a holiday).

(chrome -> isolate? :)

nodirg:
We met today and I believe agreed on the following:
* Distinguish short-term and long-term goals: unblock isolate soon, while keeping the public API extensible for re-client requirements and uses cases in the future. In particular, `FSCache` type was removed from this PR.
* `UploadResult.Digest()` API meets isolate's requirements today and can meet re-client's requirements in the future; so it is appropriate to land to unblock isolate.
* Later, reuse `filemetadata` package in `cas` package in all places where we do `LStat` syscalls or compute digest of a regular file. In particular, use that to fix the TODO about caching LStat calls.
* In the future, reclient might need to be able to pass a callback/channel via `PathSpec`, to be notified when the directory is uploaded. This will address the use case that is currently addressed by "unified uploads" (many parallel actions might be uploading the same files). In this use case, reclient needs only digests of specified directories, not every file.
* The way `UploadOptions.FSCache` was implemented in this PR wasn't correct: it was possible that two uploads try to upload the same file, where one fails (because it actually tried to upload the file and failed) and one succeeded (because it assumes that the other uploader succeeded).

I've updated the PR accordingly. I also rebased it on top of https://github.com/bazelbuild/remote-apis-sdks/pull/345.

go/pkg/cache/singleflightcache.go
```diff
@@ -34,6 +34,20 @@ func (s *SingleFlight) LoadOrStore(key interface{}, valFn func() (interface{}, e
...
 
+// Load is similar to a sync.Map.Load.
+func (s *SingleFlight) Load(key interface{}) (val interface{}, err error, loaded bool) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	var eUntyped interface{}
+	eUntyped, loaded = s.store.Load(key)
+	if loaded {
+		e := eUntyped.(*entry)
+		val = e.val
```
ramymedhat, line 45: Can you make the use of named return vars consistent in this file? They're used here but not in LoadOrStore

go/pkg/cas/upload.go
```diff
@@ -108,14 +114,53 @@ type UploadOptions struct {
...
+type UploadResult struct {
+	Stats TransferStats
+	u     *uploader
+}
+
+// Digest returns the digest computed for a file/dir at ps.Path.
+//
+// To retrieve a digest of a regular file, only ps.Path is required.
+//
+// To retrieve a digest of a directory or a symlink, ps.Exclude must match one
```
ramymedhat, line 128: All of ps should match one that was previously uploaded, right? I think in general the passed `ps` should match one that was passed to upload, regardless of whether its a file or a directory? That would make the comment simpler?

ramymedhat (approved):

go/pkg/cache/singleflightcache.go
```diff
@@ -34,6 +34,20 @@ func (s *SingleFlight) LoadOrStore(key interface{}, valFn func() (interface{}, e
 	return e.val, e.err
 }
 
+// Load is similar to a sync.Map.Load.
+func (s *SingleFlight) Load(key interface{}) (val interface{}, err error, loaded bool) {
```
atetubou, line 38: I feel returning err in the last element is consistent.

go/pkg/cas/upload.go
```diff
@@ -108,14 +114,53 @@ type UploadOptions struct {
...
+// UploadResult is the result of a Client.Upload call.
+// It provides file/dir digests and statistics.
+type UploadResult struct {
+	Stats TransferStats
+	u     *uploader
+}
+
+// Digest returns the digest computed for a file/dir at ps.Path.
+//
+// To retrieve a digest of a regular file, only ps.Path is required.
```
rubensf, line 126: I'd go as far as saying "only ps.Path is checked - other fields are ignored" (pending I haven't checked your work on Allowlists yet)

rubensf (approved):
LGTM from my side (pending other comments)

go/pkg/cache/singleflightcache.go
nodirg: I am worried about an invalid interpretation of the signature. Typically, the returned err overrides the preceding return values, so the callsite checks err first. However in this case, `loaded` has higher precedence. The client side must check `loaded` first, and only then `err`, and then `val.` I am worried that if we return `val, loaded, err,` then the programmer would check values in the wrong order. WDYT?

go/pkg/cas/upload.go

go/pkg/cas/fscache.go
```diff
@@ -0,0 +1,85 @@
...
+	ExcludeRegexp string
+}
+
+type fsCacheValue struct {
+	dirEntry proto.Message // FileNode, DirectoryNode or SymlinkNode
+	digest   *repb.Digest  // may be nil if a symlink is dangling
+}
+
+// makeKey returns a key for FSCache.c.
+func (c *FSCache) makeKey(absPath string, mode os.FileMode, pathExclude *regexp.Regexp) fsCacheKey {
```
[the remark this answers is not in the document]
nodirg: essentially done

go/pkg/cas/fscache.go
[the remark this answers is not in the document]
nodirg: the path is absolute now.

go/pkg/cache/singleflightcache.go
atetubou: I see.

go/pkg/cache/singleflightcache.go
nodirg: yes, in a separate CL

go/pkg/cas/upload.go
nodirg: I am pretty sure I responded to this, but I cannot find it now...

>  I think in general the passed ps should match one that was passed to upload, regardless of whether its a file or a directory? 

no necessarily. Suppose `&PathSpec{ Path: "/foo/"}` was passed to `Client.Upload()`. The client may call `UploadResult.Digest(&PathSpec{ Path: "/foo/bar"})` although they never called `Upload` with that pathSpec.

Consider a more nuanced case.

```go
// For each file, get a directory containing that file also "/foo/always".
for _, f := range files {
  pathC <- &PathSpec{ Path: "/foo/", Allowlist: []string{f, "/foo/always"}}
}
```
Now suppose that the client, for whatever reason, wants digest of "/foo/always". It is simpler to call `UploadResult.Digest(&PathSpec{ Path: "/foo/always"})` than savings on of the previously-passed allowlist and then using that. The digest of "/foo/always" does not depend on the allowlist.

WDYT?

> That would make the comment simpler?

I agree that it would, but at the cost of more code, which seems like a higher cost.

atetubou (commented):
change target branch to master?

go/pkg/cas/upload.go
ramymedhat: Ack.

ramymedhat (approved):

atetubou:
It would be good if you can address link checks.

nodirg:
@atetubou addressed in https://github.com/bazelbuild/remote-apis-sdks/pull/357

PR #341, opened 2021-05-19
[cas] PathExclude: use forward slashes

nodirg:
Do not force users to make `PathExclude` conditional on the OS,
by normalizing paths to use forward-slashes before checking `PathExclude`.


rubensf (approved):
I think this is good, but in the spirit of "not caring about the OS" we technically should also accommodate regex'es that use backslash for path separators on unix? :)

Tbh that sounds awful in practice, so I'd just change the PR description and comments to say something else instead of "not caring about the OS".

PR #343, opened 2021-05-20
Preserving symlink or not can be configured from Command.

yoshisatoyanagisawa:
For some toolchains, a user need to run a wrapper to run a compiler.  Such a wrapper sometimes use readlink inside and resolving symlink breaks such a script.  To make it work, we need to preserve symlinks.  Since the behavior is only needed for certain toolchains, let me make it configurable via Command.InputSpec message.


go/api/command/command.proto
```diff
@@ -87,6 +87,17 @@ message VirtualInput {
   bool is_empty_directory = 4;
 }
 
+message SymlinkBehaviorType {
+  enum Value {
```
atetubou, line 91: Why do you need this nest?

Also follow
https://developers.google.com/protocol-buffers/docs/style#enums
?

yoshisatoyanagisawa: I followed how other enums are configured.  To be honest, I do not understand why this is needed.  I just tried to guess the coding style here from CommandResultStatus and InputType.

yoshisatoyanagisawa:
I focused so much on the main logic X(
I will add tests right away.

go/pkg/command/command.go
```diff
@@ -45,6 +45,33 @@ func (s InputType) String() string {
 	return fmt.Sprintf("InvalidInputType(%d)", s)
 }
 
+// SymlinkBehavior represents how symlinks are handled.
```
atetubou, line 48: SymlinkBehaviorType?

go/pkg/command/command.go
```diff
@@ -45,6 +45,33 @@ func (s InputType) String() string {
...
+	UnspecifiedSymlinkBehavior SymlinkBehaviorType = iota
+
+	// ResolveSymlink means symlinks are resolved.
+	ResolveSymlink
+
+	// PreserveSymlink means symlinks are kept as-is.
+	PreserveSymlink
+
+	// SymlinkBehaviorSentinel represents the end of SymlinkBehavior.
+	SymlinkBehaviorSentinel
```
atetubou, line 63: no needs to be public?

go/pkg/command/command.go
```diff
@@ -45,6 +45,33 @@ func (s InputType) String() string {
...
+)
+
+var symlinkBehaviorType = [...]string{"UnspecifiedSymlinkBehavior", "ResolveSymlink", "PreserveSymlink"}
+
+func (s SymlinkBehaviorType) String() string {
+	if UnspecifiedSymlinkBehavior <= s && s < SymlinkBehaviorSentinel {
```
atetubou, line 69:
```suggestion
	if UnspecifiedSymlinkBehavior <= s && s < len(symlinkBehaviorType) {
```

go/pkg/command/command.go
```diff
@@ -45,6 +45,33 @@ func (s InputType) String() string {
 	return fmt.Sprintf("InvalidInputType(%d)", s)
 }
 
+// SymlinkBehavior represents how symlinks are handled.
+type SymlinkBehaviorType int
+
+const (
+	// UnspecifiedSymlinkBehavior means following clients.TreeSymlinkOpts
+	// or DefaultTreeSymlinkOpts if clients.TreeSymlinkOpts is null.
+	UnspecifiedSymlinkBehavior SymlinkBehaviorType = iota
```
atetubou, line 54: no needs to be public?

go/pkg/command/command.go
yoshisatoyanagisawa: good catch.

go/pkg/command/command.go
yoshisatoyanagisawa: yeah, that sounds better.

go/pkg/command/command.go
yoshisatoyanagisawa: I am not so comfortable to reveal a value nobody outside of the package can compare.  Let me defer to the repository owner for the decision.

go/pkg/command/command.go
```diff
@@ -45,6 +45,30 @@ func (s InputType) String() string {
...
+	ResolveSymlink
+
+	// PreserveSymlink means symlinks are kept as-is.
+	PreserveSymlink
+)
+
+var symlinkBehaviorType = [...]string{"UnspecifiedSymlinkBehavior", "ResolveSymlink", "PreserveSymlink"}
+
+func (s SymlinkBehaviorType) String() string {
+	if UnspecifiedSymlinkBehavior <= s && s < SymlinkBehaviorType(len(symlinkBehaviorType)) {
```
atetubou, line 66: better to cast `s` to `int`?

go/pkg/command/command.go
ramymedhat: I think this being public is fine and consistent with the rest of the file.

go/api/command/command.proto
```diff
@@ -87,6 +87,17 @@ message VirtualInput {
...
+    // Use Client.TreeSymlinkOpts or default if it is not set.
+    UNSPECIFIED = 0;
+    // Resolves symlinks.
+    RESOLVE = 1;
```
rubensf, line 95: I think we should just leave the default as "resolve" (eg resolve = 0). On inputs, unspecified actually has a meaning.

Also I think "follow" might be a more descriptive name.

go/api/command/command.proto
```diff
@@ -104,6 +115,9 @@ message InputSpec {
 
   // Environment variables the command relies on.
   map<string,string> environment_variables = 4;
+
+  // symlinks are handled in this way.
```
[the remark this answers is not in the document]
rubensf: And start sentence with capital letters.

go/pkg/command/command.go
yoshisatoyanagisawa: good catch.

go/pkg/command/command.go
yoshisatoyanagisawa: Hmm, instead of trying to write something different, let me follow how InputType is written.

yoshisatoyanagisawa:
PTAL

go/api/command/command.proto
rubensf: Ah, my bad, I think you're right on both accounts!

rubensf (approved):

yoshisatoyanagisawa:
I do not have  a write access to this repository.
Will somebody with the permission merge this?

Thank you.

PR #345, opened 2021-05-21
[cas] Require PathSpec.Path to be absolute

nodirg:
A future CL with all PathSpec.Allowlist which is a list of
PathSpec.Path's descendants. It makes more sense for such paths to
be relative to PathSpec.Path than to CWD, but it would be confusing to
have/allow one struct (PathSpec) to contais paths relative to different
dirs.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -44,7 +44,7 @@ var zstdEncoders = sync.Pool{
 // PathSpec specifies a subset of the file system.
 type PathSpec struct {
 	// Path to the file or a directory to upload.
-	// Must be absolute or relative to CWD.
+	// Must be absolute.
 	Path string
 
 	// Exclude is a file/dir filter. If Exclude is not nil and the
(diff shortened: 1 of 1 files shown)


rubensf:
I'll hold off "reviewing" this until we have more agreement on #340. While I don't see any current issue with using absolute paths, I have a low key preference of saving things relative to a specified "input root". That way any sort of information we save can more easily be moved across different paths on the same machine, or across different machines.

nodirg:
WDYM by "saving things"? If you mean that cache keys should be relative, then I think this is achievable with this CL (it is orthogonal to this CL). For example, the input root could be a part of `UploadOptions` (and be optional, since not everyone need it).  If specified, then the cache keys would be relative to that root.

nodirg:
@rubensf PTAL

rubensf (approved):
I think I'm just being too hopeful here - I like the idea that we could save metadata regarding the path to disk so we could share it across different similar checkouts of the same codebase. But that is not really realistic :)

nodirg:
> I think I'm just being too hopeful here - I like the idea that we could save metadata regarding the path to disk so we could share it across different similar checkouts of the same codebase. But that is not really realistic :)

I also have hope that this is realistic.

It is orthogonal to this CL though. It is still possible to provide additional `UploadOptions.InputRoot`, make the absolute paths relative to that root and persist with relative roots. This CL doesn't preclude that.

PR #348, opened 2021-05-25
make (*Chunker).Reset() returns error

atetubou:
This is to confirm Reset doesn't return error in
https://crbug.com/1189349#c33

### DIFF
--- a/go/pkg/chunker/chunker.go
+++ b/go/pkg/chunker/chunker.go
@@ -2,14 +2,15 @@
 package chunker
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 
+	"github.com/klauspost/compress/zstd"
+	"github.com/pkg/errors"
+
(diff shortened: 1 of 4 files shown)


go/pkg/chunker/chunker.go
```diff
@@ -108,19 +109,24 @@ func (c *Chunker) ChunkSize() int {
 // Reset the Chunker state to when it was newly constructed.
 // Useful for upload retries.
 // TODO(olaola): implement Seek(offset) when we have resumable uploads.
-func (c *Chunker) Reset() {
+func (c *Chunker) Reset() error {
 	if c.r != nil {
 		// We're ignoring the error here, as not to change the fn signature.
```
rubensf, line 114: Remove this comment

rubensf (approved):

go/pkg/chunker/chunker.go
atetubou: removed

rubensf (approved):

PR #352, opened 2021-05-27
[cas] Clean PathSpec.Path

nodirg:
https://github.com/bazelbuild/remote-apis-sdks/pull/345
removed filepath.Abs call, but it has a side effect that we no longer
clean the path. Clean it explicitly.


rubensf (approved):
If we're going to copy the `PathSpec`, maybe just receive a concrete object instead of a pointer?

nodirg:
that's exactly what the first commit did :), but later I realized that I do need the original pointer. You will see it in another CL.

PR #353, opened 2021-05-27
[cas] Fix joinFilePathsFast

nodirg:
Fix joinFilePathsFast where the first path is the root.
We don't expect it in production, but it happens in tests.


rubensf (approved):

nodirg:
I did it again: merged into a wrong branch. Second attempt: https://github.com/bazelbuild/remote-apis-sdks/pull/353

PR #357, opened 2021-06-07
Fix lint

nodirg:
Fix lint left in
https://github.com/bazelbuild/remote-apis-sdks/pull/340

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -147,7 +147,7 @@ func (r *UploadResult) Digest(ps *PathSpec) (digest.Digest, error) {
 	// TODO(nodir): cache this syscall too.
 	info, err := os.Lstat(ps.Path)
 	if err != nil {
-		return digest.Digest{}, err
+		return digest.Digest{}, errors.WithStack(err)
 	}
 
 	key := makeFSCacheKey(ps.Path, info.Mode().IsRegular(), ps.Exclude)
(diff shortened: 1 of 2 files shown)


go/pkg/cas/upload.go
```diff
@@ -139,15 +139,15 @@ func (r *UploadResult) Digest(ps *PathSpec) (*digest.Digest, error) {
...
+		return nil, errors.WithStack(err)
 	}
 
 	key := makeFSCacheKey(ps.Path, info.Mode(), ps.Exclude)
 	switch val, err, loaded := r.u.fsCache.Load(key); {
 	case !loaded:
 		return nil, nil
 	case err != nil:
-		return nil, err
+		return nil, errors.WithStack(err)
```
rubensf, line 150: I think errors.Wrap will appease the lint?

nodirg: this is unfortunate. To appease the linter, I am going to have to wrap the error with info that the error already contains. 

Further, this lint error is incorrect. This error comes from this file. What do you suggest to do?

atetubou: I made https://github.com/bazelbuild/remote-apis-sdks/pull/358 for this.
This linter has some false positives, but I think it's still useful to prevent error where we don't come from.

nodirg: The linter is happy now.

nodirg:
PTAL

rubensf (approved):

PR #359, opened 2021-06-09
[cas] Refactor code

nodirg:
This change is a pure refactoring, in preparation for the commit
that adds support for Allowlist.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -142,7 +142,7 @@ func (r *UploadResult) Digest(ps *PathSpec) (*digest.Digest, error) {
 		return nil, err
 	}
 
-	key := makeFSCacheKey(ps.Path, info.Mode(), ps.Exclude)
+	key := makeFSCacheKey(ps.Path, info.Mode().IsRegular(), ps.Exclude)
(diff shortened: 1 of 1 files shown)


atetubou:
Why do we need to support allowlist in addition to inputs given to upload functions?

nodirg:
@atetubou we need to be able to upload a directory with *some*, but not all of its subpaths. Current API let's you either upload a directory entirely, or its individual pieces, but then you don't have the digest for the root.

go/pkg/cas/upload.go
```diff
@@ -943,7 +932,7 @@ func uploadItemFromBlob(title string, blob []byte) *uploadItem {
 const pathSep = string(filepath.Separator)
 
 // joinFilePathsFast is a faster version of filepath.Join because it does not
-// call filepath.Clean.
+// call filepath.Clean. Assumes arguments are clean.
```
rubensf, line 935:
```suggestion
// call filepath.Clean. Assumes arguments are clean according to filepath.Clean specs.
```

wdyt?

rubensf (approved):

go/pkg/cas/upload.go
nodirg: was incorporated.

PR #360, opened 2021-06-09
[cas] Add PathSpec.Allowlist

nodirg:
In the Isolate use case, we need to upload a directory partially.
Currently cas package does not let doing it efficiently.

It is possible to upload individual files/directories, but then the
caller has to construct the partial merkle tree themselves. For example,
to upload "/foo" with only "/foo/bar" and "/foo/baz", but not
"/foo/qux", the client has to upload bar and baz, and then calculate the
digest of /foo manually.

It is possible to exclude some paths using UploadOptions.Prelude,
but this filtering is at the UploadOptions level, while we need it at
the PathSpec level. Placing a callback in the PathSpec level makes
caching of intermediate directories difficult. In Go, it is not possible
to tell if two function pointers represent the same function with same
captured input.

Introduce PathSpec.Allowlist which is a list of paths, relative to
PathSpec.Path, that must be uploaded. The client uploads them the usual
way, and then constructs a partial Merkle tree to upload the root, as
well as the intermediate directories between the root and the
allowlisted paths.

The allowlist does not participate in the cache key. Doing so is
possible, but would require hashing the allowlist for each PathSpec and
might be an overkill *for now*. Also it complicates the documentation.
Instead, document that if a given PathSpec pointer was passed to
Client.Upload() earlier, then the same pointer is guaranteed to work in
UploadResult.Digest(). This is in addition to structural cache key which
is used to retrieve digests of files that were uploaded implicitly.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -48,6 +48,17 @@ type UploadInput struct {
 	// Must be absolute.
 	Path string
 
+	// Allowlist is a filter for files/directories under Path.
+	// If a file is not a present in Allowlist and does not reside in a directory
+	// present in the Allowlist, then the file is ignored.
+	// This is equivalent to deleting all not-matched files/dirs before
+	// uploading.
+	//
+	// Each path in the Allowlist must be relative to UploadInput.Path.
+	//
+	// Must be empty if Path points to a regular file.
+	Allowlist []string
+
 	// Exclude is a file/dir filter. If Exclude is not nil and the
 	// absolute path of a file/dir match this regexp, then the file/dir is skipped.
 	// Forward-slash-separated paths are matched aginst the regexp: PathExclude
@@ -57,11 +68,24 @@ type UploadInput struct {
 	// See ErrSkip comments for more details on semantics regarding excluding symlinks .
 	Exclude *regexp.Regexp
 
-	cleanPath string
+	cleanPath      string
+	cleanAllowlist []string
+
 	// pathInfo is result of Lstat(UploadInput.Path)
 	pathInfo os.FileInfo
 
-	digest              *repb.Digest
+	// tree maps from a file/dir path to its digest and a directory node.
+	// The path is relative to UploadInput.Path.
+	//
+	// Once digests are computed successfully, guaranteed to have key ".".
+	// If allowlist is not empty, then also has a key for each clean allowlisted
+	// path, as well as each intermediate directory between the root and an
+	// allowlisted dir.
+	//
+	// The main purpose of this field is an UploadInput-local cache that couldn't
+	// be placed in uploader.fsCache because of UploadInput-specific parameters
+	// that are hard to incorporate into the cache key, namedly the allowlist.
+	tree                map[string]*digested
 	digestsComputed     chan struct{}
 	digestsComputedInit sync.Once
 	u                   *uploader
@@ -81,8 +105,13 @@ func (in *UploadInput) Digest(relPath string) (digest.Digest, error) {
 	if in.cleanPath == "" {
 		return digest.Digest{}, errors.Errorf("Digest called too soon")
 	}
-	if relPath == "." {
-		return digest.NewFromProtoUnvalidated(in.digest), nil
+
+	relPath = filepath.Clean(relPath)
+
+	// Check if this is the root or one of the intermediate nodes in the partial
+	// Merkle tee.
+	if dig, ok := in.tree[relPath]; ok {
+		return digest.NewFromProtoUnvalidated(dig.digest), nil
 	}
 
 	absPath := filepath.Join(in.cleanPath, relPath)
@@ -118,6 +147,102 @@ func (in *UploadInput) ensureDigestsComputedInited() chan struct{} {
 // DigestsComputed() is always safe to call.
 func (in *UploadInput) DigestsComputed() <-chan struct{} {
 	return in.ensureDigestsComputedInited()
+}
+
+var oneDot = []string{"."}
+
+// init initializes internal fields.
+func (in *UploadInput) init(u *uploader) error {
+	in.u = u
+
+	if !filepath.IsAbs(in.Path) {
+		return errors.Errorf("%q is not absolute", in.Path)
+	}
+	in.cleanPath = filepath.Clean(in.Path)
+
+	// Do not use os.Stat() here. We want to know if it is a symlink.
+	var err error
+	if in.pathInfo, err = os.Lstat(in.cleanPath); err != nil {
+		return errors.WithStack(err)
+	}
+
+	// Process the allowlist.
+	in.tree = make(map[string]*digested, 1+len(in.Allowlist))
+	switch {
+	case len(in.Allowlist) == 0:
+		in.cleanAllowlist = oneDot
+
+	case in.pathInfo.Mode().IsRegular():
+		return errors.Errorf("the Allowlist is not supported for regular files")
+
+	default:
+		in.cleanAllowlist = make([]string, len(in.Allowlist))
+		for i, subPath := range in.Allowlist {
+			if filepath.IsAbs(subPath) {
+				return errors.Errorf("the allowlisted path %q is not relative", subPath)
+			}
+
+			cleanSubPath := filepath.Clean(subPath)
+			if cleanSubPath == ".." || strings.HasPrefix(cleanSubPath, parentDirPrefix) {
+				return errors.Errorf("the allowlisted path %q is not contained by %q", subPath, in.Path)
+			}
+			in.cleanAllowlist[i] = cleanSubPath
+		}
+	}
+	return nil
+}
+
+// partialMerkleTree ensures that for each node in in.tree, not included by any
+// other node, all its ancestors are also present in the tree. For example, if
+// the tree contains only "foo/bar" and "foo/baz", then partialMerkleTree adds
+// "foo" and ".". The latter is the root.
+//
+// All tree keys must be clean relative paths.
+// Returns prepared *uploadItems that represent the ancestors that were added to
+// the tree.
+func (in *UploadInput) partialMerkleTree() (added []*uploadItem) {
+	// Establish parent->child edges.
+	children := map[string]map[string]struct{}{}
+	for relPath := range in.tree {
+		for relPath != "." {
+			parent := dirNameRelFast(relPath)
+			if childSet, ok := children[parent]; ok {
+				childSet[relPath] = struct{}{}
+			} else {
+				children[parent] = map[string]struct{}{relPath: {}}
+			}
+			relPath = parent
+		}
+	}
+
+	// Add the missing ancestors by traversing in post-order.
+	var dfs func(relPath string) proto.Message
+	dfs = func(relPath string) proto.Message {
+		if dig, ok := in.tree[relPath]; ok {
+			return dig.dirEntry
+		}
+
+		dir := &repb.Directory{}
+		for child := range children[relPath] {
+			addDirEntry(dir, dfs(child))
+		}
+
+		// Prepare an uploadItem.
+		absPath := joinFilePathsFast(in.cleanPath, relPath)
+		item := uploadItemFromDirMsg(absPath, dir) // normalizes the dir
+		added = append(added, item)
+
+		// Compute a directory entry for the parent.
+		node := &repb.DirectoryNode{
+			Name:   filepath.Base(absPath),
+			Digest: item.Digest,
+		}
+
+		in.tree[relPath] = &digested{dirEntry: node, digest: item.Digest}
+		return node
+	}
+	dfs(".")
+	return added
 }
 
 // TransferStats is upload/download statistics.
@@ -308,24 +433,61 @@ func (u *uploader) startProcessing(ctx context.Context, in *UploadInput) error {
 	if !filepath.IsAbs(in.Path) {
 		return errors.Errorf("%q is not absolute", in.Path)
 	}
-	in.u = u
-	in.cleanPath = filepath.Clean(in.Path)
+
+	if err := in.init(u); err != nil {
+		return errors.WithStack(err)
+	}
 
 	// Schedule a file system walk.
 	u.wgFS.Add(1)
 	u.eg.Go(func() error {
 		defer u.wgFS.Done()
-		// Do not use os.Stat() here. We want to know if it is a symlink.
-		var err error
-		if in.pathInfo, err = os.Lstat(in.cleanPath); err != nil {
+
+		// Concurrently visit each allowlisted path, and use the results to
+		// construct a partial Merkle tree. Note that we are not visiting
+		// the entire in.cleanPath, which may be much larger than the union of the
+		// allowlisted paths.
+		localEg, ctx := errgroup.WithContext(ctx)
+		var treeMu sync.Mutex
+		for _, relPath := range in.cleanAllowlist {
+			relPath := relPath
+			// Schedule a file system walk.
+			localEg.Go(func() error {
+				absPath := in.cleanPath
+				info := in.pathInfo
+				if relPath != "." {
+					absPath = joinFilePathsFast(in.cleanPath, relPath)
+					var err error
+					// TODO(nodir): cache this syscall too.
+					if info, err = os.Lstat(absPath); err != nil {
+						return errors.WithStack(err)
+					}
+				}
+
+				switch dig, err := u.visitPath(ctx, absPath, info, in.Exclude); {
+				case err != nil:
+					return errors.Wrapf(err, "%q", absPath)
+				case dig != nil:
+					treeMu.Lock()
+					in.tree[relPath] = dig
+					treeMu.Unlock()
+				}
+				return nil
+			})
+		}
+		if err := localEg.Wait(); err != nil {
 			return errors.WithStack(err)
 		}
 
-		dig, err := u.visitPath(ctx, in.cleanPath, in.pathInfo, in.Exclude)
-		if err != nil {
-			return errors.Wrapf(err, "%q", in.Path)
+		// At this point, all allowlisted paths are digest'ed, and we only need to
+		// compute a partial Merkle tree and upload the implied ancestors.
+		for _, item := range in.partialMerkleTree() {
+			if err := u.scheduleCheck(ctx, item); err != nil {
+				return err
+			}
 		}
-		in.digest = dig.digest
+
+		// The entire tree is digested. Notify the caller.
 		close(in.ensureDigestsComputedInited())
 		return nil
 	})
@@ -973,16 +1135,33 @@ func uploadItemFromBlob(title string, blob []byte) *uploadItem {
 	return item
 }
 
-const pathSep = string(filepath.Separator)
+const (
+	pathSep         = string(filepath.Separator)
+	parentDirPrefix = ".." + pathSep
+)
 
 // joinFilePathsFast is a faster version of filepath.Join because it does not
 // call filepath.Clean. Assumes arguments are clean according to filepath.Clean specs.
 func joinFilePathsFast(a, b string) string {
+	if b == "." {
+		return a
+	}
 	if strings.HasSuffix(a, pathSep) {
 		// May happen if a is the root.
 		return a + b
 	}
 	return a + pathSep + b
+}
+
+// dirNameRelFast is a faster version of filepath.Dir because it does not call
+// filepath.Clean. Assumes the argument is clean and relative.
+// Does not work for absolute paths.
+func dirNameRelFast(relPath string) string {
+	i := strings.LastIndex(relPath, pathSep)
+	if i < 0 {
+		return "."
+	}
+	return relPath[:i]
 }
 
 func marshalledFieldSize(size int64) int64 {
--- a/go/pkg/cas/upload_test.go
+++ b/go/pkg/cas/upload_test.go
@@ -39,12 +39,30 @@ func TestFS(t *testing.T) {
 	putFile(t, filepath.Join(tmpDir, "root", "subdir", "c"), "c")
 	cItem := uploadItemFromBlob(filepath.Join(tmpDir, "root", "subdir", "c"), []byte("c"))
 
+	putFile(t, filepath.Join(tmpDir, "root", "subdir", "d"), "d")
+	dItem := uploadItemFromBlob(filepath.Join(tmpDir, "root", "subdir", "d"), []byte("d"))
+
 	subdirItem := uploadItemFromDirMsg(filepath.Join(tmpDir, "root", "subdir"), &repb.Directory{
-		Files: []*repb.FileNode{{
-			Name:   "c",
-			Digest: cItem.Digest,
-		}},
+		Files: []*repb.FileNode{
+			{
+				Name:   "c",
+				Digest: cItem.Digest,
+			},
+			{
+				Name:   "d",
+				Digest: dItem.Digest,
+			},
+		},
 	})
+	subdirWithoutDItem := uploadItemFromDirMsg(filepath.Join(tmpDir, "root", "subdir"), &repb.Directory{
+		Files: []*repb.FileNode{
+			{
+				Name:   "c",
+				Digest: cItem.Digest,
+			},
+		},
+	})
+
 	rootItem := uploadItemFromDirMsg(filepath.Join(tmpDir, "root"), &repb.Directory{
 		Files: []*repb.FileNode{
 			{Name: "a", Digest: aItem.Digest},
@@ -66,6 +84,15 @@ func TestFS(t *testing.T) {
 		Files: []*repb.FileNode{
 			{Name: "a", Digest: aItem.Digest},
 			{Name: "b", Digest: bItem.Digest},
+		},
+	})
+	rootWithoutDItem := uploadItemFromDirMsg(filepath.Join(tmpDir, "root"), &repb.Directory{
+		Files: []*repb.FileNode{
+			{Name: "a", Digest: aItem.Digest},
+			{Name: "b", Digest: bItem.Digest},
+		},
+		Directories: []*repb.DirectoryNode{
+			{Name: "subdir", Digest: subdirWithoutDItem.Digest},
 		},
 	})
 
@@ -129,7 +156,7 @@ func TestFS(t *testing.T) {
 			desc:                "root",
 			inputs:              []*UploadInput{{Path: filepath.Join(tmpDir, "root")}},
 			wantDigests:         digSlice(rootItem),
-			wantScheduledChecks: []*uploadItem{rootItem, aItem, bItem, subdirItem, cItem},
+			wantScheduledChecks: []*uploadItem{rootItem, aItem, bItem, subdirItem, cItem, dItem},
 		},
 		{
 			desc:        "root-without-a-using-callback",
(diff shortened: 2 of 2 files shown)


rubensf:
FYI, remember to change the base :)

go/pkg/cas/upload.go
```diff
@@ -266,25 +332,87 @@ type uploader struct {
...
 
-		_, err = u.visitPath(ctx, ps.Path, info, ps.Exclude)
-		return errors.Wrapf(err, "%q", ps.Path)
+		// Allowlist was specified. This upload is partial.
+		// Concurrently visit the allowlisted paths, and use the results to
+		// construct a partial Merkle tree.
+		// Note that we are not visiting ps.Path, which may be much larger than
+		// the union of the allowlisted paths.
+		localEg, ctx := errgroup.WithContext(ctx)
+		var mu sync.Mutex
```
rubensf, line 372: `treeMu`

go/pkg/cas/upload.go
```diff
@@ -266,25 +332,87 @@ type uploader struct {
...
 	// Schedule a file system walk.
 	u.wgFS.Add(1)
 	u.eg.Go(func() error {
 		defer u.wgFS.Done()
-		// Do not use os.Stat() here. We want to know if it is a symlink.
-		info, err := os.Lstat(ps.Path)
-		if err != nil {
-			return errors.WithStack(err)
+
+		if len(ps.Allowlist) == 0 {
```
rubensf, line 354: I wonder if instead you could tweak this case to be treated as if an empty allowlist is treated as a `[]Allowlist{"."}`? It could save quite a few lines of code

go/pkg/cas/upload.go
```diff
@@ -266,25 +332,87 @@ type uploader struct {
...
+				tree[subPath] = dig
+				mu.Unlock()
+				return nil
+			})
+		}
+		if err := localEg.Wait(); err != nil {
+			return err
+		}
+
+		// At this point, all allowlisted paths are digest`ed, and we only need to
```
rubensf, line 401: digest`ed -> digest'ed 

?

go/pkg/cas/upload.go
```diff
@@ -965,6 +1106,67 @@ func marshalledRequestSize(d *repb.Digest) int64 {
...
+	digested *digested
+}
+
+// missingDirs returns missing directories on the node path from the root to
+// each of the path specified in the tree.
+// For example, if the tree contains only "foo/bar" and "foo/baz", then
+// missingDirs returns "foo" and the root ".".
+//
+// All tree keys must be clean relative paths.
+func missingDirs(tree map[string]*digested, absRoot string) []missingDir {
```
rubensf, line 1121: should we call this `partialMerkleTree`? I was a bit confused as I expected `missingDirs` to somehow go over/return the directories _not_ listed in `tree` (eg `foo` has `bar/` and `baz/`, `tree` include `bar/`, this returns `baz/`).

go/pkg/cas/upload.go
```diff
@@ -935,13 +1063,26 @@ const pathSep = string(filepath.Separator)
...
 	if strings.HasSuffix(a, pathSep) {
 		// May happen if a is the root.
 		return a + b
 	}
 	return a + pathSep + b
 }
 
+// dirNameRelFast is a faster version of filepath.Dir because it does not call
+// filepath.Clean. Assumes the argument is clean and relative.
+func dirNameRelFast(relPath string) string {
```
rubensf, line 1078: is this technically just `dirNameFast`? IIUC there's no requirement for path to be relative to anything.

rubensf (commented):
Initial set of comments - I'm still digging through the code :)

go/pkg/cas/upload.go
```diff
@@ -57,6 +69,48 @@ type PathSpec struct {
...
+func (ps *PathSpec) deepCopy() *PathSpec {
+	ret := *ps
+	if len(ps.Allowlist) > 0 {
+		ret.Allowlist = make([]string, len(ps.Allowlist))
+		copy(ret.Allowlist, ps.Allowlist)
+	}
+	return &ret
+}
+
+// validateAndNormalize validates and normalizes ps.
```
ramymedhat, line 81: Can you add more info here that cannot be inferred from the function name?

go/pkg/cas/upload.go
```diff
@@ -123,11 +177,13 @@ type UploadResult struct {
...
 //
-// To retrieve a digest of a regular file, only ps.Path is checked - other
-// fields are ignored.
-//
-// To retrieve a digest of a directory or a symlink, ps.Exclude must match one
+// If ps is a PathSpec pointer that was passed to Upload() earlier, then Digest
+// returns the digest computed for that PathSpec.
+// Otherwise PathSpec is matched structurally:
+//  - for regular files, only ps.Path is checked - other fields are ignored.
+//  - for directories and symlinks, ps.Exclude must match one
```
ramymedhat, line 184: Did you consider some sort of Context struct where the request (PathSpec) and result can live? That way you avoid documenting constraints around using the same pointer because its implicit. For instance:
```
type UploadContext struct {
  ps *PathSpec
  ur *UploadResult
}

func NewUploadContext(ps *PathSpec) *UploadContext {
}

func (c *UploadContext) RootDigest() *digest.Digest {
}

func (c *UploadContext) SubpathDigest(path string) *digest.Digest {
}
```

go/pkg/cas/upload_test.go
```diff
@@ -145,6 +145,12 @@ func TestFS(t *testing.T) {
 			},
 			wantScheduledChecks: []*uploadItem{rootWithoutAItem, bItem, subdirItem, cItem},
 		},
+		{
+			desc:                "root-without-a-using-allowlist",
+			paths:               []*PathSpec{{Path: filepath.Join(tmpDir, "root"), Allowlist: []string{"b", "subdir"}}},
```
rubensf, line 150: Add a check where only `subdir/c` is included? 

May also be nice to have a slightly more complicated one like `dir/f2`, `dir/sub/f2` are added whereas in there are more files in those subfolders.

go/pkg/cas/upload_test.go
```diff
@@ -453,6 +470,185 @@ func TestStreaming(t *testing.T) {
 	}
 }
```
rubensf, line 472: I think there's some value on adding a test that checks the same pathspec with two different `Allowlist` don't interfere with each other in a same `Upload` call.

go/pkg/cas/upload_test.go
```diff
@@ -453,6 +470,185 @@ func TestStreaming(t *testing.T) {
...
+
+	testCases := []struct {
+		desc     string
+		tree     map[string]*digested
+		wantDirs []missingDir
+	}{
+		{
+			desc: "works",
+			tree: map[string]*digested{
+				"foo/bar": {
```
rubensf, line 511: I think it's worth it to have a slightly more complex scenario here too.

go/pkg/cas/upload.go
```diff
@@ -137,22 +192,30 @@ func (r *UploadResult) Digest(ps *PathSpec) (*digest.Digest, error) {
...
-		return nil, err
-	default:
-		dig := digest.NewFromProtoUnvalidated(val.(*digested).digest)
-		return &dig, nil
-	}
+	dig, err, _ := loadCache(makeFSCacheKey(ps.Path, info.Mode().IsRegular(), ps.Exclude))
+	return dig, err
 }
 
 // Upload uploads all files/directories specified by pathC.
```
rubensf, line 221: Regarding `Upload` - do we want to return a channel instead of an `*UploadResult`? Otherwise, I think we lose the benefits of having a pathC channel input. The user likely wants to get the results of its uploads (including dirs using allowlist) before it is done "uploading all things".

go/pkg/cas/upload.go
```diff
@@ -57,6 +69,48 @@ type PathSpec struct {
 	Exclude *regexp.Regexp
 }
 
+func (ps *PathSpec) deepCopy() *PathSpec {
```
atetubou, line 72: no need to use https://golang.org/pkg/regexp/#Regexp.Copy for Exclude?

go/pkg/cas/upload.go
```diff
@@ -265,31 +332,93 @@ type uploader struct {
...
 func (u *uploader) startProcessing(ctx context.Context, ps *PathSpec) error {
-	if !filepath.IsAbs(ps.Path) {
-		return errors.Errorf("%q is not absolute", ps.Path)
+	origPsPtr := ps
+
+	// Do not use os.Stat() here. We want to know if it is a symlink.
+	// TODO(nodir): cache this syscall, perhaps using filemetadat package.
+	info, err := os.Lstat(ps.Path)
+	if err != nil {
+		return errors.WithStack(err)
```
atetubou, line 341: do you see the linter error after rebase?

go/pkg/cas/upload.go
```diff
@@ -970,3 +1105,79 @@ func marshalledRequestSize(d *repb.Digest) int64 {
...
+// missingDirs returns "foo" and the root ".".
+//
+// All tree keys must be clean relative paths.
+func missingDirs(tree map[string]*digested, absRoot string) []missingDir {
+	// Establish parent->child edges.
+	children := map[string]map[string]struct{}{}
+	for relPath := range tree {
+		for relPath != "." {
+			parent := dirNameRelFast(relPath)
+			if childSet := children[parent]; childSet != nil {
```
atetubou, line 1127:
```suggestion
			if childSet, ok := children[parent]; ok {
```

go/pkg/cas/upload.go
```diff
@@ -970,3 +1105,79 @@ func marshalledRequestSize(d *repb.Digest) int64 {
...
+			}
+			relPath = parent
+		}
+	}
+
+	// Add the missing directories by traversing in post-order.
+	var missing []missingDir
+	var dfs func(relPath string) proto.Message
+	dfs = func(relPath string) proto.Message {
+		if dig := tree[relPath]; dig != nil {
```
atetubou, line 1140: Better to check like
```suggestion
		if dig, ok := tree[relPath]; ok {
```

Or is there a case where `tree[relPath] == nil` becomes true?

go/pkg/cas/upload_test.go
```diff
@@ -453,6 +470,185 @@ func TestStreaming(t *testing.T) {
...
+	for _, tc := range testCases {
+		gotDirs := missingDirs(tc.tree, "/")
+		sort.Slice(gotDirs, func(i, j int) bool {
+			return gotDirs[i].relPath < gotDirs[j].relPath
+		})
+
+		compareDirs := func(x, y missingDir) bool {
+			return proto.Equal(x.digested.digest, y.digested.digest) &&
+				proto.Equal(x.digested.dirEntry, y.digested.dirEntry) &&
+				compareUploadItems(x.item, y.item)
```
atetubou, line 562: Is it possible to pass only this function to `cmp.Comparer` with `protocmp.Transform` to `cmp.Diff`?

go/pkg/cas/upload_test.go
```diff
@@ -453,6 +470,185 @@ func TestStreaming(t *testing.T) {
...
+		}
+
+		if diff := cmp.Diff(tc.wantDirs, gotDirs, cmp.Comparer(compareDirs)); diff != "" {
+			t.Errorf("unexpected digests (-want +got):\n%s", diff)
+		}
+	}
+}
+
+func TestValidateAndNormalize(t *testing.T) {
```
atetubou, line 571: use `t.Parallel()`?

go/pkg/cas/upload_test.go
```diff
@@ -453,6 +470,185 @@ func TestStreaming(t *testing.T) {
 	}
 }
 
+func TestMissingDirs(t *testing.T) {
```
atetubou, line 473: use `t.Parallel()`?

go/pkg/cas/upload.go
```diff
@@ -58,6 +69,48 @@ type PathSpec struct {
...
+			return errors.Errorf("the allowlisted path %q is not relative", subPath)
+		}
+
+		cleanPath := filepath.Clean(subPath)
+		if cleanPath == "." {
+			ps.Allowlist = nil
+			break
+		}
+
+		for _, comp := range strings.Split(cleanPath, pathSep) {
```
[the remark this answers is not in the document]
nodirg: Done

go/pkg/cas/upload.go
```diff
@@ -266,25 +332,87 @@ type uploader struct {
 
 // startProcessing adds the item to the appropriate stage depending on its type.
 func (u *uploader) startProcessing(ctx context.Context, ps *PathSpec) error {
-	if !filepath.IsAbs(ps.Path) {
-		return errors.Errorf("%q is not absolute", ps.Path)
+	origPsPtr := ps
+
+	// Do not use os.Stat() here. We want to know if it is a symlink.
+	// TODO(nodir): cache this syscall, perhaps using filemetadat package.
```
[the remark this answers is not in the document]
nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: This is exactly what I did in an earlier version :), but then problem was that then it would mean I'd have to have the "if root, then persist with ps pointer cache key" logic in two places: 
1) when iterating over return value of `missingDirs()` like now
2) but also, in the ps.Allowlist loop.

however, thinking about this a bit more, we have it in two places anyway. So I am happy to change this if you prefer.

go/pkg/cas/upload.go
nodirg: Done

go/pkg/cas/upload.go
nodirg: In an earlier version, this func was called `partialMerkleTree` but then I realized that it doesn't quite build merkle tree. Rather, it builds those missing ancestors.

I agree with your point and renamed it to `missingAncestors`. I think the reader should no longer expect children to be returned. WDYT?

go/pkg/cas/upload.go
nodirg: it does not work for "/" and "C:/"

go/pkg/cas/upload.go
nodirg: I wasn't sure about this one. My true intention was to just document that mode must match ps.Path, but Go requires the comment to start with the symbol name. At the same time, I'd like to avoid documenting the implementation details of this func. I.e. I don't we need to document that it returns an error if ps.Path is relative

Ideas?

go/pkg/cas/upload.go
nodirg: I did not consider this. If we want to be more explicit and make APIs say what English says, then perhaps we could have

```go
// checks only pointer
func (r *UploadResult) Digest(*PathSpec) (digest.Digest, error)
func (r *UploadResult) DigestRegular(fileName string) (digest.Digest, error)
func (r *UploadResult) DigestDir(path string, exclude) (digest.Digest, error)
```

`DigestDir` isn't future proof though because if we add another PathSpec field that controls file traversal and must be propagated here, then we'd have to make a breaking change in DigestDir. The reason I proposed PathSpec struct (and called it "spec") is exactly this scenario: we can add new fields to PathSpec and they become the API of `UploadResult.Digest()`.

`DigestRegular` is future-proof though, since there is no file traversal. I can add that. WDYT?

nodirg: I did not fully understand the `UploadContext` idea or how it would in this case. What part of the contract-in-English does it alleviate? We still have to explain that Allowlist is ignored, and that Exclude must be same, and that none of them matter for regular files.

go/pkg/cas/upload.go
```diff
@@ -935,13 +1063,26 @@ const pathSep = string(filepath.Separator)
...
 		// May happen if a is the root.
 		return a + b
 	}
 	return a + pathSep + b
 }
 
+// dirNameRelFast is a faster version of filepath.Dir because it does not call
+// filepath.Clean. Assumes the argument is clean and relative.
+func dirNameRelFast(relPath string) string {
+	i := strings.LastIndexByte(relPath, os.PathSeparator)
```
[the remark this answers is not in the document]
nodirg: Looks like `strings.LastIndex` is smart enough to switch to `LastIndexByte` if the len(substring)==1, so done.

go/pkg/cas/upload.go
nodirg: I think we discussed this before. Earlier I suggested `Digest` callback in `UploadInput` for exactly this reason.
I think we agreed that we can add it on demand.

I think a channel return value is a bit hard to work with:
- even if you don't care about having digests in a streaming fashion, you are forced to setup a separate goroutine that receives from the channel. IMO in simple cases the code should be simple too.
- suppose PathSpecs are generated by different goroutines and send to pathC. Even if we channel returned by Upload(), then can't easily know the digest of the root. The programmer would have to somehow figure out which goroutine is interested in a particular digest received from the channel.

That's why I initially proposed `UploadItem.Digest` callback. It covers both concerns:
- If you don't care, then you just don't initialize it. Life is simple.
- Each goroutine may pass their callback.

So I agree that currently we are not benefiting from the fact that pathC is a channel, but I think the way to solve it is to add callback to PathSpec.

Thoughts?

go/pkg/cas/upload.go
nodirg: good point. Renamed to `copy` and add a comment.

go/pkg/cas/upload.go
nodirg: no, there is no such case.

go/pkg/cas/upload.go
nodirg: rebasing will screw up comments, so going to rebase later.

go/pkg/cas/upload_test.go
nodirg: I can't figure out how to do that. How would you do that?

go/pkg/cas/upload_test.go
nodirg: Done

go/pkg/cas/upload_test.go
nodirg: Done.

go/pkg/cas/upload.go
nodirg: done

go/pkg/cas/upload.go
ramymedhat: The purpose of `UploadContext` is to avoid documenting that you need to pass the `PathSpec` with the exact same pointer to get the correct digest, which I feel is a bit awkward. Specially that PathSpec doesn't have any state, its a spec, and the requirement that it be the same pointer is an implementation detail of caching.  If the UploadResult is really `PathSpec` specific then imo we should tie these together so that either the `UploadResult` is retrievable from the `PathSpec` object or both are encapsulated in something like an `UploadContext`.

nodirg:
Will work on comments in `upload_test.go` after we are happy with `upload.go`

go/pkg/cas/upload.go
nodirg: OK, how about this

1) Remove UploadResult.Digest. Keep UploadResult though, as we might add something else later.
2) Rename PathSpec back to UploadItem. It was called PathSpec to be used with UploadResult.Digest(), but looks like we aren't happy with it
3) Add `UploadItem.Digest(relPath string) digest.Digest` which returns the digest for the path relative to `UploadItem.Path`. For example, `myItem.Digest(".")` returns the path for the root. It is pretty easy to use. One nice property is that it will work even for non-root ancestors for allowlisted-dirs.

This somewhat similar to the original proposal, where UploadItem.Digest is a callback, but this is simpler.

*For now*, Digest() will be guaranteed to work only after Upload() returns, but later can add `Done() <- chan struct{}` which is closed when Digest() is ready to be used.

WDYT?

nodirg: @rubensf

ramymedhat: Sgtm, I like point 3, that seems fairly simple to use.

rubensf: I'm not a big fan of a channel that indicates that data is available _somewhere else_. I don't think it's wrong but I'd hope it could be simpler than that. I _think_ it will look simpler to have a callback for results (if the user cares about them) - so instead of calling it `Digest()`, make it `OnComplete(ur UploadResults)` (as per other comment regarding channel inputs).

We could then add a `UploadResult.Digest(relPath string) digest.Digest`, instead of at the UploadItem. 

wdyt?

go/pkg/cas/upload.go
rubensf: I _think_ the code readability is worth it the small overhead of creating a tree/waiting on a lock. I _think_ that we wouldn't need to check it in the Allowlist loop, since it'd automatically satisfy the `if dir.relPath == "."` condition on the `missingAncestors` function. Unless I'm missing some other condition (and you concur), I'd prefer the general treatment.

rubensf (commented):
Apologies for the slow review!

go/pkg/cas/upload.go
nodirg: > I'm not a big fan of a channel that indicates that data is available somewhere else. 

Isn't this the standard practice in Go? https://pkg.go.dev/context#Context.Done
A `chan struct{}` closed to indicate an event seems pretty standard: https://blog.golang.org/pipelines

> OnComplete(UploadResults)

is this UploadResults different from the one we have today? Or are you proposing to add a new struct? A func returning a channel seems simpler.

This, seems more complicated to use. The digests are guaranteed to be computed after Upload() succeeds. So if we ensured that it succeeded, then we can just call `Digest`. Compare:

```go
// simple
dig := in.Digest(".")

// complex
var dig digest.Digest
input.OnCompleted(func(r UploadResults) { 
  dig = r.Digest(".")
})
```

nodirg: also in general, once cannot `select` on a callback. How would a separate goroutine "block" on `OnComplete`?

also this would be more complicated to implement, since we'd need to store the callbacks, and then call them all, etc.

Anyway, let's continue in https://github.com/bazelbuild/remote-apis-sdks/pull/363
I didn't see your comment when I uploaded the new PR.

go/pkg/cas/upload.go
nodirg: see the other thread. 
Impl in https://github.com/bazelbuild/remote-apis-sdks/pull/363 seems pretty simple to me. Storing callbacks to call later would be more complicated.
Also a channel for "digests are ready to be read" seems easier to use (see the other thread).

go/pkg/cas/upload.go
rubensf: Apologies, I had a different idea in mind of what you meant with the `Done` function. #363 lgtm.

go/pkg/cas/upload.go
nodirg: added. Kind expected that "Assumes the argument is clean and relative." is enough

go/pkg/cas/upload.go
nodirg: done. The UploadInput refactoring incidentally made this much simpler.

go/pkg/cas/upload.go
nodirg: no longer applicable

go/pkg/cas/upload.go
nodirg: with the UploadInput refactoring, this func now mutates `tree`, and the caller actually receives the merkle tree, so done.

go/pkg/cas/upload.go

go/pkg/cas/upload.go
nodirg: no longer applicable

go/pkg/cas/upload.go
nodirg: now that each `UploadInput` has `DigestsComputed`, the underlying problem is solved.

go/pkg/cas/upload_test.go
nodirg: no longer applicable

nodirg:
PTAL. I've rebased this PR on top of the `UploadInput` refactoring, and now it is much simpler.
Still working on tests though.

go/pkg/cas/upload.go
```diff
@@ -120,6 +149,102 @@ func (in *UploadInput) DigestsComputed() <-chan struct{} {
...
+			} else {
+				children[parent] = map[string]struct{}{relPath: {}}
+			}
+			relPath = parent
+		}
+	}
+
+	// Add the missing ancestors by traversing in post-order.
+	var dfs func(relPath string) proto.Message
+	dfs = func(relPath string) proto.Message {
```
rubensf, line 220: Maybe just return it, and add "dfs" to the comment? (post-order should be explicit enough, but dfs may be easier to ring bells to the reader). No strong feelings.
```suggestion
	return func(relPath string) proto.Message { ... } (".")
```

rubensf (commented):
waiting on tests, but otherwise LGTM so far

go/pkg/cas/upload.go
nodirg: this is a recursive func. Without `dfs`, it cannot call itself.
If it wasn't recursive, we wouldn't need the func to begin with.

nodirg: also, the return value of dfs() is not what partialMerkleTree returns. The latter returns `added`, which is captured in the func.

go/pkg/cas/upload_test.go
nodirg: Done

go/pkg/cas/upload_test.go
nodirg: Dne

go/pkg/cas/upload_test.go
nodirg: added a couple more

nodirg:
Addressed comments for tests.

go/pkg/cas/upload.go
```diff
@@ -120,6 +149,102 @@ func (in *UploadInput) DigestsComputed() <-chan struct{} {
...
+
+	default:
+		in.cleanAllowlist = make([]string, len(in.Allowlist))
+		for i, subPath := range in.Allowlist {
+			if filepath.IsAbs(subPath) {
+				return errors.Errorf("the allowlisted path %q is not relative", subPath)
+			}
+
+			cleanSubPath := filepath.Clean(subPath)
+			if cleanSubPath == ".." || strings.HasPrefix(cleanSubPath, parentDirPrefix) {
```
atetubou, line 186: Why not do like?
```suggestion
			if strings.HasPrefix(cleanSubPath, "..") {
```

go/pkg/cas/upload.go
```diff
@@ -120,6 +149,102 @@ func (in *UploadInput) DigestsComputed() <-chan struct{} {
...
+				return errors.Errorf("the allowlisted path %q contains `..`", subPath)
```
atetubou, line 187:
```suggestion
				return errors.Errorf("the allowlisted path %q starts from `..`", subPath)
```

go/pkg/cas/upload.go
```diff
@@ -120,6 +149,102 @@ func (in *UploadInput) DigestsComputed() <-chan struct{} {
...
+// other node, all its ancestors are also present in the tree. For example, if
+// the tree contains only "foo/bar" and "foo/baz", then partialMerkleTree adds
+// "foo" and ".". The latter is the root.
+//
+// All tree keys must be clean relative paths.
+// Returns prepared *uploadItems that represent the ancestors that were added to
+// the tree.
+func (in *UploadInput) partialMerkleTree() (added []*uploadItem) {
+	// Establish parent->child edges.
+	children := map[string]map[string]struct{}{}
```
atetubou, line 205: Is it possible to use slice?
```suggestion
	children := map[string][]string
```

go/pkg/cas/upload.go
nodirg: because `..a` is a valid name

go/pkg/cas/upload.go
nodirg: no. We currently don't prevent/validate that Allowlist doesn't have redundant paths, e.g. "foo" and "foo/bar". In this case, "foo/bar" is redundant. If we use a slice, then root would know about foo twice.

We could remove redundant paths and use a slice here, but removal impl would go back to the hashset, so the overall complexity would not decrease.

go/pkg/cas/upload.go
nodirg: subPath doesn't necessarily start with "..". Consider `foo/../..`. 

Clarified the error message.

go/pkg/cas/upload.go
atetubou: I see.

atetubou (approved):
LGTM, but I hope RBE folks also approve this.

ramymedhat (approved):

rubensf (approved):

PR #361, opened 2021-06-09
[cas] Fix joinFilePathsFast

nodirg:
Fix joinFilePathsFast where the first path is the root.
We don't expect it in production, but it happens in tests.


nodirg:
same as https://github.com/bazelbuild/remote-apis-sdks/pull/353 which I merged into a wrong branch

rubensf (approved):

PR #362, opened 2021-06-09
[cas] Simplify UploadResult.Digest() signature

nodirg:
Instead of returning a possibly nil digest, introduce a new error
ErrNoDigest and return that instead. It simplifies most callsites: now
they check only 2 cases instead of 3.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -106,13 +106,19 @@ type digested struct {
 	digest   *repb.Digest  // may be nil, e.g. for dangling symlinks
 }
 
-// ErrSkip when returned by UploadOptions.Prelude, means the file/dir must be
-// not be uploaded.
-//
-// Note that if UploadOptions.PreserveSymlinks is true and the ErrSkip is
-// returned for a symlink target, but not the symlink itself, then it may
-// result in a dangling symlink.
-var ErrSkip = errors.New("skip file")
+var (
+	// ErrSkip when returned by UploadOptions.Prelude, means the file/dir must be
+	// not be uploaded.
+	//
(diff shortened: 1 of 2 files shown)


go/pkg/cas/upload.go
```diff
@@ -121,6 +121,9 @@ type UploadResult struct {
 	u     *uploader
 }
 
+// ErrNoDigest indicates that the requested digest is uknown.
+var ErrNoDigest = errors.New("the requested digest is unknown")
```
rubensf, line 125: Move it up with ErrSkip?

rubensf (approved):

go/pkg/cas/upload.go
```diff
@@ -131,26 +134,25 @@ type UploadResult struct {
...
 	if err != nil {
-		return nil, err
+		return digest.Digest{}, err
 	}
 
 	key := makeFSCacheKey(ps.Path, info.Mode(), ps.Exclude)
 	switch val, err, loaded := r.u.fsCache.Load(key); {
 	case !loaded:
-		return nil, nil
+		return digest.Digest{}, ErrNoDigest
```
atetubou, line 151: better to wrap the error with key?

atetubou (approved):

go/pkg/cas/upload.go
nodirg: Could you elaborate on "skip"? 

The name "ErrSkip" is often used as a return value of a callback which tells the calling function to skip something during iteration, e.g. skip a directory and don't dive into it. Or skip a test or something.

What are we skipping here?

nodirg: Done

go/pkg/cas/upload.go
nodirg: Done

PR #363, opened 2021-06-16
[cas] Refactor Digest()

nodirg:
UploadResult.Digest(PathSpec) semantics turned out to be complicated.
Remove it, rename PathSpec back to UploadInput and add
UploadInput.Digest(relPath string). The latter is easy to use, e.g.
the digest for UploadInput.Path is just UploadInput.Digest(".").

Since UploadInputs are sent to the client over the channel, the caller
might want to produce them in different goroutines, where each goroutine
might want to read the computed digest back. Since those goroutines
don't necessarily know when Upload() call returns, add
UploadInput.DigestComputed() function that returns a channel which is
closed when the digest for the UploadInput is computed.
This means Upload() call can be long-lived and multiple goroutines can
upload files independently.

Add more tests.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -29,7 +29,7 @@ import (
 )
 
 // ErrFilteredSymlinkTarget is returned when a symlink's target was filtered out
-// via PathSpec.Exclude or ErrSkip, while the symlink itself wasn't.
+// via UploadInput.Exclude or ErrSkip, while the symlink itself wasn't.
 var ErrFilteredSymlinkTarget = errors.New("symlink's target was filtered out")
 
 // zstdEncoders is a pool of ZStd encoders.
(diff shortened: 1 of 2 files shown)


go/pkg/cas/upload.go
```diff
@@ -42,11 +42,14 @@ var zstdEncoders = sync.Pool{
...
 
-// PathSpec specifies a subset of the file system.
-type PathSpec struct {
+// UploadInput specifies a file or directory to upload.
+type UploadInput struct {
 	// Path to the file or a directory to upload.
 	// Must be absolute.
-	Path string
+	Path      string
+	cleanPath string
```
rubensf, line 50: Maybe group exported and unexported fields, for people reading the code who just want to figure out what options are available? no strong feelings.

rubensf (approved):

go/pkg/cas/upload.go

nodirg:
@ramymedhat PTAL

ramymedhat (approved):

PR #364, opened 2021-06-29
Add UploadInput.OSNativePattern

nodirg:
UploadInput.Exclude is OS-agnostic (good), but Chrome's isolate uses
OS-native patterns (legacy). Add an option to switch UploadInput.Exclude
to be OS-native.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -65,7 +65,7 @@ type UploadInput struct {
 	// in the subtree.
 	// See ErrSkip comments for more details on semantics regarding excluding symlinks .
 	//
-	// Forward-slash-separated paths are matched aginst the regexp by default, so
+	// Forward-slash-separated paths are matched against the regexp by default, so
 	// that the pattern does not have to be conditional on the OS.
 	// This behavior can be changed by setting OSNativePattern to true.
 	Exclude *regexp.Regexp
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -61,12 +61,19 @@ type UploadInput struct {
 
 	// Exclude is a file/dir filter. If Exclude is not nil and the
 	// absolute path of a file/dir match this regexp, then the file/dir is skipped.
-	// Forward-slash-separated paths are matched aginst the regexp: PathExclude
-	// does not have to be conditional on the OS.
 	// If the Path is a directory, then the filter is evaluated against each file
 	// in the subtree.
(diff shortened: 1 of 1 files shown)


nodirg:
This should help with https://logs.chromium.org/logs/infra/buildbucket/cr-buildbucket.appspot.com/8843109528375521808/+/u/go_test/stdout

nodirg:
apparently Rubens is OOO

go/pkg/cas/upload.go
```diff
@@ -61,13 +61,20 @@ type UploadInput struct {
 
 	// Exclude is a file/dir filter. If Exclude is not nil and the
 	// absolute path of a file/dir match this regexp, then the file/dir is skipped.
-	// Forward-slash-separated paths are matched aginst the regexp: PathExclude
-	// does not have to be conditional on the OS.
 	// If the Path is a directory, then the filter is evaluated against each file
 	// in the subtree.
 	// See ErrSkip comments for more details on semantics regarding excluding symlinks .
+	//
+	// Forward-slash-separated paths are matched aginst the regexp by default, so
```
atetubou, line 68:
```suggestion
	// Forward-slash-separated paths are matched against the regexp by default, so
```

atetubou:
Is it difficult to fix our tests?

nodirg:
> Is it difficult to fix our tests?

our tests are correct. Can you elaborate?

atetubou:
I meant to do path conversion in our code instead of SDK.
Or can we even disallow backslach in our filter?

ramymedhat:
> I meant to do path conversion in our code instead of SDK.
> Or can we even disallow backslach in our filter?

+1. The PR lgtm but would it be possible instead to make the isolate regexs use forward slash?

nodirg:
yes, path conversation is hard, because it is a path regexp, not a path literal. For example, a backslash could be `[\]`, `\`, and the replacement is not just `\\`, but `\\\\`, FWIU, I might be missing some other edge cases.

re disallow: this might be possible, but I won't have cycles for hunt for callsites which might be passing backslashes (this would be breaking change). Do you have cycles to do this migration?

atetubou (approved):
OK, then could you leave a TODO to me for the cleanup?

atetubou:
Or you can just change our tests, and let me handle incompatibility when rolling our client.

nodirg:
> Or you can just change our tests, and let me handle incompatibility when rolling our client.

Thanks Takuto! I've dropped support for US-native path patterns and tests are passing now. This should get Linux-migration unblocked. We can deal with Windows a bit later.

PR #365, opened 2021-07-02
add size check in uploadNonUnified

atetubou:
This is to know which file we see InvalidArgument error in https://crbug.com/1224224#c5.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -438,6 +438,11 @@ func (c *Client) uploadNonUnified(ctx context.Context, data ...*uploadinfo.Entry
 					if err != nil {
 						return errors.Wrapf(err, "error while fetching FullData for digest %v", dg)
 					}
+
+					if dg.Size != int64(len(data)) {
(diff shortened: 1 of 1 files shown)


go/pkg/client/cas.go
```diff
@@ -438,6 +438,11 @@ func (c *Client) uploadNonUnified(ctx context.Context, data ...*uploadinfo.Entry
 					if err != nil {
 						return errors.Wrapf(err, "error while fetching FullData for digest %v", dg)
 					}
+
+					if dg.Size != int64(len(data)) {
+						return errors.Errorf("blob size changed while uploading, given:%d now:%d for %s", dg.Size, int64(len(data)), ue.Path)
```
nodirg, line 443: This looks like something that must never happen, so I'd be OK with even panic (like assertion), but this is good too

nodirg (approved):

go/pkg/client/cas.go
atetubou: Ack, let me return error this time.

PR #375, opened 2021-07-29
cas: check file existence for large file

atetubou:
Zstd encoder seems reading some amount of file content before uploader
got response about existing file from byte stream client.
This avoids such unnecessary file read.

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -651,6 +651,16 @@ func (u *uploader) visitRegularFile(ctx context.Context, absPath string, info os
 		// advantage of ByteStream's built-in presence check.
 		// https://github.com/bazelbuild/remote-apis/blob/0cd22f7b466ced15d7803e8845d08d3e8d2c51bc/build/bazel/remote/execution/v2/remote_execution.proto#L250-L254
 
+		if res, err := u.findMissingBlobs(ctx, []*uploadItem{item}); err != nil {
+			return nil, errors.Wrapf(err, "failed to check existence")
+		} else if len(res.MissingBlobDigests) == 0 {
+			// File is already there.
+			log.Infof("do not upload %s", absPath)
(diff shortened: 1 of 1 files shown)


go/pkg/cas/upload.go
```diff
@@ -651,6 +651,16 @@ func (u *uploader) visitRegularFile(ctx context.Context, absPath string, info os
 		// advantage of ByteStream's built-in presence check.
 		// https://github.com/bazelbuild/remote-apis/blob/0cd22f7b466ced15d7803e8845d08d3e8d2c51bc/build/bazel/remote/execution/v2/remote_execution.proto#L250-L254
 
+		if res, err := u.findMissingBlobs(ctx, []*uploadItem{item}); err != nil {
+			return nil, errors.Wrapf(err, "failed to check existence")
+		} else if len(res.MissingBlobDigests) == 0 {
```
Jwata, line 656: nit: I wonder "if" is better than "else if" since it's not an error case.

```
res, err := u.findMissinBlobs(...)
if err != nil {
  ...
}
len(res.MissingBlobDigests) == 0 {
  ...
}
```

Jwata (approved):

go/pkg/cas/upload.go
```diff
@@ -651,6 +651,16 @@ func (u *uploader) visitRegularFile(ctx context.Context, absPath string, info os
...
+			// File is already there.
+			log.Infof("do not upload %s", absPath)
```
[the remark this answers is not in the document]
atetubou: done

go/pkg/cas/upload.go
atetubou: I prefer to have narrower scope for variables here.

Jwata (approved):

go/pkg/cas/upload.go
Jwata: I see, ok.

PR #377, opened 2021-08-10
cas: elaborate error from os.Lstat in visitSymlink

atetubou:
This is to debug https://crbug.com/1237607

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -398,7 +398,7 @@ func (c *Client) Upload(ctx context.Context, opt UploadOptions, inputC <-chan *U
 		log.Errorf("failed to upload at least one blob: %+v", err)
 	}
 
-	return &UploadResult{Stats: u.stats, u: u}, err
+	return &UploadResult{Stats: u.stats, u: u}, errors.WithStack(err)
 }
 
 // uploader implements a concurrent multi-stage pipeline to read blobs from the
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -392,7 +392,13 @@ func (c *Client) Upload(ctx context.Context, opt UploadOptions, inputC <-chan *U
 		}
 	})
 
-	return &UploadResult{Stats: u.stats, u: u}, errors.WithStack(eg.Wait())
+	err := eg.Wait()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


go/pkg/cas/upload.go
```diff
@@ -395,7 +395,13 @@ func (c *Client) Upload(ctx context.Context, opt UploadOptions, inputC <-chan *U
 		}
 	})
 
-	return &UploadResult{Stats: u.stats, u: u}, errors.WithStack(eg.Wait())
+	err := eg.Wait()
+	if err != nil {
+		// Caller may not unwrap stack frame of github.com/pkg/error, so log error here.
+		log.Errorf("upload err %+v", err)
```
rubensf, line 401: Maybe a more descriptive message? If only for a client grepping for error code to find out easier :) eg "failed to upload at least one blob: %+v"

rubensf (approved):

go/pkg/cas/upload.go
```diff
@@ -395,7 +395,13 @@ func (c *Client) Upload(ctx context.Context, opt UploadOptions, inputC <-chan *U
...
+	}
+
+	return &UploadResult{Stats: u.stats, u: u}, err
```
nodirg, line 404: Why not have the callsite log the error? Modifying a common library seems more intrusive than the less global callsite.

nodirg (approved):

go/pkg/cas/upload.go
atetubou: Are you suggesting to support stack extraction from github.com/pkg/errors in go.chromium.org/luci/common/errors?

go/pkg/cas/upload.go
atetubou: done

Jwata (approved):

go/pkg/cas/upload.go
nodirg: no? I mean, this code returns the exactly same error object that it logs. It means the callsite can do the same.

nodirg: I see the latest version now returns an error with a stack annotation.
I am not sure what's the problem logging with stack, but perhaps it does not matter much.

atetubou: I understood what you mean. Thanks.

PR #383, opened 2021-10-15
Make numExecCalls increment atomic

WizardStan:
When execute() is called from multiple threads,
the standard increment of numExecCalls can result
in a race condition leading to incorrect counts.
Using atomic eliminates the issue.

### DIFF
--- a/go/pkg/fakes/exec.go
+++ b/go/pkg/fakes/exec.go
@@ -3,6 +3,7 @@ package fakes
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"testing"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/client"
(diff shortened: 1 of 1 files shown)


go/pkg/fakes/exec.go
```diff
@@ -144,7 +145,7 @@ func (s *Exec) Execute(req *repb.ExecuteRequest, stream regrpc.Execution_Execute
 	} else if err = stream.Send(op); err != nil {
 		return err
 	}
-	s.numExecCalls++
+	atomic.AddInt32(&s.numExecCalls, 1)
```
ramymedhat, line 148: Also wrap the call in Clear()? Sorry github won't let me comment on an unchanged line

go/pkg/fakes/exec.go
```diff
@@ -61,7 +62,7 @@ func (s *Exec) Clear() {
 
 // ExecuteCalls returns the total number of Execute calls.
 func (s *Exec) ExecuteCalls() int {
-	return s.numExecCalls
+	return int(s.numExecCalls)
```
ramymedhat, line 65: use atomic Load here as well?

ramymedhat (approved):

PR #386, opened 2021-11-10
Throw errors for incorrectly tagged as empty paths

rubensf:


go/pkg/client/tree.go
```diff
@@ -293,6 +297,9 @@ func buildTree(files map[string]*fileSysNode) *treeNode {
 			if node.dirs == nil {
 				node.dirs = make(map[string]*treeNode)
 			}
+			if node.dirs[base] != nil {
+				return nil, fmt.Errorf("path %v is was tagged as an empty dir but isn't empty", name)
```
andusy, line 301: > %v is was tagged

Just nitpicking, but there's a slight grammar mistake in the error message. Other than that LGTM :)

andusy (approved):

go/pkg/client/tree.go
rubensf: Fixed

PR #387, opened 2021-11-11
Add optional extended file attribute use to hash compute function

WizardStan:
Instructs the cache hash compute function to use the provided file
attribute (if available) as the hash value, instead of loading the
file into memory and hashing its contents.


go/pkg/filemetadata/filemetadata.go
```diff
@@ -30,6 +31,26 @@ type FileError struct {
 	Err        error
 }
 
+type xattributeAccessorInterface interface {
```
rubensf, line 34: Can you leave some comments that this is for test mocking?

go/pkg/filemetadata/filemetadata.go
```diff
@@ -30,6 +31,26 @@ type FileError struct {
...
+}
+
+type xattributeAccessor struct{}
+
+func (x xattributeAccessor) getXAttr(path string, name string) ([]byte, error) {
+	return xattr.Get(path, name)
+}
+
+var (
+	XattrName   string
```
rubensf, line 45: wdyt of `XattrDigestName` or `XattrSHA256Name`? Since we're gating digest read on this one

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -64,7 +74,62 @@ func TestComputeFiles(t *testing.T) {
...
+		},
+		{
+			name:       "executable",
+			contents:   "foo",
+			executable: true,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			getXAttrMock = func(path string, name string) ([]byte, error) {
```
rubensf, line 101: `func(_, _ string)`

rubensf (commented):
nits

go/pkg/filemetadata/filemetadata.go
rubensf: Also, maybe this should become a flag?

go/pkg/filemetadata/filemetadata.go
```diff
@@ -79,6 +100,18 @@ func Compute(filename string) *Metadata {
 		md.IsDirectory = true
 		return md
 	}
+
+	if len(XattrName) > 0 {
```
rubensf, line 104: Should we throw an error on windows?

go/pkg/filemetadata/filemetadata.go
ramymedhat: I thought we were going to make this a constant?

go/pkg/filemetadata/filemetadata.go
WizardStan: On Windows, or Linux if extended file attributes aren't supported or doesn't exist, getXAttr will just error and fall through to the original file hashing step.  Do you think there'd be value in explicitly saying "not supported on Windows"?

go/pkg/filemetadata/filemetadata.go
WizardStan: I think "DigestName", on the chance sometime in the far future someone decides SHA256 isn't good enough anymore.

WizardStan: We discussed making it constant because of testing issues, but the issues exist anyway because I'm mocking the xattr package.  Since it didn't save any headache I put it back as a global so it's at least configurable, disabled by leaving it empty.

go/pkg/filemetadata/filemetadata.go
rubensf: IIUC the code just returns "0, nil" on windows: https://github.com/pkg/xattr/blob/master/xattr_unsupported.go#L12, which would likely make the code fail somewhere down the pipeline. Having an explicit error would be easier to root-cause.

WizardStan: You are correct, I was wrong.  I should've dug deeper.  Thank you, I'll add some checks.

WizardStan: Check added

go/pkg/filemetadata/filemetadata.go
```diff
@@ -30,6 +31,32 @@ type FileError struct {
 	Err        error
 }
 
+// External xattr package can be mocked for testing through this interface
+type xattributeAccessorInterface interface {
```
gkousik, line 35: Actually, do you need to mock this in the test? If our tests don't create files at all and rather mock the file creation process, we would also want to mock the fact that we set / get xattributes.

But our tests already have the ability to create temporary files on disk, so why not just directly set / get xattributes from the actual file on disk and skip the mock?

go/pkg/filemetadata/filemetadata.go
```diff
@@ -79,6 +106,18 @@ func Compute(filename string) *Metadata {
 		md.IsDirectory = true
 		return md
 	}
+
+	if len(XattrDigestName) > 0 && XattrAccess.isSupported() {
+		xattrValue, err := XattrAccess.getXAttr(filename, XattrDigestName)
+		if err == nil {
```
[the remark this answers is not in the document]
WizardStan: My intention was that if there's an error (including "attribute not found" and "not supported") it would fall back on the old file hashing. Also didn't want to return the error because getXAttr returns an errnoErr but the Compute() comments promise it'll return a FileError and I'm worried of the possibility that someone has made that assumption somewhere and suddenly returning something else will cause problems, so I'd have to convert.
I could just log the error before falling through?

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -17,7 +18,20 @@ var (
 	ignoreMtime = cmpopts.IgnoreFields(Metadata{}, "MTime")
 )
 
-func TestComputeFiles(t *testing.T) {
+var getXAttrMock func(path string, name string) ([]byte, error)
```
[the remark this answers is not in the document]
WizardStan: Moved. Make sure I understood the suggestion correctly and didn't move too much.

go/pkg/filemetadata/filemetadata.go
```diff
@@ -30,6 +31,32 @@ type FileError struct {
...
+	return xattr.Get(path, name)
+}
+
+var (
+	XattrDigestName string
+	XattrAccess     xattributeAccessorInterface
+)
+
+func init() {
+	XattrDigestName = ""
```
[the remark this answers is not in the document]
WizardStan: In my reclient code I've got 
`flag.StringVar(&filemetadata.XattrDigestName, "xattr_hash", "", "Extended file attribute to obtain the hash from, if available")`
Do we want this to be on by default?  Because that's how I've chosen to configure it: if the name is there, attempt to use it, otherwise skip to the full file hashing.  If it gets a default value then it'll need to actively be turned off.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -64,7 +78,62 @@ func TestComputeFiles(t *testing.T) {
...
+			}
+			if got.MTime.Before(before) || got.MTime.After(after) {
+				t.Errorf("Compute(%v) returned MTime %v, want time in (%v, %v).", filename, got.MTime, before, after)
+			}
+		})
+	}
+}
+
 func TestComputeDirectory(t *testing.T) {
+	XattrDigestName = ""
```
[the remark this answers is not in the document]
WizardStan: Yeah, that was the issue I was trying to avoid: the Compute() function would need to be changed which could have consequences; that's why I was going to go with the suggestion to just make it a const.  But then I realized the Mock is in the same situation anyway, can't be done in parallel because there's no way to pass extra data through the Compute() function without changing the API.  My logic is that since it needs to be that way to properly unit test anyway, and all tests are already sequential anyway, there's no (current) downside to doing it this way, and plenty of upside.  If we want to optimize the tests in the future to run in parallel the API can be modified then.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -64,7 +78,62 @@ func TestComputeFiles(t *testing.T) {
...
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			getXAttrMock = func(_ string, _ string) ([]byte, error) {
+				return []byte(tc.name), nil
+			}
+
+			before := time.Now().Truncate(time.Second)
+			filename, err := testutil.CreateFile(t, tc.executable, tc.contents)
```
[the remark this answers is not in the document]
WizardStan: We can't even guarantee it's being run on a file system with extended attribute support.  That's what the Mock is for.  My workstation, including the /tmp folder, doesn't have extended attributes turned on, I had to do all integration testing out of google3.
By using the mock and providing a fixed string to return it also gives us verifiability that the Compute() function is actually using the attribute; otherwise the digest for the xattribute and for hashing the file would be identical and I can't think of a way to prove that two identical outputs were produced via two different code paths.

go/pkg/filemetadata/filemetadata.go
WizardStan: See previous response (no guarantees can be made about the file system)
Technically I should also have unit tests for isSupported() returning false, and getXAttr() returning an error which can only be done consistently with mocking

go/pkg/filemetadata/filemetadata.go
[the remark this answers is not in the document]
rubensf: I think we do want to return an error here - and converting is ok. 
Presumably the user will intentionally turn request the configuration/feature at a build level, and it'd generally be frustrating for it to just silently error.

go/pkg/filemetadata/filemetadata.go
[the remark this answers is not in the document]
gkousik: > Do we want this to be on by default? Because that's how I've chosen to configure it: if the name is there, attempt to use it, otherwise skip to the full file hashing. If it gets a default value then it'll need to actively be turned off.

Ah right, no we don't want it to be on by default. What you have in re-client sgtm.

go/pkg/filemetadata/filemetadata_test.go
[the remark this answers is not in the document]
gkousik: Sgtm to do it when we decide on optimizing the test.

go/pkg/filemetadata/filemetadata.go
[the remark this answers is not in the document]
gkousik: +1 to _not_ failing silently and bubbling up the error!

go/pkg/filemetadata/filemetadata_test.go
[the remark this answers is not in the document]
gkousik: > We can't even guarantee it's being run on a file system with extended attribute support. That's what the Mock is for. My workstation, including the /tmp folder, doesn't have extended attributes turned on, I had to do all integration testing out of google3.

Aha ok, makes sense.

go/pkg/filemetadata/filemetadata.go
gkousik: Right, sg!

go/pkg/filemetadata/filemetadata.go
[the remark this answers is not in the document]
WizardStan: Fair enough.  Fixed, and I've also made it error when xattr is not supported.

go/pkg/filemetadata/filemetadata_test.go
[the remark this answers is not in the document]
gkousik: Yep, lgtm now!

gkousik (approved):

rubensf (approved):

PR #391, opened 2021-12-08
Remove GetOutputDigests and add OutputDirectoryDigests to Metadata

andusy:
- Remove Remove GetOutputDigests function
- Rename OutputDigests in Metadata to OutputFileDigests
- Add new OutputDirectoryDigests field to Metadata
- Set directory digests in setOutputMetadata


go/pkg/rexec/rexec.go
```diff
@@ -419,27 +419,21 @@ func (ec *Context) GetOutputFileDigests(useAbsPath bool) (map[string]digest.Dige
 // This function only returns the digests of output files and output directories.
 // GetOutputFileDigests will return the digest of output files as well as the digest of all the files
 // that exist under the output directories.
+// This function is supposed to be run after a successful cache-hit / remote-execution
+// has been run with the given execution context. If called before the completion of
+// remote-execution, the function returns a nil result.
```
ramymedhat, line 424: It returns an error in this case?

go/pkg/rexec/rexec_test.go
```diff
@@ -441,71 +441,6 @@ func TestGetOutputFileDigests(t *testing.T) {
 	}
 }
 
-func TestGetOutputDigests(t *testing.T) {
```
ramymedhat, line 444: Why remove this test?

andusy: I can add a test if you think it makes sense. I was just thinking that there's not really any new logic here since I changed it so that this function no longer computes the directory digest. Instead it is just exposing whatever is in the ActionResult after a remote execution has been run.

go/pkg/rexec/rexec.go
andusy: fixed

go/pkg/rexec/rexec_test.go
ramymedhat: Yeah I think we should try to always have tests for our public functions to maintain coverage.

go/pkg/rexec/rexec.go
```diff
@@ -419,27 +419,21 @@ func (ec *Context) GetOutputFileDigests(useAbsPath bool) (map[string]digest.Dige
 // This function only returns the digests of output files and output directories.
 // GetOutputFileDigests will return the digest of output files as well as the digest of all the files
 // that exist under the output directories.
+// This function is supposed to be run after a successful cache-hit / remote-execution
+// has been run with the given execution context. If called before the completion of
+// remote-execution, the function returns an error.
 func (ec *Context) GetOutputDigests() (outputFileDigests map[string]digest.Digest, outputDirectoryDigests map[string]digest.Digest, err error) {
```
rubensf, line 425: Looking at it now, `GetOutputDigests` just passes forward `ec.respPb.OutputFiles/OutputDirectories`.

It may make more sense to just remove this and instead add `OutputDirectoriesDigests` to the `ec.Metadata`? (since it already has `OutputDigests`.

gkousik: So we discussed this. We do want to separate out OutputFile digests and OutputDir digests and that means ec.Metadata now will have three things - OutputDigests, OutputFileDigests and OutputDirDigests. Given that, it seemed easier to add a function to return this info instead of sideloading it through the Metadata.

go/pkg/rexec/rexec_test.go
andusy: I added the test, but currently the fake server doesn't support output directories so I left a TODO on line 474.

andusy: Test no longer needed since the function is removed.

go/pkg/rexec/rexec.go
andusy: As discussed in the meeting, I changed the OutputDigests field to OutputFileDigests and added a new OutputDirectoryDigests field.

rubensf (approved):

gkousik (approved):
Thanks for the changes!

PR #392, opened 2021-12-22
Remove execution retries based on operation errors.

bergsieker:
Retrying based on an operation error is always incorrect, although
mostly harmless. Once an operation has an error, it is marked as Done
and should never be unmarked, so retries will simply fetch the same
operation (with the same error) again.

### DIFF
--- a/go/pkg/client/exec.go
+++ b/go/pkg/client/exec.go
@@ -218,12 +218,6 @@ func buildCommand(ac *Action) *repb.Command {
 //   1) If an error occurs before the first operation is returned, or after the final operation is
 //      returned (i.e. the one with op.Done==true), retry by calling Execute again.
 //   2) Otherwise, retry by calling WaitExecution with the last operation name.
-// In addition, we want the retrier to trigger based on certain operation statuses as well as on
-// explicit errors. (The shouldRetry function knows which statuses.) We do this by mapping statuses,
-// if present, to errors inside the closure and then throwing away such "fake" errors outside the
-// closure (if we ran out of retries or if there was never a retrier enabled). The exception is
-// deadline-exceeded statuses, which we never give to the retrier (and hence will always propagate
-// directly to the caller).
 func (c *Client) ExecuteAndWait(ctx context.Context, req *repb.ExecuteRequest) (op *oppb.Operation, err error) {
 	return c.ExecuteAndWaitProgress(ctx, req, nil)
 }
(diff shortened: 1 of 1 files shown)


go/pkg/client/exec.go
```diff
@@ -263,18 +256,10 @@ func (c *Client) ExecuteAndWaitProgress(ctx context.Context, req *repb.ExecuteRe
...
-			if st.Code() == codes.DeadlineExceeded {
-				return nil
-			}
-			return st.Err()
-		}
 		return nil
 	}
 	err = c.Retrier.Do(ctx, func() error { return c.CallWithTimeout(ctx, "Execute", closure) })
-	if err != nil && !opError {
+	if err != nil {
```
bozydarsz, line 262: If I got this right, the intent of the changes is to not retry if operation if it returned the error which makes sense to me; however, doesn't this line introduce additional change in the behavior? 

Before the change, if `Retrier.Do` returned the error, but `opError` is false `ExecuteAndWait` would still return lastOp (the failed operation) . After the change `lastOp` is always discarded whenever err != nil, potentially removing part of the information from the client. Is this also an intended change?

bergsieker: I don't think that's correct. Before this change, if Retrier.Do returned an error, *and* opError was false, we would take the branch and return (nil, err). It was only when opError was *true* that we'd skip the branch and return (lastOp, nil).

With the current change, if we get to the point where we previously would have checked for an operation failure, we know that Retrier.Do will return nil, so we won't take this branch.

(I find this code really, really hard to reason about...)

go/pkg/client/exec.go
```diff
@@ -263,18 +256,10 @@ func (c *Client) ExecuteAndWaitProgress(ctx context.Context, req *repb.ExecuteRe
 				}
 			}
 		}
-		st := OperationStatus(lastOp)
```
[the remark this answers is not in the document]
bergsieker: OperationStatus is part of the public interface of the client package. I don't think we can remove it without going through some sort of deprecation period. (Note: it's also still used in several unit tests, although if it weren't part of the public interface we could just move it into the test files.)

go/pkg/client/exec.go
bozydarsz: You're right. Thank you for the explanation. Indeed, my concern doesn't seem to be valid.

bozydarsz (approved):
looks good to me

rubensf (approved):

go/pkg/client/exec.go
rubensf: I agree this is terribly hard to read :)

PR #393, opened 2021-12-22
Add testing functionality for WaitExecution

bergsieker:
The existing ExecuteAndWait function will rarely actually call
WaitExecution when used against a live server. This change adds the
ability to terminate the Execute call early and switch to WaitExecution
immediately so that it's possible for service implementors to test
WaitExecution.

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -112,6 +112,10 @@ type Client struct {
 	casDownloadRequests chan *downloadRequest
 	rpcTimeouts         RPCTimeouts
 	creds               credentials.PerRPCCredentials
+	// ForceEarlyWaitCalls specifies whether ExecuteAndWait should call WaitExecution immedately
+	// after receiving an Operation from Execute. It is primarily useful for testing
+	// WaitExecution.
+	ForceEarlyWaitCalls bool
 }
 
 const (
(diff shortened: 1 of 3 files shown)


go/pkg/client/exec.go
```diff
@@ -255,6 +256,9 @@ func (c *Client) ExecuteAndWaitProgress(ctx context.Context, req *repb.ExecuteRe
 					progress(metadata)
 				}
 			}
+			if returnEarly {
+				return status.Error(codes.Internal, "fake error to for wait call")
```
bozydarsz, line 260: nit: "(...) error to for wait (...)", shouldn't it be, "(...) error to trigger the wait (...)"

go/pkg/client/client.go
```diff
@@ -112,6 +112,10 @@ type Client struct {
 	casDownloadRequests chan *downloadRequest
 	rpcTimeouts         RPCTimeouts
 	creds               credentials.PerRPCCredentials
+	// ForceEarlyWaitCalls specifies whether ExecuteAndWait should call WaitExecution immedately
+	// after receiving an Operation from Execute. It is primarily useful for testing
+	// WaitExecution.
+	ForceEarlyWaitCalls bool
```
rubensf, line 118: I'm not inheretly opposed to this... But I think the most appropriate place for this behavior would be on the fake server: https://github.com/bazelbuild/remote-apis-sdks/blob/af1232ee0d79920ab8390fb855f95dca18a69cf0/go/pkg/fakes/exec.go. Wdyt?

PR #395, opened 2022-01-04
Added RemoteWorkingDir

bozydarsz:
RemoteWorkingDir allows to configure RE server to execute the command
in a working directory that's different than local one.
If not provided, the command is executed in WorkingDir

The feature allows RE clients to specify common RemoteWorkingDir value
and, in effect, enable action result caching for build commands that were
executed from different working directories


go/pkg/client/tree.go
```diff
@@ -112,6 +113,18 @@ func getRelPath(base, path string) (string, error) {
 	return rel, nil
 }
 
+// overrideWorkingDir changes workingDir component of execRoot relative path
```
rubensf, line 116: > // overrideWorkingDir changes workingDir component of execRoot relative path

workingDir component of execRoot relative path?

I think this statement is a bit confusing - workingDir shouldn't be part of the exec root.

go/api/command/command.proto
```diff
@@ -33,6 +33,11 @@ message Command {
 
   // The platform to use for the remote execution.
   map<string,string> platform = 8;
+
+  // The working directory when executing the command on RE server.
+	// It's relative to exec root and, if provided, needs to have the same number of levels
+	// as WorkingDir. If not provided, the remote command is run from the WorkingDir
```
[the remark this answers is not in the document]
bozydarsz: Addressed in a follow-up commit

go/pkg/client/tree.go
```diff
@@ -112,6 +113,18 @@ func getRelPath(base, path string) (string, error) {
...
+// from workingDir to remoteWorkingDir
+func overrideWorkingDir(path, workingDir, remoteWorkingDir string) (string, error) {
+	workingDirRelPath, err := filepath.Rel(workingDir, path)
+	if err != nil {
+		return "", fmt.Errorf("Failed to change working dir of %q, err: %v", path, err)
```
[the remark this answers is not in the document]
bozydarsz: Addressed in a follow-up commit

go/pkg/client/tree.go
bozydarsz: Updated the comment and renamed the function to make it role more straightforward

go/pkg/client/tree.go
```diff
@@ -131,9 +132,36 @@ func getTargetRelPath(execRoot, path string, symMeta *filemetadata.SymlinkMetada
 	return relExecRoot, relSymlinkDir, err
 }
 
+// getRemotePath returns a remote path by transforming the local input path
+// in the transformed path workingDir component is replaced with remoteWorkingDir
```
ramymedhat, line 136: Can you rephrase this comment? I think its supposed to be two separate sentences maybe?

go/pkg/client/tree.go
```diff
@@ -131,9 +132,36 @@ func getTargetRelPath(execRoot, path string, symMeta *filemetadata.SymlinkMetada
...
+func getRemotePath(path, workingDir, remoteWorkingDir string) (string, error) {
+	workingDirRelPath, err := filepath.Rel(workingDir, path)
+	if err != nil {
+		return "", fmt.Errorf("getRemotePath failed while trying to get working dir relative path of %q, err: %v", path, err)
+	}
+	remotePath := filepath.Join(remoteWorkingDir, workingDirRelPath)
+	return remotePath, nil
+}
+
+// getNormalizedPaths returns local and remote normalized paths for a given local absolute path
```
ramymedhat, line 146: What is "normalized" in this context, can you clarify please?

go/pkg/fakes/server.go
```diff
@@ -171,7 +171,8 @@ func (e *TestEnv) Set(cmd *command.Command, opt *command.ExecutionOptions, res *
 		}
 	}
 
-	root, inputs, _, err := e.Client.GrpcClient.ComputeMerkleTree(cmd.ExecRoot, cmd.InputSpec, e.Client.FileMetadataCache)
+	execRoot, workingDir, remoteWorkingDir := cmd.ExecRoot, cmd.WorkingDir, cmd.RemoteWorkingDir
```
ramymedhat, line 174: Can you add some verification here to make sure RemoteWorkingDir has the same number of levels as WorkingDir?

go/pkg/client/tree.go
```diff
@@ -131,9 +132,36 @@ func getTargetRelPath(execRoot, path string, symMeta *filemetadata.SymlinkMetada
...
+func getNormalizedPaths(absPath, execRoot, workingDir, remoteWorkingDir string) (string, string, error) {
+	normPath, err := getRelPath(execRoot, absPath)
+	if err != nil {
+		return "", "", err
+	}
+	remoteNormPath := normPath
+	if remoteWorkingDir != "" && remoteWorkingDir != workingDir {
```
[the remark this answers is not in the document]
bozydarsz: with current implementation we return `normPath` both as local and remote path if `remoteWorkingDir` is either empty or equal to `workingDir`. Do you mean changing this behavior or refactoring it to return early?

go/pkg/client/tree.go
bozydarsz: addressed in the latest commit

go/pkg/client/tree.go
bozydarsz: good point. it made more sense within `loadFiles` before extracting it to a separate function but it's not clear now after the context from `loadFiles` is missing. I've renamed it to `getExecRootRelPaths` to clarify the function's purpose

go/pkg/fakes/server.go
bozydarsz: Added #levels verification to `command.Validate` and triggered command.Validate within `server.TestEnv.Set`

rubensf (approved):
+1 on my end

PR #396, opened 2022-01-07
Added com_google_cloud_go_compute dependency

bozydarsz:
The change is required to address recent build failures
that are caused by compute module being carved out of
cloud/google.com/go:
https://github.com/googleapis/google-cloud-go/pull/5273

### DIFF
--- a/remote-apis-sdks-deps.bzl
+++ b/remote-apis-sdks-deps.bzl
@@ -63,7 +63,6 @@ def remote_apis_sdks_go_deps():
         go_repository,
         name = "org_golang_x_oauth2",
         commit = "bf48bf16ab8d622ce64ec6ce98d2c98f916b6303",
-        #commit = "7af32f14d0a25aec7873e0683e8e48dcead159a8",
         importpath = "golang.org/x/oauth2",
     )
     _maybe(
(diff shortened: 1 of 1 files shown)


remote-apis-sdks-deps.bzl
```diff
@@ -63,6 +63,7 @@ def remote_apis_sdks_go_deps():
         go_repository,
         name = "org_golang_x_oauth2",
         commit = "bf48bf16ab8d622ce64ec6ce98d2c98f916b6303",
+        #commit = "7af32f14d0a25aec7873e0683e8e48dcead159a8",
```
rubensf, line 66: Remove this?

rubensf (commented):
You also need to add it to go.mod file (forgot the specifics but there are some go commands eg `go mod add package@version`

remote-apis-sdks-deps.bzl
bozydarsz: done

bozydarsz:
> You also need to add it to go.mod file (forgot the specifics but there are some go commands eg `go mod add package@version`

for some reason only bazel build fails, while `go build ./...` and `go test ./.../` run without issues, but makes sense to add it for the parity. Will post a follow-up commit

rubensf (approved):
After some offline discussion, go.mod files are confusing and didn't care about this change much. go build/test still works.

PR #398, opened 2022-01-14
Add DownloadAction and ExecuteAction operations to remotetool.

ramymedhat:
DownloadAction downloads the action proto, the command proto and the
inputs into a directory. This allows inspection of the full action data
as well as modification and execution via the ExecuteAction operation.

The action is downloaded to a directory with the following structure:
- ac.textproto (the action proto in text format)
- cmd.textproto (the command proto in text format)
- input/ (the root of the input tree)

The above can be modified locally and then executed on the remote server
using the ExecuteAction command. For instance, you can add args to the
command, change environment variables, or change the input tree.


gkousik (approved):
Yay for this!

PR #399, opened 2022-02-07
Ignore empty directory virtual inputs if inputs exist under them.

ramymedhat:
Currently if a treeNode has children and is also marked as an empty directory (virtual input), ComputeMerkleTree returns an error. This behavior is actually inconsistent, since it relies on the order of inputs checked in a loop over a map. This PR makes the behavior consistent and ignores empty directory virtual inputs if physical inputs exist under them. Note that physical input precedence is already enforced in ComputeMerkleTree where a virtual input has the same name as a physical input. This PR extends this behavior to virtual inputs that are ancestors to physical inputs.

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -322,7 +322,9 @@ func buildTree(files map[string]*fileSysNode) (*treeNode, error) {
 				node.dirs = make(map[string]*treeNode)
 			}
 			if node.dirs[base] != nil {
-				return nil, fmt.Errorf("path %v was tagged as an empty dir but isn't empty", name)
+				// Ignore empty directory if there exists a physical file or
+				// directory in this location.
+				continue
 			}
 			node.dirs[base] = &treeNode{}
 			continue
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -321,12 +321,9 @@ func buildTree(files map[string]*fileSysNode) (*treeNode, error) {
 			if node.dirs == nil {
 				node.dirs = make(map[string]*treeNode)
 			}
-			if node.dirs[base] != nil {
-				// Ignore empty directory if there exists a physical file or
-				// directory in this location.
-				continue
+			if node.dirs[base] == nil {
+				node.dirs[base] = &treeNode{}
 			}
-			node.dirs[base] = &treeNode{}
 			continue
 		}
 		if fn.file != nil {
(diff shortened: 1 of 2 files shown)


go/pkg/client/tree.go
```diff
@@ -322,7 +322,9 @@ func buildTree(files map[string]*fileSysNode) (*treeNode, error) {
 				node.dirs = make(map[string]*treeNode)
 			}
 			if node.dirs[base] != nil {
-				return nil, fmt.Errorf("path %v was tagged as an empty dir but isn't empty", name)
+				// Ignore empty directory if there exists a physical file or
+				// directory in this location.
+				continue
 			}
 			node.dirs[base] = &treeNode{}
 			continue
```
bozydarsz, line 330: nit: it might be a matter of taste, but it could be slightly simplified by avoiding unnecessary `continue` statement if written as 
``` 
if node.dirs[base] == nil {
    node.dirs[base] = &treeNode{}
}
```

bozydarsz (approved):
lgtm

go/pkg/client/tree.go
ramymedhat: Agreed. Fixed.

go/pkg/client/tree.go
```diff
@@ -321,10 +321,9 @@ func buildTree(files map[string]*fileSysNode) (*treeNode, error) {
 			if node.dirs == nil {
```
gkousik, line 321: Not related to this change (and don't have to fix it): it looks like buildTree() is supposed to work with files and empty virtual dir is more of a special case handling that it does? If so it maybe more reasonable to have this done in loadFiles()?

gkousik (approved):

go/pkg/client/tree.go
ramymedhat: I think buildTree is a reasonable place since it builds the tree of directories and files, and is the best place to know about collisions. loadFiles doesn't iterate over path segments for instance, so it won't have visibility into collections in ancestor directories.

ramymedhat: Ah I see what you're saying. I think its fair to rethink the interface of this function to not give that impression, the core of it is to construct a tree of directories and files. It should take in leaf nodes, whether directories or files, and build a tree out of them. Perhaps we can address this when we refactor the whole virtual inputs area.

gkousik: sgtm!

rubensf:
FYI you could tweak to somewhat revert https://github.com/bazelbuild/remote-apis-sdks/pull/386, as buildTree returns no error anymore.

PR #406, opened 2022-02-22
Add zstd compression check in SDK capabilities

rubensf:


go/pkg/client/capabilities.go
```diff
@@ -28,6 +28,18 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
...
+	if c.serverCaps.CacheCapabilities.SupportedCompressors != nil && c.CompressedBytestreamThreshold >= 0  {
+		foundZstd := false
+		for _, sComp := range c.serverCaps.CacheCapabilities.SupportedCompressors {
+			if sComp == repb.Compressor_ZSTD {
+				foundZstd = true
+				break
+			}
+		}
+		if !foundZstd {
+			return errors.New("zstd compression is not supported by server")
```
bozydarsz, line 40: should it be returned also when `CompressedBytestreamThreshold >= 0` and `SupportedCompressors = nil`?

bozydarsz (commented):
it might be worth updating the comment of `CheckCapabilities` function to indicate that it checks whether zstd compressor is supported if the compression is enabled in client config

go/pkg/client/capabilities.go
```diff
@@ -28,6 +28,18 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
...
 	}
+	if c.serverCaps.CacheCapabilities.SupportedCompressors != nil && c.CompressedBytestreamThreshold >= 0  {
+		foundZstd := false
+		for _, sComp := range c.serverCaps.CacheCapabilities.SupportedCompressors {
+			if sComp == repb.Compressor_ZSTD {
+				foundZstd = true
+				break
+			}
+		}
+		if !foundZstd {
```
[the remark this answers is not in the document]
rubensf: Updated the error message - zstd is the only compression supported by the SDK (technically the only other one that the API supports is zip, but we haven't had a need to add support: https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L1735).

go/pkg/client/capabilities.go
```diff
@@ -28,6 +28,18 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
 	if c.serverCaps.CacheCapabilities != nil {
 		c.MaxBatchSize = MaxBatchSize(c.serverCaps.CacheCapabilities.MaxBatchTotalSizeBytes)
 	}
+	if c.serverCaps.CacheCapabilities.SupportedCompressors != nil && c.CompressedBytestreamThreshold >= 0  {
```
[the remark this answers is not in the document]
rubensf: Oh, good point! I updated to add a `useCompression` variable. PTAL

go/pkg/client/capabilities.go
rubensf: Yup -fixed.

go/pkg/client/capabilities.go
```diff
@@ -28,6 +28,23 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
 	if c.serverCaps.CacheCapabilities != nil {
 		c.MaxBatchSize = MaxBatchSize(c.serverCaps.CacheCapabilities.MaxBatchTotalSizeBytes)
 	}
+
+	if useCompression := c.CompressedBytestreamThreshold >= 0; useCompression {
```
gkousik, line 32: nit: `compressionSupported` as variable name?

gkousik (approved):

go/pkg/client/capabilities.go
rubensf: Technically, this is whether the _user_ requested to use Compression. 

The next few checks are the ones checking whether compression is supported :)

gkousik: Ah ok, then there isn't a need to use a variable here to explicitly identify it (it was a user specified option afterall). But tbh, I'm just ambivalent about it, so lgtm!

rubensf: it's purely for readability (particularly if someone is dropped in midst of the code base) - see bozydar's comment above

gkousik: Yep I saw that comment (which was what I disagreed with). I think the condition `c.CompressedBytestreamThreshold >= 0` is in itself very straightforward in stating that this block will be executed only if we are going to be using compression (there's very little room for misinterpretation).
Using a variable also seemed like a misdirection to me (I was searching for where this variable was being used again) - but again, this is all very subjective, which is why I didn't want to emphasize on it :)

PR #407, opened 2022-02-28
Expose errors in closing the Writer.

ramymedhat:
In case the machine runs out of disk space, the write error will be
captured in the done channel, but is suppressed if no other errors
were encountered. This PR exposes the error in the done channel.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -1002,12 +1002,12 @@ func (c *Client) readBlobStreamed(ctx context.Context, d digest.Digest, offset, 
 
 			if err == nil && errC != nil {
 				err = errC
-			} else {
+			} else if errC != nil {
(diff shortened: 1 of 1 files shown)


go/pkg/client/cas.go
```diff
@@ -1000,11 +1000,15 @@ func (c *Client) readBlobStreamed(ctx context.Context, d digest.Digest, offset,
...
+			if err == nil && errC != nil {
 				err = errC
+			} else {
+				log.Errorf("Failed to close writer: %v", errC)
 			}
-			if err != nil && errD != nil {
-				err = fmt.Errorf("Failed to finalize writing downloaded data downstream: %v", err)
+			if err == nil && errD != nil {
+				err = errD
+			} else {
```
rubensf, line 1010: else if errD != nil -- idem above

rubensf (approved):

PR #408, opened 2022-02-28
Improve error message when symlink target is out of tree.

ramymedhat:


atetubou (changes requested):
fix build?

go/pkg/client/tree.go
```diff
@@ -192,7 +192,7 @@ func loadFiles(execRoot, localWorkingDir, remoteWorkingDir string, excl []*comma
 			}
 			targetExecRoot, targetSymDir, err := getTargetRelPath(execRoot, normPath, meta.Symlink)
 			if err != nil {
-				return err
+				return errors.Wrapf(err, "failed to determine the target of symlink %v as a child of %v", normPath, execRoot)
```
atetubou, line 195:
```suggestion
				return errors.Wrapf(err, "failed to determine the target of symlink %q as a child of %q", normPath, execRoot)
```

atetubou (approved):

PR #410, opened 2022-03-02
Respect per RPC creds even during Dial()

gkousik:
Also added an option "use_external_auth_token" that when set to true, would mandate the usae of per RPC creds option starting with the initial Dial to setup the connection

Test: Also implemented the corresponding re-client side of things and ensured that reproxy starts up

### DIFF
--- a/go/pkg/flags/flags.go
+++ b/go/pkg/flags/flags.go
@@ -33,6 +33,8 @@ var (
 	UseGCECredentials = flag.Bool("use_gce_credentials", false, "If true (and --use_application_default_credentials is false), use the default GCE credentials to authenticate with remote execution.")
 	// UseRPCCredentials can be set to false to disable all per-RPC credentials.
 	UseRPCCredentials = flag.Bool("use_rpc_credentials", true, "If false, no per-RPC credentials will be used (disables --credential_file, --use_application_default_credentials, and --use_gce_credentials.")
+	// UseExternalAuthToken specifies whether to use an externally provided auth token, given via PerRPCCreds dial option, should be used.
+	UseExternalAuthToken = flag.Bool("use_external_auth_token", false, "If true, se an externally provided auth token, given via PerRPCCreds when the SDK is initialized.")
 	// Service represents the host (and, if applicable, port) of the remote execution service.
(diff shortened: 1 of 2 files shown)


go/pkg/flags/flags.go
```diff
@@ -33,6 +33,8 @@ var (
 	UseGCECredentials = flag.Bool("use_gce_credentials", false, "If true (and --use_application_default_credentials is false), use the default GCE credentials to authenticate with remote execution.")
 	// UseRPCCredentials can be set to false to disable all per-RPC credentials.
 	UseRPCCredentials = flag.Bool("use_rpc_credentials", true, "If false, no per-RPC credentials will be used (disables --credential_file, --use_application_default_credentials, and --use_gce_credentials.")
+	// UseExternalAuthToken specifies whether to use an externally provided auth token, given via PerRPCCreds dial option, should be used.
+	UseExternalAuthToken = flag.Bool("use_external_auth_token", false, "If true, uses a blah blah blah.")
```
ramymedhat, line 37: Fix comment?

go/pkg/flags/flags.go
```diff
@@ -93,6 +95,14 @@ func NewClientFromFlags(ctx context.Context, opts ...client.Opt) (*client.Client
 		}
 		opts = append(opts, client.RPCTimeouts(timeouts))
 	}
+	var perRPCCreds *client.PerRPCCreds
+	for _, opt := range opts {
+		switch opt.(type) {
```
ramymedhat, line 100: Just an if would suffice?

ramymedhat (approved):

go/pkg/flags/flags.go
gkousik: ah actually its not possible to use `.(type)` outside a switch case - https://stackoverflow.com/questions/34820469/can-i-compare-variable-types-with-type-in-golang.

go/pkg/flags/flags.go
gkousik: Ah embarassing :D done!

ramymedhat (approved):

PR #415, opened 2022-03-28
Returned a customized error when initializing a new client

gkousik:
This is to also indicate the type of authenticated used by the SDK to
connect to the RBE so that clients of the SDK can report better error
messages to indicate how the authentication failure can be fixed.

Test: Applied this change in re-client and confirmed that the returned
object has the proper auth type set.

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -43,6 +43,62 @@ const (
 	HomeDirMacro = "${HOME}"
 )
 
+// AuthType indicates the type of authentication being used.
+type AuthType int
+
+const (
+	// Unknown auth type.
+	UnknownAuth AuthType = iota
+
+	// No authentication used to connect to the RBE service.
+	NoAuth
+
+	// External auth token used to connect to the RBE service.
+	ExternalTokenAuth
+
+	// A JSON credentials file used to connect to the RBE service.
+	CredsFileAuth
+
+	// Application default credentials used to connect to the RBE service.
+	ApplicationDefaultCredsAuth
+
+	// GCE machine credentials used to connect to the RBE service.
+	GCECredsAuth
+)
+
+// String returns a human readable form of authentication used to connect to RBE.
+func (a AuthType) String() string {
+	switch a {
+	case NoAuth:
+		return "no authentication"
+	case ExternalTokenAuth:
+		return "external authentication token (gcert?)"
+	case CredsFileAuth:
+		return "credentials file"
+	case ApplicationDefaultCredsAuth:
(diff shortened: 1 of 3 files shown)


go/pkg/client/client.go
```diff
@@ -401,13 +455,16 @@ func getImpersonatedRPCCreds(ctx context.Context, actAs string, cred credentials
 	}
 }
 
-func getRPCCreds(ctx context.Context, credFile string, useApplicationDefault bool, useComputeEngine bool) (credentials.PerRPCCredentials, error) {
+func getRPCCreds(ctx context.Context, credFile string, useApplicationDefault bool, useComputeEngine bool, authUsed *AuthType) (credentials.PerRPCCredentials, error) {
```
WizardStan, line 458: Why not return the AuthType?

gkousik: I initially thought about it, but it involved a bit of change to predeclare variables and I figured just passing a pointer would make it easier. But now that I think about it, it seems cleaner and standardized to keep return values as return values instead of as pointers passed to the function. Changed!

WizardStan (approved):

go/pkg/client/client.go
```diff
@@ -43,6 +43,60 @@ const (
...
+		at = "no authentication"
+	case ExternalTokenAuth:
+		at = "external authentication token (gcert?)"
+	case CredsFileAuth:
+		at = "credentials file"
+	case ApplicationDefaultCredsAuth:
+		at = "application default credentials"
+	case GCECredsAuth:
+		at = "gce credentials"
+	}
```
bozydarsz, line 96: nit: this could be actually a part of `AuthType` enum definition (`func (s AuthType) String() string`)

bozydarsz (approved):

go/pkg/client/client.go
gkousik: Done!

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -41,6 +41,7 @@ func TestComputeFilesNoXattr(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			before := time.Now().Truncate(time.Second)
+			time.Sleep(5 * time.Second)
```
gkousik, line 44: fyi @WizardStan This is fixing a flaky test (mtimes have a margin of error and are rounded to the nearest second, so we usually have a sleep in the tests that test for timestamp just before we create the file).

PR #416, opened 2022-04-14
Fix all lint errors and enable golint

gkousik:
This is so that we catch lint issues here before we import the SDK to
google3. Also added the lint check to presubmit hooks just like how we
do for gofmt.

### DIFF
--- a/check-golint.sh
+++ b/check-golint.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+linterrs=$(golint -min_confidence 1.0 ./...)
+[ -z "$linterrs" ] && exit 0
+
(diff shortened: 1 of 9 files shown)


check-golint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+linterrs=$(golint -min_confidence 1.0 ./...)
+[ -z "$linterrs" ] && exit 0
+
+# Some files are not gofmt'd. Print message and fail.
```
andusy, line 6: nit: golint'd

andusy (approved):

check-golint.sh
gkousik: Done!

PR #421, opened 2022-11-10
Download Specified Outputs

banikharbanda:
Adds functions to retrieve outputs and download specified outputs instead of the ones in working dir. Also adds a new command option for preserving mtimes of unchanged outputs.

### DIFF
--- a/go/pkg/rexec/rexec.go
+++ b/go/pkg/rexec/rexec.go
@@ -335,6 +335,7 @@ func (ec *Context) ExecuteRemotely() {
 		ec.Result = command.NewRemoteErrorResult(fmt.Errorf("unexpected operation result type: %v", or))
 		return
 	}
+
 	resp := &repb.ExecuteResponse{}
 	if err := or.UnmarshalTo(resp); err != nil {
 		ec.Result = command.NewRemoteErrorResult(err)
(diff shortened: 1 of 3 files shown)


go/pkg/rexec/rexec.go
```diff
@@ -399,6 +401,36 @@ func (ec *Context) DownloadOutputs(outputDir string) {
...
+	}
+}
+
+// GetFlattenedOutputs flattens the outputs from the ActionResult of the context and returns
+// a map of output paths relative to the working directory and their corresponding TreeOutput
+func (ec *Context) GetFlattenedOutputs() (map[string]*rc.TreeOutput, error) {
+	if ec.resPb == nil {
+		return nil, nil
+	}
+	return ec.client.GrpcClient.FlattenActionOutputs(ec.ctx, ec.resPb)
```
WizardStan, line 431: Replace this with something like
```
out, err := ec.client.GrpcClient.FlattenActionOutputs(ec.ctx, ec.resPb)
if err != nil {
    return nil, fmt.Errorf("Failed to flatten outputs: %v", err)
}
return out, nil
```

WizardStan (approved):

PR #422, opened 2022-12-02
Optionally process symlinks pointing outside exec root

orn688:
Previously if `TreeSymlinkOpts.Preserved` was set to `true`, symlinks pointing outside of the exec root would always be rejected. However, it can be useful to preserve symlinks by default while replacing them with the targeted file if the targeted file is outside of the exec root.

So add a new `TreeSymlinkOpts.MaterializeOutsideExecRoot` option that enables this behavior if set to true.

See https://crbug.com/1216363 for context on why this is useful.

### DIFF
--- a/go/pkg/client/BUILD.bazel
+++ b/go/pkg/client/BUILD.bazel
@@ -84,6 +84,7 @@ go_test(
         "@org_golang_google_grpc//codes:go_default_library",
         "@org_golang_google_grpc//status:go_default_library",
         "@org_golang_google_protobuf//proto:go_default_library",
+        "@org_golang_google_protobuf//testing/protocmp",
         "@org_golang_google_protobuf//types/known/anypb:go_default_library",
         "@org_golang_google_protobuf//types/known/emptypb:go_default_library",
         "@org_golang_x_sync//errgroup:go_default_library",
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -66,6 +66,11 @@ type TreeSymlinkOpts struct {
 	Preserved bool
 	// If true, the symlink target (if not dangling) is followed.
 	FollowsTarget bool
+	// If true, overrides Preserved=true for symlinks that point outside the
+	// exec root, converting them into their targeted files while preserving
+	// symlinks that point to files within the exec root.  Has no effect if
+	// Preserved=false, as all symlinks are materialized.
+	MaterializeOutsideExecRoot bool
 }
 
 // DefaultTreeSymlinkOpts returns a default DefaultTreeSymlinkOpts object.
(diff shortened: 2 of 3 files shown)


orn688:
The code changes are a bit ugly, but I wanted to get feedback to make sure the general idea is okay before spending too much time optimizing the code. I'm also happy to add more tests.

gkousik:
Apologies for the late review on this (and thanks for the ping)!

I think this should be an ok change to make. In re-client (which is one of the main users of the remote-apis-sdks), we do not want to pickup anything outside of exec-root, however, I think the SDK can be flexible enough to allow this with this option. We can probably add a warning to this flag and clearly indicate that this has the possibility of making an action non-hermetic (particularly because now you are depending on a file outside the execution root in case of sandboxed actions).

Do you want to take a stab at refactoring the if{} conditions as you described to make it cleaner? I can do another review post that!

maruel:
@gkousik this is a regression of behavior from Isolate, and I recall that ChromeOS was (at least a few years ago) strongly dependent on this behavior. It would be interesting to double-check what changed since on their side.

orn688:
@gkousik thanks! I spent a while trying to clean up the code. The best option I could find uses `goto`, so not ideal but I think it's better than the alternatives (convoluted if statements or confusing closures). Let me know what you think.

gkousik:
Hi Oliver! I have two questions on https://github.com/bazelbuild/remote-apis-sdks/pull/422

1. Did you look at maruel@'s comment on the change about behavior change in Isolate? 

2. On the change itself, did you verify that the contents of the symlink were replaced with the file contents when the file is outside the exec root? I am a little uncertain that the code does that - in particular, from reading the code, I'm not sure how its replacing the symlink with the target file, but perhaps that's at a different location.

gkousik:
Oh gotcha! Filtering for symlink targets within the exec-root should have been the default behaviour (maybe we fixed this recently). So in that regard - I think optionally escaping that behaviour like what Oliver has done in this change makes sense (i think we can focus the discussion on making sure the change does what we need).

orn688:
> 2. On the change itself, did you verify that the contents of the symlink were replaced with the file contents when the file is outside the exec root? I am a little uncertain that the code does that - in particular, from reading the code, I'm not sure how its replacing the symlink with the target file, but perhaps that's at a different location.

Yes, sorry it's not very clear – the `goto processNonSymlink` switches from the symlink processing codepath to the regular file processing codepath (resulting in the symlink being replaced with the file contents) when it's pointing outside the exec root. I'm not very happy I had to resort to a `goto` but I think it's the least invasive option.

gkousik (approved):

orn688:
Thanks! I resolved the merge conflict, let me know if there's anything else I need to do to get this merged.

PR #425, opened 2023-01-27
Update writeChunked method in bytestream.go to allow resumable upload with ByteStream Write API.

YukaSadaoka:
Update writeChunked method in bytestream.go to allow resumable upload with ByteStream Write API.

I also added a new struct ResumableWriteOpts and it’s added to Client to keep track of write offset so that we can start writing data at an arbitrary offset.

### DIFF
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -29,6 +29,14 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 // writeChunked uploads chunked data with a given resource name to the CAS.
 func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
 	var totalBytes int64
+	var resume bool
+	var offset int64
+	var initialOffset int64
+	if c.ResumableWriteOpts != nil {
+		resume = true
+		initialOffset = c.ResumableWriteOpts.LastLogicalOffset
+		offset = initialOffset
+	}
 	closure := func() error {
 		// Retry by starting the stream from the beginning.
 		if err := ch.Reset(); err != nil {
@@ -50,9 +58,14 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
 			if chunk.Offset == 0 {
 				req.ResourceName = name
 			}
+			if resume {
+				req.WriteOffset = offset
+			} else {
+				req.WriteOffset = chunk.Offset
+			}
 			req.WriteOffset = chunk.Offset
 			req.Data = chunk.Data
-			if !ch.HasNext() {
+			if (!resume && !ch.HasNext()) || (resume && !ch.HasNext() && c.ResumableWriteOpts.FinishWrite) {
 				req.FinishWrite = true
 			}
 			err = c.CallWithTimeout(ctx, "Write", func(_ context.Context) error { return stream.Send(req) })
@@ -63,6 +76,7 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
 				return err
 			}
 			totalBytes += int64(len(req.Data))
+			offset = initialOffset + totalBytes
 		}
 		if _, err := stream.CloseAndRecv(); err != nil {
 			return err
@@ -70,6 +84,9 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
 		return nil
 	}
 	err := c.Retrier.Do(ctx, closure)
+	if err == nil && resume {
+		c.ResumableWriteOpts.LastLogicalOffset = offset
+	}
 	return totalBytes, err
 }
 
@@ -116,7 +133,6 @@ func (c *Client) readStreamed(ctx context.Context, name string, offset, limit in
 	if err != nil {
 		return 0, err
 	}
-
 	var n int64
 	for {
 		var resp *bspb.ReadResponse
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -15,6 +15,28 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
 )
 
+// ByteStreamWriteOption sets options required for resumable upload.
+type ByteStreamWriteOption interface {
+	Apply(*ByteStreamWriteOpts)
+}
+
+// ByteSteramOptFinishWrite is a boolean flag to set FinishWrite
+// in ByteStream.WriteRequest.
+type ByteSteramOptFinishWrite bool
+
+// Apply sets ByteSteramOptFinishWrite in ByteStreamWriteOpts.
+func (o ByteSteramOptFinishWrite) Apply(opts *ByteStreamWriteOpts) {
+	opts.FinishWrite = bool(o)
+}
+
+// ByteStreamOptOffset is offset for resumable upload.
+type ByteStreamOptOffset int64
+
+// Apply sets ByteStreamOptOffset in ByteStreamWriteOpts.
+func (o ByteStreamOptOffset) Apply(opts *ByteStreamWriteOpts) {
+	opts.LastLogicalOffset = int64(o)
+}
+
 // WriteBytes uploads a byte slice.
 func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error {
 	ue := uploadinfo.EntryFromBlob(data)
@@ -26,67 +48,30 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 	return err
 }
 
-// writeChunked uploads chunked data with a given resource name to the CAS.
-func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
-	var totalBytes int64
-	closure := func() error {
-		// Retry by starting the stream from the beginning.
-		if err := ch.Reset(); err != nil {
-			return errors.Wrap(err, "failed to Reset")
-		}
-		totalBytes = int64(0)
-		// TODO(olaola): implement resumable uploads.
-
-		stream, err := c.Write(ctx)
-		if err != nil {
-			return err
-		}
-		for ch.HasNext() {
-			req := &bspb.WriteRequest{}
-			chunk, err := ch.Next()
-			if err != nil {
-				return err
-			}
-			if chunk.Offset == 0 {
-				req.ResourceName = name
-			}
-			req.WriteOffset = chunk.Offset
-			req.Data = chunk.Data
-			if !ch.HasNext() {
-				req.FinishWrite = true
-			}
-			err = c.CallWithTimeout(ctx, "Write", func(_ context.Context) error { return stream.Send(req) })
-			if err == io.EOF {
-				break
-			}
-			if err != nil {
-				return err
-			}
-			totalBytes += int64(len(req.Data))
-		}
-		if _, err := stream.CloseAndRecv(); err != nil {
-			return err
-		}
-		return nil
-	}
-	err := c.Retrier.Do(ctx, closure)
-	return totalBytes, err
-}
-
-// WriteBytesWithOffset uploads a byte slice.
-func (c *Client) WriteBytesWithOffset(ctx context.Context, name string, data []byte, opts *ByteStreamWriteOpts) (int64, error) {
+// WriteBytesWithOffset uploads a byte slice with a given resource name to the CAS
+// at an arbitrary offset.
+func (c *Client) WriteBytesWithOffset(ctx context.Context, name string, data []byte, opts ...ByteStreamWriteOption) (int64, error) {
 	ue := uploadinfo.EntryFromBlob(data)
 	ch, err := chunker.New(ue, false, int(c.ChunkMaxSize))
 	if err != nil {
 		return 0, err
 	}
-	return c.writeChunkedWithOffset(ctx, name, ch, opts)
+	return c.writeChunked(ctx, name, ch, opts...)
 }
 
-// writeChunkedWithOffset uploads chunked data with a given resource name to the CAS at an arbitrary offset.
-func (c *Client) writeChunkedWithOffset(ctx context.Context, name string, ch *chunker.Chunker, opts *ByteStreamWriteOpts) (int64, error) {
-	var totalBytes int64
+// writeChunked uploads chunked data with a given resource name to the CAS.
+func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker, rawOpts ...ByteStreamWriteOption) (int64, error) {
+	if len(rawOpts) == 1 {
+		return 0, errors.Errorf("Missing an arugment")
+	}
+	opts := &ByteStreamWriteOpts{}
+	var resume bool
+	for _, o := range rawOpts {
+		resume = true
+		o.Apply(opts)
+	}
 
+	var totalBytes int64
 	closure := func() error {
 		// Retry by starting the stream from the beginning.
 		if err := ch.Reset(); err != nil {
@@ -99,14 +84,20 @@ func (c *Client) writeChunkedWithOffset(ctx context.Context, name string, ch *ch
 			return err
 		}
 		for ch.HasNext() {
-			req := &bspb.WriteRequest{ResourceName: name, WriteOffset: opts.LastLogicalOffset + ch.Offset()}
+			req := &bspb.WriteRequest{ResourceName: name}
 			chunk, err := ch.Next()
 			if err != nil {
 				return err
 			}
+			req.WriteOffset = chunk.Offset
+			// If LastLogicaloffset is provided as an option, we want uploading bytes
+			// at the last logcial offset + the chunk offset which is relative to Chunker passed in.
+			if resume {
+				req.WriteOffset += opts.LastLogicalOffset
+			}
 			req.Data = chunk.Data
 
-			if !ch.HasNext() && opts.FinishWrite {
+			if (!resume && !ch.HasNext()) || (resume && !ch.HasNext() && opts.FinishWrite) {
 				req.FinishWrite = true
 			}
 			err = c.CallWithTimeout(ctx, "Write", func(_ context.Context) error { return stream.Send(req) })
@@ -121,7 +112,6 @@ func (c *Client) writeChunkedWithOffset(ctx context.Context, name string, ch *ch
 		if _, err := stream.CloseAndRecv(); err != nil {
 			return err
 		}
-
 		return nil
 	}
 	err := c.Retrier.Do(ctx, closure)
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -29,14 +29,6 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 // writeChunked uploads chunked data with a given resource name to the CAS.
 func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
 	var totalBytes int64
-	var resume bool
-	var offset int64
-	var initialOffset int64
-	if c.ResumableWriteOpts != nil {
-		resume = true
-		initialOffset = c.ResumableWriteOpts.LastLogicalOffset
-		offset = initialOffset
-	}
 	closure := func() error {
 		// Retry by starting the stream from the beginning.
 		if err := ch.Reset(); err != nil {
@@ -58,14 +50,70 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
 			if chunk.Offset == 0 {
 				req.ResourceName = name
 			}
-			if resume {
-				req.WriteOffset = offset
-			} else {
-				req.WriteOffset = chunk.Offset
-			}
 			req.WriteOffset = chunk.Offset
 			req.Data = chunk.Data
-			if (!resume && !ch.HasNext()) || (resume && !ch.HasNext() && c.ResumableWriteOpts.FinishWrite) {
+			if !ch.HasNext() {
+				req.FinishWrite = true
+			}
+			err = c.CallWithTimeout(ctx, "Write", func(_ context.Context) error { return stream.Send(req) })
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			totalBytes += int64(len(req.Data))
+		}
+		if _, err := stream.CloseAndRecv(); err != nil {
+			return err
+		}
+		return nil
+	}
+	err := c.Retrier.Do(ctx, closure)
+	return totalBytes, err
+}
+
+// WriteBytesWithOffset uploads a byte slice.
+func (c *Client) WriteBytesWithOffset(ctx context.Context, name string, data []byte, opts *ByteStreamWriteOpts) (int64, error) {
+	ue := uploadinfo.EntryFromBlob(data)
+	ch, err := chunker.New(ue, false, int(c.ChunkMaxSize))
+	if err != nil {
+		return 0, err
+	}
+	return c.writeChunkedWithOffset(ctx, name, ch, opts)
+}
+
+// writeChunkedWithOffset uploads chunked data with a given resource name to the CAS at an arbitrary offset.
+func (c *Client) writeChunkedWithOffset(ctx context.Context, name string, ch *chunker.Chunker, opts *ByteStreamWriteOpts) (int64, error) {
+	var totalBytes int64
+	var offset = opts.LastLogicalOffset
+	var initialOffset = opts.LastLogicalOffset
+
+	closure := func() error {
+		// Retry by starting the stream from the beginning.
+		if err := ch.Reset(); err != nil {
+			return errors.Wrap(err, "failed to Reset")
+		}
+		totalBytes = int64(0)
+
+		stream, err := c.Write(ctx)
+		if err != nil {
+			return err
+		}
+		for ch.HasNext() {
+			req := &bspb.WriteRequest{}
+			chunk, err := ch.Next()
+			if err != nil {
+				return err
+			}
+			if chunk.Offset == 0 {
+				req.ResourceName = name
+			}
+
+			req.WriteOffset = offset
+			req.Data = chunk.Data
+
+			if !ch.HasNext() && opts.FinishWrite {
 				req.FinishWrite = true
 			}
 			err = c.CallWithTimeout(ctx, "Write", func(_ context.Context) error { return stream.Send(req) })
@@ -81,12 +129,10 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
 		if _, err := stream.CloseAndRecv(); err != nil {
 			return err
 		}
+
 		return nil
 	}
 	err := c.Retrier.Do(ctx, closure)
-	if err == nil && resume {
-		c.ResumableWriteOpts.LastLogicalOffset = offset
-	}
 	return totalBytes, err
 }
 
--- a/go/pkg/client/bytestream.go
+++ b/go/pkg/client/bytestream.go
@@ -179,6 +179,7 @@ func (c *Client) readStreamed(ctx context.Context, name string, offset, limit in
 	if err != nil {
 		return 0, err
 	}
+
 	var n int64
 	for {
 		var resp *bspb.ReadResponse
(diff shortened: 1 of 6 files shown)


go/pkg/client/client.go
```diff
@@ -474,6 +475,17 @@ func getRPCCreds(ctx context.Context, credFile string, useApplicationDefault boo
 	return rpcCreds, CredsFileAuth, nil
 }
 
+// ResumableWriteOpts sets options required for resumable ByteStream write operations.
+type ResumableWriteOpts struct {
+	FinishWrite       bool
+	LastLogicalOffset int64
```
ramymedhat, line 481: I'm not sure this is what you want to do here. This is an option that applies to the entire client, not just a single request. There can be many individual bytestream requests in flight for the same client.

YukaSadaoka: Thanks for the input. Updated the code. 
I removed the field 'ResumableWriteOpts' from Client struct in client.go. I wanted to make the code backward compatible, so I removed changes from the existing method(writeChunked) in bytestream.go and I added new methods taking write options such as offset and a boolean flag for determining to finish a write operation. Please let me know what you think.

go/pkg/client/bytestream.go
```diff
@@ -73,6 +73,69 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
...
+		return 0, err
+	}
+	return c.writeChunkedWithOffset(ctx, name, ch, opts)
+}
+
+// writeChunkedWithOffset uploads chunked data with a given resource name to the CAS at an arbitrary offset.
+func (c *Client) writeChunkedWithOffset(ctx context.Context, name string, ch *chunker.Chunker, opts *ByteStreamWriteOpts) (int64, error) {
+	var totalBytes int64
+	var offset = opts.LastLogicalOffset
+	var initialOffset = opts.LastLogicalOffset
```
ramymedhat, line 90: Doesn't look like you're using the initialOffset in the first write?

ramymedhat (commented):
Can you clarify how you tested this? Also can you add unit tests?

go/pkg/client/bytestream.go
YukaSadaoka: No it's no being used. I created initialOffset to update offset after writing in line 127.

YukaSadaoka:
> Can you clarify how you tested this? Also can you add unit tests?
I used logstream integration and prober tests to test the changes. 
Acknowledged adding unit tests for this. I will add it and will update here once I'm done.

go/pkg/client/bytestream.go
ramymedhat: I'm confused. The chunker has chunked the entire blob, right? So its actually disregarding the initial offset? How will you make sure that you're uploading the right chunk relative to the desired offset?

go/pkg/client/client.go
```diff
@@ -474,6 +474,12 @@ func getRPCCreds(ctx context.Context, credFile string, useApplicationDefault boo
 	return rpcCreds, CredsFileAuth, nil
 }
 
+// ByteStreamWriteOpts sets options required for resumable ByteStream write operations.
+type ByteStreamWriteOpts struct {
```
mrahs, line 478: Have you considered using [variadic options](https://google.github.io/styleguide/go/best-practices#variadic-options)? It allows you to reuse the existing method without affecting existing call sites.

go/pkg/client/client.go
```diff
@@ -474,6 +474,12 @@ func getRPCCreds(ctx context.Context, credFile string, useApplicationDefault boo
...
+	FinishWrite       bool
```
mrahs, line 479: What is the use-case for this option?

It looks like the new method sends all the given bytes to the server and tells the server to start writing at a specific offset. Since all the given bytes are sent, `req.FinishWrite` should always be sent at the end. No?

YukaSadaoka: The boolean flag is for the case where clients didn't send all bytes in the previous write requests and want to resume writing the rest of bytes. Yes it's always sent but I wanted to allow clients to set the flag based on their remaining data so that they can resume writing.

go/pkg/client/client.go
YukaSadaoka: Thanks for the input. I have considered to use it but adding my changes to the current existing method(WriteBytes in bytestream.go) seems making readability not great. Would you recommend me to use variadic options over an option struct?

go/pkg/client/bytestream.go
YukaSadaoka: Right the current implementation is disregarding the initial offset(offset property in Chunker), I think I can update the code to use the offset property for a clarity. Let me try to update the code.

go/pkg/client/client.go
mrahs: That makes sense. Just noticed that the client associates the flag with the given resource name. Your implementation allows users to call the public method multiple times for the same resource name without causing errors, which makes sense with the new arbitrary offset option.

go/pkg/client/client.go
mrahs: The new method shares most of its code with the existing one, which seems like a good candidate for an option parameter. Variadic options allow backwards compatibility with existing call sites.

I'm thinking something like:
```go
type ByteStreamWriteOption interface {
  Apply(*ByteStreamWriteOpts)
}
type ByteStreamOptFinishWrite bool
func (o ByteSteramOptFinishWrite) Apply(opts *ByteStreamWriteOpts) {
  opts.FinishWrite = bool(o)
}

func (c *Client) WriteBytesWithOffset(ctx context.Context, name string, data []byte, rawOpts ...ByteStreamWriteOption) (int64, error) {
  opts := &ByteStreamWriteOpts{}
  for _, o := range rawOpts {
    o.Apply(opts)
  }
  ...
}
```

go/pkg/client/bytestream.go
```diff
@@ -73,6 +73,61 @@ func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chun
...
+func (c *Client) WriteBytesWithOffset(ctx context.Context, name string, data []byte, opts *ByteStreamWriteOpts) (int64, error) {
+	ue := uploadinfo.EntryFromBlob(data)
+	ch, err := chunker.New(ue, false, int(c.ChunkMaxSize))
+	if err != nil {
+		return 0, err
+	}
+	return c.writeChunkedWithOffset(ctx, name, ch, opts)
+}
+
+// writeChunkedWithOffset uploads chunked data with a given resource name to the CAS at an arbitrary offset.
```
mrahs, line 86: I think this comment is worth having for the exported method to expose it to IDEs and users.

go/pkg/client/client.go
YukaSadaoka: Update the code per your suggestion. Please let me know if I'm missing something.

go/pkg/client/bytestream.go
YukaSadaoka: Updated the exported method's comment.

go/pkg/client/client.go
mrahs: My bad, I meant to write the original public method name `WriteBytes` rather than add another version:
```go
func (c *Client) WriteBytes(ctx context.Context, name string, data []byte, rawOpts ...ByteStreamWriteOption) (int64, error) {
  ...
}

File: go/pkg/client/bytestream.go [hunk 6]
```diff
@@ -26,33 +48,56 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
...
+		return 0, err
+	}
+	return c.writeChunked(ctx, name, ch, opts...)
+}
+
 // writeChunked uploads chunked data with a given resource name to the CAS.
-func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
+func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker, rawOpts ...ByteStreamWriteOption) (int64, error) {
+	if len(rawOpts) == 1 {
+		return 0, errors.Errorf("Missing an arugment")
```
mrahs, line 65: Doesn't this break `WriteBytes` since it shares this internal method without the new options?

go/pkg/client/bytestream.go
```diff
@@ -26,33 +48,56 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
...
 			if err != nil {
 				return err
 			}
-			if chunk.Offset == 0 {
-				req.ResourceName = name
-			}
 			req.WriteOffset = chunk.Offset
+			// If LastLogicaloffset is provided as an option, we want uploading bytes
+			// at the last logcial offset + the chunk offset which is relative to Chunker passed in.
+			if resume {
```
mrahs, line 95: Is this condition necessary? The default value for `opts.LastLogicalOffset` is `0`, which means `chunk.Offset + opts.LastLogicalOffset` should give the correct offset in both cases.

go/pkg/client/bytestream.go
```diff
@@ -26,33 +48,56 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
...
+	return c.writeChunked(ctx, name, ch, opts...)
+}
+
 // writeChunked uploads chunked data with a given resource name to the CAS.
-func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker) (int64, error) {
+func (c *Client) writeChunked(ctx context.Context, name string, ch *chunker.Chunker, rawOpts ...ByteStreamWriteOption) (int64, error) {
+	if len(rawOpts) == 1 {
+		return 0, errors.Errorf("Missing an arugment")
+	}
+	opts := &ByteStreamWriteOpts{}
```
mrahs, line 67: The `FinishWrite` option should default to `true` to maintain current semantics. That is, we'll always finish at the end of the stream unless instructed otherwise.

go/pkg/client/bytestream.go
```diff
@@ -26,33 +48,56 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
...
+	}
+
 	var totalBytes int64
 	closure := func() error {
 		// Retry by starting the stream from the beginning.
 		if err := ch.Reset(); err != nil {
 			return errors.Wrap(err, "failed to Reset")
 		}
 		totalBytes = int64(0)
-		// TODO(olaola): implement resumable uploads.
```
mrahs, line 38: Hmm, this implementation is not fully resumable since retries will resend all the chunks. A true resumable implementation would need to track the offset across retries to avoid resending successfully sent chunks. If that is not in the scope of your change, you could simply clarify that in the doc comment (that the new option allows to upload to an arbitrary offset,  but retries still resend from the specified offset).

go/pkg/client/bytestream.go
YukaSadaoka: Resolved this with Anas offline. I removed this code block after refactoring.

go/pkg/client/bytestream.go
YukaSadaoka: Done.

go/pkg/client/bytestream.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+	"testing"
+
+	"google.golang.org/grpc"
+
+	bsgrpc "google.golang.org/genproto/googleapis/bytestream"
+	bspb "google.golang.org/genproto/googleapis/bytestream"
+)
+
+var (
+	logStreams    map[string]*logStream
```
mrahs, line 16: This can be a member of the `ByteStream` strucrt on

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+	logStreamData = []byte("Hello World! This is large data to send.")
```
mrahs, line 16: The value of this variable is constant throughout its lifetime. Might as well make it official by declaring as a constant.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+	logicalOffset int64
+	finalized     bool
+}
+
+func TestWriteBytesWithOffsetSuccess_LogStream(t *testing.T) {
+	tests := []struct {
+		description string
+		ls          *logStream
+		data        []byte
+		numIterate  int
```
mrahs, line 30: Nit: I think this could have a more descriptive name that makes it immediately obvious that 1) it is related to the field data and 2) it specifies how many parts that data will be split into. Maybe something like `dataPartsLen` or `dataPartsCount`?

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+		opts        []ByteStreamWriteOption
+		wantBytes   int64
```
mrahs, line 32: Nit: `wantBytes` sounds like a `[]byte` value. `wantBytesLen` might be a less ambiguous name.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+		},
+	}
+
+	b := setup(t)
+	defer b.shutDown()
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			size := len(test.data)/test.numIterate + 1
+			var start int
+			start = int(test.opts[0].(ByteStreamOptOffset))
```
mrahs, line 74: Nit: Those two lines can be joined into one using a short declaration: `start := int(...)`.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+		description string
+		ls          *logStream
+		data        []byte
+		numIterate  int
+		opts        []ByteStreamWriteOption
+		wantBytes   int64
+	}{
+		{
+			description: "valid data with offset 0",
+			ls:          &logStream{logStreamID: "logstream1", logicalOffset: 0},
```
mrahs, line 36: Nit: `logStreamID` can probably just reuse the `test.description`. That way, it's less likely that a future developer might forget to change the id to something unique after copying an existing test case. The test description is likely to be unique in the same test set (see my comment about encapsulating the streams map).

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+				writtenBytes, err := b.client.WriteBytesWithOptions(b.ctx, lsID, test.data[start:end], test.opts...)
+				if err != nil {
+					t.Errorf("WriteBytesWithOffset() failed unexpectedly: %v", err)
+				}
+				test.opts[0] = test.opts[0].(ByteStreamOptOffset) + ByteStreamOptOffset(writtenBytes)
+				start = end
+				end += size
+			}
+
+			if logStreams[lsID].logicalOffset != test.wantBytes {
```
mrahs: It's worth asserting a similar condition within the parts loop above. That is, after each call the stream's offset is incremented by the part's length.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+				t.Errorf("WriteBytesWithOffset() = %d, want %d", logStreams[lsID].logicalOffset, test.wantBytes)
+			}
+			if logStreams[lsID].finalized == false {
```
mrahs: Same here as well. It's worth asserting this condition within the parts loop.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+		})
+	}
+}
+
+func TestWriteBytesWithOffsetErrors_LogStream(t *testing.T) {
+	tests := []struct {
+		description string
+		ls          *logStream
+		data        []byte
+		notFound    bool
```
mrahs, line 112: I couldn't figure out how this field is used. If you're trying to test for a non-existing stream, I might suggest making the `ls` field of the corresponding test `nil`.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+			ChunkMaxSize(dataSize).Apply(b.client)
+
+			_, err := b.client.WriteBytesWithOptions(b.ctx, lsID, logStreamData[int(offset.(ByteStreamOptOffset)):], offset)
+			if err != nil {
+				t.Errorf("WriteBytesWithOffset() failed unexpectedly: %v", err)
+			}
+			if logStreams[lsID].logicalOffset != int64(dataSize) {
+				t.Errorf("WriteBytesWithOffset() = %d, want %d", logStreams[lsID].logicalOffset, int64(dataSize))
+			}
+			if logStreams[lsID].finalized != true {
```
mrahs: Nit: `!= true` is redundant.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+type Server struct {
+	client   *Client
+	listener net.Listener
+	server   *grpc.Server
+	fake     *Bytestream
+	ctx      context.Context
+}
+
+func setup(t *testing.T) *Server {
```
mrahs, line 235: `setup` is a generic name. The implementation seems to be doing `newServer` or `newFakeServer`.

go/pkg/client/retries_test.go
```diff
@@ -285,6 +285,36 @@ func TestWriteRetries(t *testing.T) {
 	}
 }
 
+func TestWriteRetriesWithOptions(t *testing.T) {
+	opts := [][]client.ByteStreamWriteOption{
+		{client.ByteSteramOptFinishWrite(true), client.ByteStreamOptOffset(0)},
+		{client.ByteSteramOptFinishWrite(false), client.ByteStreamOptOffset(0)},
+		{client.ByteSteramOptFinishWrite(true)},
+		{client.ByteSteramOptFinishWrite(false)},
+		{client.ByteStreamOptOffset(0)},
```
mrahs, line 294: Can you add a test case with a non-zero offset?

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done. Updated the logStreamID in TestWriteBytesWithOffsetSuccess_LogStream and TestWriteBytesWithOffsetErrors_LogStream.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/retries_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+				writtenBytes, err := b.client.WriteBytesWithOptions(b.ctx, lsID, test.data[start:end], test.opts...)
+				if err != nil {
+					t.Errorf("WriteBytesWithOffset() failed unexpectedly: %v", err)
+				}
+				if b.fake.logStreams[lsID].logicalOffset != int64(end) {
+					t.Errorf("WriteBytesWithOffset() = %d, want %d", b.fake.logStreams[lsID].logicalOffset, end)
+				}
+				// LogStream shouldn't be finalized when we set ByteSteramOptFinishWrite false.
+				if i != test.dataPartsLen-1 && b.fake.logStreams[lsID].finalized == true {
```
mrahs, line 94:
```suggestion
				if i != test.dataPartsLen-1 && b.fake.logStreams[lsID].finalized {
```

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+				test.opts[0] = test.opts[0].(ByteStreamOptOffset) + ByteStreamOptOffset(writtenBytes)
+				start = end
+				end += size
+			}
+
+			if b.fake.logStreams[lsID].logicalOffset != test.wantBytesLen {
+				t.Errorf("WriteBytesWithOffset() = %d, want %d", b.fake.logStreams[lsID].logicalOffset, test.wantBytesLen)
+			}
+			if b.fake.logStreams[lsID].finalized == false {
```
mrahs, line 106:
```suggestion
			if !b.fake.logStreams[lsID].finalized {
```

go/pkg/client/bytestream_test.go
mrahs: Right. I must have thought of it as a string constant. Not a big deal.

go/pkg/client/bytestream.go
```diff
@@ -26,33 +48,52 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
...
+			req := &bspb.WriteRequest{ResourceName: name}
 			chunk, err := ch.Next()
 			if err != nil {
 				return err
 			}
-			if chunk.Offset == 0 {
-				req.ResourceName = name
-			}
-			req.WriteOffset = chunk.Offset
+			req.WriteOffset = chunk.Offset + opts.LastLogicalOffset
```
ramymedhat, line 93: This would apply to all chunks?

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream.go
```diff
@@ -15,6 +15,28 @@ import (
...
 )
 
+// ByteStreamWriteOption sets options required for resumable upload.
+type ByteStreamWriteOption interface {
+	Apply(*ByteStreamWriteOpts)
+}
+
+// ByteSteramOptFinishWrite is a boolean flag to set FinishWrite
+// in ByteStream.WriteRequest.
+type ByteSteramOptFinishWrite bool
```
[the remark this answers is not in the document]
YukaSadaoka: Done.

go/pkg/client/bytestream.go
YukaSadaoka: Yes but LastLogicalOffset doesn't basically affect WriteOffset in the case where LastLogicalOffset is not passed in as ByteStreamWriteOption. If LastLogicalOffset is not passed in, the value of LastLogicalOffset will be 0 as default and it will be basically req.WriteOffset = chunk.Offset. 

Refactored this code block in this way so that it's easier to read and avoid adding many if statements.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+	}
+}
+
+func TestWirteBytesWithOptions_LogStream_Offset(t *testing.T) {
+	b := newServer(t)
+	defer b.shutDown()
+	dataSize := len(logStreamData)
+	for _, offset := range []ByteStreamWriteOption{ByteStreamOptOffset(0), ByteStreamOptOffset(6)} {
+		t.Run(fmt.Sprintf("ByteSteramOptFinishWrite=%t", offset), func(t *testing.T) {
```
mrahs, line 205: Typo
```suggestion
		t.Run(fmt.Sprintf("ByteStreamOptFinishWrite=%t", offset), func(t *testing.T) {
```

go/pkg/client/bytestream.go
ramymedhat: Ack

go/pkg/client/bytestream.go
```diff
@@ -26,33 +48,52 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 	return err
 }
 
+// WriteBytesWithOptions uploads a byte slice with a given resource name to the CAS
+// at an arbitrary offset but retries still resend from the specified offset.
+func (c *Client) WriteBytesWithOptions(ctx context.Context, name string, data []byte, opts ...ByteStreamWriteOption) (int64, error) {
```
ramymedhat, line 53: Can you just pass in the offset and finishWrite here? I think that would be simpler since its only two args.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,295 @@
...
+			wantBytesLen: 0,
+		},
+	}
+
+	b := newServer(t)
+	defer b.shutDown()
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			size := len(test.data)/test.dataPartsLen + 1
+			start := int(test.opts[0].(ByteStreamOptOffset))
```
ramymedhat, line 71: Just using function parameters rather than an opt struct will make this simpler as well.

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,295 @@
...
+				t.Error("WriteBytesWithOffset() got nil error, want non-nil error")
+			}
+			if writtenBytes != 0 {
+				t.Errorf("WriteBytesWithOffset() got %d byte(s), want 0 byte", writtenBytes)
+			}
+		})
+	}
+}
+
+func TestWirteBytesWithOptions_LogStream_FinishWrite(t *testing.T) {
```
ramymedhat, line 174: Write

go/pkg/client/bytestream_test.go
```diff
@@ -0,0 +1,295 @@
...
+			opts:        []ByteStreamWriteOption{ByteStreamOptOffset(0), ByteStreamOptFinishWrite(false)},
+		},
+		{
+			description: "not found",
+			ls:          nil,
+			data:        logStreamData,
+			opts:        []ByteStreamWriteOption{ByteStreamOptOffset(0), ByteStreamOptFinishWrite(false)},
+		},
+		{
+			description: "invlid smaller offset",
```
ramymedhat, line 133: s/invlid/invalid here and below

go/pkg/client/bytestream.go
```diff
@@ -26,33 +48,52 @@ func (c *Client) WriteBytes(ctx context.Context, name string, data []byte) error
 	return err
 }
 
+// WriteBytesWithOptions uploads a byte slice with a given resource name to the CAS
+// at an arbitrary offset but retries still resend from the specified offset.
```
ramymedhat, line 52: s/offset/remote offset?

go/pkg/client/bytestream.go
ramymedhat, line 53: s/WriteBytesWithOptions/WriteBytesAtRemoteOffset? I think that makes it clearer what this is doing.

YukaSadaoka: In case if we add more options to the method(or to writeChunked method) in the future and this method keeps a variadic option, we can add them easily without affecting the call sites of this method. I'm not sure how likely we will add more options but I think that scales better. What do you think?

go/pkg/client/bytestream_test.go
YukaSadaoka: Removed the unit test method as the implementation has changed a bit.

go/pkg/client/bytestream.go
YukaSadaoka: Resolved offline. Updated the method signatures.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream_test.go
YukaSadaoka: Done.

go/pkg/client/bytestream.go
YukaSadaoka: Done.

ramymedhat (approved):

PR #430, opened 2023-02-08
Refactor Entry struct to ensure immutability

mrahs:
The changes preserve semantics and provide strong immutability guarantees. This is stop-gap change for extending the Entry struct to support the implementation from pkg/cas.

### DIFF
--- a/go/pkg/chunker/chunker.go
+++ b/go/pkg/chunker/chunker.go
@@ -52,8 +52,7 @@ func New(ue *uploadinfo.Entry, compressed bool, chunkSize int) (*Chunker, error)
 	}
 	var c *Chunker
 	if ue.IsBlob() {
-		contents := make([]byte, len(ue.Contents))
-		copy(contents, ue.Contents)
+		contents := ue.Blob()
 		if compressed {
 			contents = fullCompressor.EncodeAll(contents, nil)
 		}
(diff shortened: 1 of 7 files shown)


go/pkg/chunker/chunker.go
```diff
@@ -77,7 +76,7 @@ func New(ue *uploadinfo.Entry, compressed bool, chunkSize int) (*Chunker, error)
 			chunkSize = IOBufferSize
 		}
 	} else {
-		return nil, errors.New("invalid Entry")
+		return nil, errors.New(fmt.Sprintf("invalid entry: BlobLength=%d, PathLength=%d", ue.BlobLen(), len(ue.Path())))
```
ramymedhat, line 79: What's the value of PathLength?

mrahs: This was based on a past iteration where I've attempted to get rid of the internal type field. I should change this line to print the entry type here.

go/pkg/uploadinfo/entry.go
```diff
@@ -8,37 +8,76 @@ import (
...
+	// A kindBlob entry can have a nil blob field and a kindFile entry can have an empty file path.
+	kind int
 }
 
-// IsBlob returns whether this Entry is for a blob in memory.
-func (ue *Entry) IsBlob() bool {
-	return ue.ueType == ueBlob
+// Kind is a debugging helper function that returns a string representation
+// of the entry's kind.
+func (e *Entry) Kind() string {
```
mrahs, line 28: On second thought, I'm going to remove this function because of Hyrum's law.

PR #432, opened 2023-02-22
[reader] Fix compressed reader retries

orn688:
**Note: this introduces a behavior change in addition to the bug fix described below. After `compressedSeeker.Read()` returns an error, subsequent calls continue to drain the internal buffer until it announces EOF, whereas before the method short-circuited and reported the last error on subsequent calls. Since this is an internal package, it's unlikely that there have been any external dependencies on this behavior.**

There was a subtle bug in the `compressedSeeker` class that caused retried upload requests to occasionally fail under a very specific set of conditions:

1. The file being uploaded must be greater than the client's `MaxBatchSize` so it's not batched together with other files (compression is always disabled for files that are batched together).
2. Upload compression must be enabled for the file, e.g. by setting `client.CompressedBytestreamThreshold(x)` where x is less than the file size. This will cause the upload to enter the `compressedSeeker` code path.

Here's the process by which the bug happens (the relevant code is mostly in `go/pkg/client/bytestream.go` in the `writeChunked()` function):
1. All chunks of the file upload normally up until the final chunk.
2. `ch.Next()` is called and retrieves the final chunk of the file.
3. The `reader.compressedSeeker` underlying the `chunker.Chunker` that was passed into `writeChunked()` gets an `io.EOF` error when reading from its underlying reader. This is completely normal and expected since the end of the file has been reached.
4. `reader.compressedSeeker.Read()` correctly propagates the `io.EOF` error but also caches it to its `cfs.err` field, so all subsequent calls to `Read()` will return `io.EOF`.
5. The `WriteRequest` to upload the final chunk of the file fails transiently, e.g. due to an internal server error.
6. The entire upload gets retried to do the `c.Retrier.Do()` call within `writeChunked()`.
7. `ch.Reset()` gets called to reset the chunker to the beginning of the file, calling `compressedSeeker.SeekOffset(0)` under the hood. This correctly seeks to the beginning of the file but *does not* unset the cached `err`.
7. The first call to `ch.Next()` tries to call `reader.compressedSeeker.Read()`, expecting to read from the beginning of the file again, but it immediately returns the cached `io.EOF` and a length of zero bytes, suggesting that the file is now empty.
8. `writeChunked()` dutifully tries to upload the apparently empty file, but under the digest of the file's actual contents.
9. This results in a server-side error when the server tries to validate that the uploaded file matches the digest under which it is being stored, something like: `Uploaded blob had calculated digest 19f3/0 but was being stored to 34ba/<size>`, where `<size>` is the actual size of the file in bytes. This is the same error that happens if the file is changed by another process between the first and the second upload attempt, but in this case it happens even though the file is not modified on disk. This error is not considered transient, so is not retried, and bubbles up to an error.

I could find no indication as to why errors needed to be cached on the `compressedSeeker`, and since the error caching was the underlying cause of the bug (I reproduce this in a test case), it was simplest to just remove the error caching logic.

### DIFF
--- a/go/pkg/reader/reader.go
+++ b/go/pkg/reader/reader.go
@@ -201,42 +201,26 @@ func (cfs *compressedSeeker) Read(p []byte) (int, error) {
 		return 0, errNotInitialized
 	}
 
-	var n int
-	var errR, errW error
-	for cfs.encdW != nil && cfs.buf.Len() < len(p) && errR == nil && errW == nil {
-		// Read is allowed to use the entity of p as a scratchpad.
-		n, errR = cfs.fs.Read(p)
-		// errW must be non-nil if written bytes != from n.
-		_, errW = cfs.encdW.Encoder.Write(p[:n])
-	}
-
-	// We have to treat EOF differently. It's the only "everything is going according
-	// to plan" error. EOF implies our uncompressed input data has finished reading,
-	// but it doesn't mean that we've finished preparing/returning our compressed data.
-	// We only want to return EOF to the user when the _compressed_ data is finished reading.
-	var finalErr error
-	if errR != nil && errR != io.EOF {
-		finalErr = errR
-	} else if errW != nil {
-		finalErr = errW
-	}
-
-	// When the buffer ends, or in case of _any_ error, we compress all the bytes we
-	// had available. The encoder requires a Close call to finish writing compressed
-	// data smaller than zstd's window size.
-	if finalErr != nil || errR == io.EOF {
-		if err := cfs.encdW.Encoder.Close(); finalErr == nil {
-			finalErr = err
+	var err error
+	if cfs.encdW != nil && cfs.buf.Len() < len(p) {
+		// Read is allowed to use the entirety of p as a scratchpad.
+		_, err = io.CopyBuffer(cfs.encdW.Encoder, cfs.fs, p)
+		// When the buffer ends, or in case of an error, we compress all the
+		// bytes we had available. The encoder requires a Close call to finish
+		// writing compressed data smaller than zstd's window size.
+		closeErr := cfs.encdW.Encoder.Close()
+		if err == nil {
+			err = closeErr
 		}
 		encoders.Put(cfs.encdW)
 		cfs.encdW = nil
 	}
 
-	m, err := cfs.buf.Read(p)
-	if finalErr == nil {
-		finalErr = err
+	n, readErr := cfs.buf.Read(p)
+	if err == nil {
+		err = readErr
 	}
-	return m, finalErr
+	return n, err
 }
 
 func (cfs *compressedSeeker) SeekOffset(offset int64) error {
--- a/go/pkg/reader/reader.go
+++ b/go/pkg/reader/reader.go
@@ -188,7 +188,7 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
 		return nil, errors.New("failed creating new encoder")
 	}
 
-	encdW.Encoder.Reset(sb)
+	encdW.Reset(sb)
 	return &compressedSeeker{
 		fs:    fs,
 		encdW: encdW,
@@ -202,14 +202,28 @@ func (cfs *compressedSeeker) Read(p []byte) (int, error) {
 	}
 
 	var err error
-	if cfs.encdW != nil && cfs.buf.Len() < len(p) {
-		// Read is allowed to use the entirety of p as a scratchpad.
-		_, err = io.CopyBuffer(cfs.encdW.Encoder, cfs.fs, p)
-		// When the buffer ends, or in case of an error, we compress all the
-		// bytes we had available. The encoder requires a Close call to finish
-		// writing compressed data smaller than zstd's window size.
-		closeErr := cfs.encdW.Encoder.Close()
+	// Repeatedly encode chunks of input data until we have enough compressed
+	// data to fill the output buffer. It can't be known ahead of time how much
+	// uncompressed data will correspond to a certain amount of compressed data,
+	// hence the need for a loop.
+	for cfs.encdW != nil && cfs.buf.Len() < len(p) && err == nil {
+		var n int
+		// Read is allowed to use the entity of p as a scratchpad.
+		n, err = cfs.fs.Read(p)
+		// errW must be non-nil if written bytes != n.
+		_, errW := cfs.encdW.Write(p[:n])
 		if err == nil {
+			err = errW
+		}
+	}
+
+	if err != nil {
+		// When the buffer ends (EOF), or in case of an unexpected error, we
+		// compress all the bytes we had available. The encoder requires a Close
+		// call to finish writing compressed data smaller than zstd's window
+		// size.
+		closeErr := cfs.encdW.Close()
+		if err == io.EOF {
 			err = closeErr
 		}
 		encoders.Put(cfs.encdW)
@@ -232,19 +246,15 @@ func (cfs *compressedSeeker) SeekOffset(offset int64) error {
 		if !ok || cfs.encdW == nil {
 			return errors.New("failed to get a new encoder")
 		}
-	} else if err := cfs.encdW.Encoder.Close(); err != nil {
+	} else if err := cfs.encdW.Close(); err != nil {
 		encoders.Put(cfs.encdW)
 		cfs.encdW = nil
 		return err
 	}
 
 	cfs.buf.Reset()
-	cfs.encdW.Encoder.Reset(cfs.buf)
-	if err := cfs.fs.SeekOffset(offset); err != nil {
-		return err
-	}
-
-	return nil
+	cfs.encdW.Reset(cfs.buf)
+	return cfs.fs.SeekOffset(offset)
 }
 
 func (cfs *compressedSeeker) IsInitialized() bool { return cfs.fs.IsInitialized() }
--- a/go/pkg/reader/reader.go
+++ b/go/pkg/reader/reader.go
@@ -13,6 +13,8 @@ import (
 	"github.com/mostynb/zstdpool-syncpool"
 )
 
+var errNotInitialized = errors.New("Not yet initialized")
+
 // Initializable is an interface containing methods to initialize a ReadSeeker.
 type Initializable interface {
 	IsInitialized() bool
@@ -64,7 +66,7 @@ func (fio *fileSeeker) Close() (err error) {
 // Read implements io.Reader.
 func (fio *fileSeeker) Read(p []byte) (int, error) {
 	if !fio.IsInitialized() {
-		return 0, errors.New("Not yet initialized")
+		return 0, errNotInitialized
 	}
 
 	return fio.reader.Read(p)
@@ -152,7 +154,6 @@ type compressedSeeker struct {
 	encdW *syncpool.EncoderWrapper
 	// This keeps the compressed data
 	buf *syncedBuffer
-	err error
 }
 
 var encoderInit sync.Once
@@ -187,13 +188,12 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
 		fs:    fs,
 		encdW: encdW,
 		buf:   sb,
-		err:   nil,
 	}, nil
 }
 
 func (cfs *compressedSeeker) Read(p []byte) (int, error) {
-	if cfs.err != nil {
-		return 0, cfs.err
+	if !cfs.IsInitialized() {
+		return 0, errNotInitialized
 	}
 
 	var n int
@@ -209,17 +209,18 @@ func (cfs *compressedSeeker) Read(p []byte) (int, error) {
 	// to plan" error. EOF implies our uncompressed input data has finished reading,
 	// but it doesn't mean that we've finished preparing/returning our compressed data.
 	// We only want to return EOF to the user when the _compressed_ data is finished reading.
+	var finalErr error
 	if errR != nil && errR != io.EOF {
-		cfs.err = errR
+		finalErr = errR
 	} else if errW != nil {
-		cfs.err = errW
+		finalErr = errW
 	}
 
 	// When the buffer ends, or in case of _any_ error, we compress all the bytes we
 	// had available. The encoder requires a Close call to finish writing compressed
 	// data smaller than zstd's window size.
 	var errC error
-	if cfs.err != nil || errR == io.EOF {
+	if finalErr != nil || errR == io.EOF {
 		errC = cfs.encdW.Encoder.Close()
 		encoders.Put(cfs.encdW)
 		cfs.encdW = nil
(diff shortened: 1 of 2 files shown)


orn688:
For context on user impact, see https://crbug.com/1279920.

go/pkg/reader/reader.go
```diff
@@ -187,13 +188,12 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
...
 }
 
 func (cfs *compressedSeeker) Read(p []byte) (int, error) {
-	if cfs.err != nil {
-		return 0, cfs.err
+	if !cfs.IsInitialized() {
+		return 0, errNotInitialized
 	}
 
 	var n int
```
tamird, line 199: (I can't seem to anchor it below)

The check for `errR == nil && errW == nil` on line 201 is always true.

go/pkg/reader/reader.go
```diff
@@ -209,33 +209,34 @@ func (cfs *compressedSeeker) Read(p []byte) (int, error) {
...
+		finalErr = errW
 	}
 
 	// When the buffer ends, or in case of _any_ error, we compress all the bytes we
 	// had available. The encoder requires a Close call to finish writing compressed
 	// data smaller than zstd's window size.
 	var errC error
-	if cfs.err != nil || errR == io.EOF {
+	if finalErr != nil || errR == io.EOF {
 		errC = cfs.encdW.Encoder.Close()
```
tamird, line 224: why bother with errC at all? can this just assign directly to finalErr?

go/pkg/reader/reader.go
```diff
@@ -209,33 +209,34 @@ func (cfs *compressedSeeker) Read(p []byte) (int, error) {
...
 		encoders.Put(cfs.encdW)
 		cfs.encdW = nil
 	}
 
 	m, errR2 := cfs.buf.Read(p)
 
-	if cfs.err == nil {
+	if finalErr == nil {
 		if errC != nil {
```
tamird, line 232: If you remove `errC` above, I think all the conditionals here can go away and you can write:
```
m, errR := cfs.buf.Read(p) // no need to name it errR2 since we're done with errR by now
if finalErr == nil {
  finalErr = errR
}
return m, finalErr
```

go/pkg/reader/reader.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/mostynb/zstdpool-syncpool"
 )
 
+var errNotInitialized = errors.New("Not yet initialized")
```
tamird, line 16: +1 to golint complaining about the capitalization here

go/pkg/reader/reader_test.go
```diff
@@ -165,28 +165,41 @@ func TestCompressedReader(t *testing.T) {
...
-			fullData := make([]byte, len(blob)+100)
-			dataBuf := make([]byte, 5)
-			var n, m int
-			for err = nil; err == nil; {
-				m, err = r.Read(dataBuf)
-				copy(fullData[n:], dataBuf)
-				n += m
+			got, err := io.ReadAll(r)
+			if err != nil {
+				t.Errorf("Read() returned error: %v", err)
```
tamird, line 181: should this be fatal?

go/pkg/reader/reader_test.go
```diff
@@ -165,28 +165,41 @@ func TestCompressedReader(t *testing.T) {
...
 			}
-			if err != io.EOF {
-				t.Errorf("Expected EOF, got nil")
+			if err := r.Initialize(); err != nil {
+				t.Fatalf("Failed to initialize reader: %v", err)
+			}
+
+			got, err = io.ReadAll(r)
+			if err != nil {
+				t.Errorf("Read() returned error: %v", err)
```
tamird, line 199: ditto

go/pkg/reader/reader_test.go
```diff
@@ -165,28 +165,41 @@ func TestCompressedReader(t *testing.T) {
 
 			r, err := NewCompressedFileSeeker(path, 10)
 			if err != nil {
-				t.Fatalf("Failed to initialize compressor reader: %v", err)
+				t.Fatalf("Failed to create compressor reader: %v", err)
 			}
+
+			if _, err := r.Read(make([]byte, 5)); err == nil {
```
tamird, line 171: this test is in the same package as the implementation; i think you can make a stronger assertion here that the error is `errNotInitialized`.

go/pkg/reader/reader.go
```diff
@@ -209,33 +209,34 @@ func (cfs *compressedSeeker) Read(p []byte) (int, error) {
...
 
 	// When the buffer ends, or in case of _any_ error, we compress all the bytes we
 	// had available. The encoder requires a Close call to finish writing compressed
 	// data smaller than zstd's window size.
 	var errC error
-	if cfs.err != nil || errR == io.EOF {
+	if finalErr != nil || errR == io.EOF {
 		errC = cfs.encdW.Encoder.Close()
 		encoders.Put(cfs.encdW)
 		cfs.encdW = nil
```
tamird, line 226: does the test exercise this code path? on the second time through it seems `cfs.encdW` will be nil and the call on line 223 will NPE?

before this change, the error caching logic prevented this block from executing a second time, but now it can.

go/pkg/reader/reader.go
orn688: Are you  sure? It's in the for loop condition, so in theory `errR` or `errW` could become non-nil during one iteration and this condition would trigger at the next iteration. Maybe it wouldn't happen in practice if the only possible error is an `io.EOF`, in which case `cfs.buf.Len() < len(p)` would be false and terminate the loop, but even in that case I think it's good to include this check in the loop condition for completeness.

tamird: Ha, I completely missed that this was `for` rather than `if`.

go/pkg/reader/reader.go
orn688: That would technically be an API change since the capitalization was present in the original code, so I'd rather not change it as part of this bug fix.

tamird: Sounds reasonable. This complaint comes from golint-ci, could you add a `//nolint:golint` inline comment per [this article](https://golangci-lint.run/usage/false-positives/#nolint-directive)?

go/pkg/reader/reader.go
orn688: Good question, I think it's actually fine because `cfs.encdW != nil` in the `for` condition will evaluate to false and cause the loop to not run at all, so `finalErr` and `errR` will both be nil when we get here and the `if` won't be entered. The `cfs.buf.Read()` call after this should then correctly return `io.EOF`.

I added an assertion to cover this code path by calling `Read()` immediately after the `io.ReadAll()` and checking that it returns `io.EOF`.

It took a bit of inspection for me to figure that out though, so I'm open to suggestions for making the code more obviously correct.

go/pkg/reader/reader_test.go
orn688: Done.

go/pkg/reader/reader_test.go
orn688: Done.

go/pkg/reader/reader_test.go
orn688: Good catch, done.

go/pkg/reader/reader.go
orn688: Done.

go/pkg/reader/reader.go
orn688: Done.

go/pkg/reader/reader.go
orn688: Done.

go/pkg/reader/reader.go
```diff
@@ -187,13 +193,12 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
...
 }
 
 func (cfs *compressedSeeker) Read(p []byte) (int, error) {
-	if cfs.err != nil {
-		return 0, cfs.err
+	if !cfs.IsInitialized() {
+		return 0, errNotInitialized
 	}
 
 	var n int
```
tamird, line 204: Looking at the code block that follows, can we simplify the error handling using [`io.CopyBuffer`](https://pkg.go.dev/io#CopyBuffer)? unlike the current code, `io.CopyBuffer` (and `io.Copy`) internally handles `io.EOF`, so the code below gets simpler as well. I think that means you can write:

```
var err error
if cfs.encdW != nil && cfs.buf.Len() < len(p) { // <-- `for` became `if` because `io.CopyBuffer` loops internally.
  _, err1 := io.CopyBuffer(cfs.encdW.Encoder, cfs.fs)
  err2 := cfs.encdW.Encoder.Close()
  encoders.Put(cfs.encdW)
  cfs.encdW = nil
  if err1 != nil {
    err = err1
  } else if err2 != nil {
    err = err2
  }
}
n, err3 := cfs.buf.Read(p)
if err == nil {
  err = err3
}
return n, err
```

tamird: Oops, that should be `io.CopyBuffer(cfs.encdW.Encoder, cfs.fs, p)`.

go/pkg/reader/reader.go
```diff
@@ -187,55 +193,34 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
...
-	// data smaller than zstd's window size.
-	var errC error
-	if cfs.err != nil || errR == io.EOF {
-		errC = cfs.encdW.Encoder.Close()
+	if !cfs.IsInitialized() {
+		return 0, errNotInitialized
+	}
+
+	var err error
+	if cfs.encdW != nil && cfs.buf.Len() < len(p) {
```
tamird, line 205: Do you know why the check `cfs.buf.Len() < len(p)` matters?

tamird (approved):

go/pkg/reader/reader.go
orn688: I believe it's an optimization – if there's already enough cached compressed data available to fill `p`, then we can skip compressing more data and just use the already-compressed data.

I think that only happens when the compressed data is *longer* than the noncompressed data (since that's the only time that a previous `Read()` call might leave some data in the buffer). A comment in the test file says that can happen if the size of the data is very small due to the addition of headers.

tamird: Ah, I see. Seems like this implementation is not as efficient as it could be due to this staging. Oh well.

tamird (approved):

go/pkg/reader/reader.go
```diff
@@ -187,55 +193,34 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
...
+		// Read is allowed to use the entirety of p as a scratchpad.
+		_, err = io.CopyBuffer(cfs.encdW.Encoder, cfs.fs, p)
```
mrahs, line 207: It seems that this line would copy the entire content of the reader into the underlying buffer of the encoder (the provided buffer is used for staging, but does not limit the number of bytes copied). However, only up to `len(p)` is actually returned to the caller. A subsequent call to the method will simply continue to fill the specified buffer from the cached content of the first call. This could lead to OOM errors. [`io.CopyN`](https://pkg.go.dev/io#CopyN) might be a more suitable API here.

go/pkg/reader/reader_test.go
```diff
@@ -69,8 +69,8 @@ func TestFileReaderSeeks(t *testing.T) {
 			}
 
 			r.SeekOffset(tc.seekOffset)
-			if _, err := r.Read(data); err == nil {
-				t.Errorf("Read() = should have err'd on unitialized reader")
+			if _, err := r.Read(data); err != errNotInitialized {
+				t.Errorf("Read() should have produced %q, got %q", errNotInitialized, err)
```
mrahs, line 73: Nit: 
```suggestion
				t.Errorf("Read() should have failed with %q, got %q", errNotInitialized, err)
```

go/pkg/reader/reader_test.go
```diff
@@ -165,28 +165,51 @@ func TestCompressedReader(t *testing.T) {
 
 			r, err := NewCompressedFileSeeker(path, 10)
 			if err != nil {
-				t.Fatalf("Failed to initialize compressor reader: %v", err)
+				t.Fatalf("Failed to create compressor reader: %v", err)
 			}
+
+			if _, err := io.ReadAll(r); err != errNotInitialized {
+				t.Errorf("Read() should have produced %q, got %q", errNotInitialized, err)
```
mrahs, line 172: Nit:
```suggestion
				t.Errorf("Read() should have failed with %q, got %q", errNotInitialized, err)
```

mrahs (changes requested):
Thank you for taking the time to improve this package. That's a great PR description for such a subtle bug!

go/pkg/reader/reader.go
tamird: `io.CopyN` doesn't swallow `io.EOF` the same way that `io.Copy` and `io.CopyBuffer` do. It also doesn't take a buffer to use as scratch like `io.CopyBuffer` does. That said, the implementation is pretty simple.

Could we try to adapt this to use `io.CopyBuffer` + `io.LimitReader`?

```
_, err = io.CopyBuffer(cfs.encdW.Encoder, io.LimitReader(cfs.fs, len(p)), p)
```

go/pkg/reader/reader.go
```diff
@@ -182,60 +188,53 @@ func NewCompressedSeeker(fs ReadSeeker) (ReadSeeker, error) {
...
 
-	var n int
-	var errR, errW error
-	for cfs.encdW != nil && cfs.buf.Len() < len(p) && errR == nil && errW == nil {
+	var err error
+	// Repeatedly encode chunks of input data until we have enough compressed
+	// data to fill the output buffer. It can't be known ahead of time how much
+	// uncompressed data will correspond to a certain amount of compressed data,
+	// hence the need for a loop.
+	for cfs.encdW != nil && cfs.buf.Len() < len(p) && err == nil {
```
mrahs, line 209: Nit: It would be great if you can reorder the conditionals as:
```suggestion
	for cfs.buf.Len() < len(p) && err == nil && cfs.encdW != nil {
```
That way, the functional conditional is read first, that is to say the loop is for filling up `cfs.buf` by at least `len(p)`.
It would also be great if you added comments to explain the other two conditionals: `err != nil` means the _loop_ hasn't encountered any errors and `encdW != nil` means the _reader_ hasn't encountered any errors (including EOF).

mrahs (approved):
LGTM. It's worth noting that this PR changes the behavior such that once the method has errored out, subsequent calls continue to drain the internal buffer until it announces EOF, whereas before the method short-circuited and reported the last error. Clients that are sensitive to this change might break. I think it's acceptable, but worth noting in the PR description for posterity.

go/pkg/reader/reader.go
tamird: I think that expression was wrong. It should be
```
_, err = io.CopyBuffer(cfs.encdW.Encoder, io.LimitReader(cfs.fs, len(p) - cfs.buf.Len()), p)
```

go/pkg/reader/reader.go
orn688: Thanks for the suggestion.

go/pkg/reader/reader.go
orn688: I still can't get `CopyBuffer` working even with `LimitReader`. I could continue digging into it, but at this point I'd prefer to submit the PR as-is, especially because I'd rather limit the scope of the refactoring that's bundled with the bug fix.

orn688:
> LGTM. It's worth noting that this PR changes the behavior such that once the method has errored out, subsequent calls continue to drain the internal buffer until it announces EOF, whereas before the method short-circuited and reported the last error. Clients that are sensitive to this change might break. I think it's acceptable, but worth noting in the PR description for posterity.

Thanks, I added a note to the description.

PR #434, opened 2023-04-03
New config file for pluggable CAS implementation

mrahs:
These configuration values will be used by the implementation which will posted later.

Reusing existing configuration structs risks overloading them or changing their semantics.

### DIFF
--- a/go/pkg/cas/config.go
+++ b/go/pkg/cas/config.go
@@ -1,0 +1,249 @@
+package cas
+
+import (
+	"errors"
+	"math"
+	"sync"
+	"time"
+
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
+)
+
+var (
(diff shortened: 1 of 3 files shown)


go/pkg/cas/config.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// DefaultOpenFilesLimit is based on GCS recommendations.
+	// See: https://cloud.google.com/compute/docs/disks/optimizing-pd-performance#io-queue-depth
+	DefaultOpenFilesLimit = 32
+
+	// DefaultOpenLargeFilesLimit is arbitrarily set.
+	DefaultOpenLargeFilesLimit = 2
+
+	// DefaultCompressionSizeThreshold is disabled by default.
+	DefaultCompressionSizeThreshold = -1
```
mrahs, line 46: This should be set to `math.MaxInt64` as per the documentation below.

bozydarsz (approved):

PR #438, opened 2023-04-13
introduce immutable path package

mrahs:
This implementation helps eliminate all ambiguity stemming from using strings as path, namely differentiating between absolute and relative paths.

### DIFF
--- a/go/pkg/errors/errors.go
+++ b/go/pkg/errors/errors.go
@@ -1,0 +1,81 @@
+// Package errors provides the ability to wrap multiple errors while maintaining API compatibility with the standard package.
+// This functionality is a drop-in replacement from go1.20. Remove this package once the SDK is updated to g1.20.
+package errors
+
+import "errors"
+
+type joinError struct {
+	errs []error
+}
+
+// Error joins the texts of the wrapped errors using newline as the delimiter.
+func (e *joinError) Error() string {
+	var b []byte
+	for i, err := range e.errs {
+		if i > 0 {
+			b = append(b, '\n')
+		}
+		b = append(b, err.Error()...)
+	}
+	return string(b)
+}
+
+// Join wraps the specified errors into one error that prints the entire list using newline as the delimiter.
+func Join(errs ...error) error {
+	n := 0
+	for _, err := range errs {
+		if err != nil {
+			n++
+		}
+	}
+
+	if n == 0 {
+		return nil
+	}
+
+	// Unlike go1.20, this allows for efficient and convenient wrapping of errors without additional guards.
+	// E.g. the following code returns err as is without any allocations if errClose is nil:
+	// errClose := f.Close(); err = errors.Join(errClose, err)
+	if n == 1 {
+		for _, err := range errs {
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	e := &joinError{
+		errs: make([]error, 0, n),
+	}
+	for _, err := range errs {
+		if err != nil {
+			e.errs = append(e.errs, err)
+		}
+	}
+	return e
+}
+
+// Is implements the corresponding interface allowing the joinError type to be compatible with
+// errors.Is(error) bool of the standard package without having to override it.
+func (e *joinError) Is(target error) bool {
+	if target == nil || e == nil {
+		return e == target
+	}
+
+	for _, e := range e.errs {
+		if errors.Is(e, target) {
+			return true
+		}
+	}
+	return false
+}
+
+// Is delegates to the standard package.
+func Is(err, target error) bool {
+	return errors.Is(err, target)
+}
+
+// New delegates to the standard package.
+func New(text string) error {
+	return errors.New(text)
+}
--- a/go/pkg/io/impath/impath.go
+++ b/go/pkg/io/impath/impath.go
@@ -1,0 +1,155 @@
+// Package impath (immutable path) provides immutable and distinguishable types for absolute and relative paths.
+// This allows for strong guarantees at compile time, improves legibility and reduces maintenance burden.
+package impath
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/errors"
+)
+
+var (
+	ErrNotAbsolute   = errors.New("path is not absolute")
+	ErrNotRelative   = errors.New("path is not relative")
+	ErrNotDescendant = errors.New("target is not a descendant of base")
+	Root             = os.Getenv("SYSTEMDRIVE") + string(os.PathSeparator)
+)
+
+// Absolute represents an immutable absolute path.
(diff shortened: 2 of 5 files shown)


go/pkg/io/impath/impath.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return p
+}
+
+// JoinAbs is a convenient method to join multiple paths into an absolute path.
+//
+// If the specified arguments do not join to a valid absolute path, ErrNotAbsolute is returned.
+// This happens if the parent is a zero path (not created using a function from this package).
+// If error is not nil, the returned value is not valid.
+func JoinAbs(parent Abs, parts ...Rel) (Abs, error) {
```
bozydarsz, line 112: The name suggests joining Abs might be misleading. Might be clearer to name it `AppendToAbs` or even make it a member function (called sth like `Append`) of of `Abs` type.

go/pkg/io/impath/impath.go
```diff
@@ -0,0 +1,152 @@
...
+	ps := make([]string, len(parts)+1)
+	ps[0] = parent.String()
+	for i, p := range parts {
+		ps[i+1] = p.String()
+	}
+	p := filepath.Join(ps...)
```
bozydarsz, line 118: lines 118-123 can be replaced with `return ToAbs(ps)`

go/pkg/io/impath/impath.go
```diff
@@ -0,0 +1,152 @@
...
+
+var (
+	zeroAbs = Abs{}
+	zeroRel = Rel{}
+)
+
+// ToAbs creates a new absolute and clean path from the specified parts.
+//
+// If the specified parts do not join to a valid absolute path, ErrNotAbsolute is returned.
+// If error is not nil, the returned value is not valid.
```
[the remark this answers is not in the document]
mrahs: The value is usually an interface or a pointer that is conventionally nil when an error is returned. Since this function returns a concrete value, I thought it doesn't hurt to reiterate that it should not be used.

go/pkg/io/impath/impath.go
```diff
@@ -0,0 +1,152 @@
...
+func ToAbs(pathParts ...string) (Abs, error) {
```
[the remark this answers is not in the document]
mrahs: `New` usually returns a pointer or an interface type. This doesn't, which is why I wanted it to be clear from the name. Another alternative I considered is `FromString`, but this is more concise.

go/pkg/io/impath/impath.go
```diff
@@ -0,0 +1,152 @@
...
+	if filepath.IsAbs(p) {
+		return Abs{path: path(p)}, nil
+	}
+	return zeroAbs, ErrNotAbsolute
+}
+
+// MustAbs is a convenient wrapper of ToAbs that is useful for constant paths.
+//
+// If the specified parts do not join to a valid absolute path, it panics.
+func MustAbs(pathParts ...string) Abs {
```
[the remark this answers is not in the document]
mrahs: It's convenient for tests, including from outside this package, where paths are constants and known to be valid.

go/pkg/io/impath/impath.go
mrahs: Good point. I have eyed this name multiple times. `Join` does suggest that all arguments are of the same type, which is not ideal in this case. I was hesitant to define it as a member since it doesn't depend on any state. I made an exception above for `Dir` and `Base` to maintain consistent naming between the two types. I guess it makes sense to make the same exception here where `Append` would have the same argument list for both types.

mrahs: Replaced this with an `Append` method that has identical signature for both types.

go/pkg/errors/errors.go
```diff
@@ -0,0 +1,79 @@
+// Package errors provides the ability to wrap multiple errors while maintaining API compatibility with the standard package.
```
bozydarsz, line 1: what's the added value over standard error wrapping with %w?

for instance, instead of ` errors.Join(ErrNotAbsolute, fmt.Errorf("path %q", p))` you could use standard `fmt.Errorf("path %q: %w", p, ErrNotAbsolute)` and [then use errors.Is](https://pkg.go.dev/errors#Is)

go/pkg/errors/errors.go
```diff
@@ -0,0 +1,79 @@
...
+package errors
+
+import "errors"
+
+type joinError struct {
+	errs []error
+}
+
+func (e *joinError) Error() string {
```
bozydarsz, line 10: missing comment

go/pkg/io/impath/impath.go
```diff
@@ -0,0 +1,152 @@
...
+// when joined together with the base using filepath.Join(base, path), the result
+// is lexically equivalent to the specified target path.
+//
+// The returned error is nil, ErrNotRelative, or ErrNotDescendant.
+func Descendant(base Absolute, target Absolute) (Relative, error) {
+	p, err := filepath.Rel(base.String(), target.String())
+	if err != nil {
+		// Since the zero Absolute path is the root, this error may only happen if base and target
+		// have different roots, which is possible on Windows.
+		return zeroRel, errors.Join(ErrNotRelative, err)
```
bozydarsz, line 146: why the error type is `ErrNotRelative` and not `ErrNotDescendant`. If I got the comment right, this can happen if both are absolute paths but belong to different system drives.

go/pkg/errors/errors.go
mrahs: I'm wrapping 3+ errors in the uploader implementation and would like to keep it consistent throughout the SDK. There are code blocks in the SDK where only the last error is captured while intermediate ones are swallowed which is not great.

go/pkg/io/impath/impath.go
mrahs: At this point the function has only tested for a relative path, which is a superset of a descendant path. Returning that error conveys more information to the caller. I.e. a non-descendant path may still be relative, but a non-relative path cannot be descendant.

go/pkg/errors/errors.go
bozydarsz: Understand. Do you think you could add a TODO to remove errors package once we're migrated to go 1.20

go/pkg/io/impath/impath.go
bozydarsz: Oh, so you mean here that path `target` cannot be made relative to `base`, is that right? If yes, can you provide the explanation in the function's comment? The error's message mentions `path is not relative` and it's easy to for the users to assume to it's due to arguments not being relative.

bozydarsz (approved):
added some minor comments, but otherwise lgtm

go/pkg/io/impath/impath.go
mrahs: Will do. The returned error from `filepath.Rel` includes that information. The combined errors would lool like:
```
path is not relative
Rel: can't make baz relative to /foo/bar
```

PR #439, opened 2023-04-18
Avoid delays in unified uploads due to batching of FindMissingBlobs.

ramymedhat:
This PR removes the behavior to batch FindMissingBlobs calls as part of unified uploads. This behavior has been benchmarked to slow builds down by ~2x. FindMissingBlobs is cheap and calling it per UploadIfMissing call is a reasonable way to prune upload requests going into a buffer, specially that the blob cache hit rate is expected to be very high.

This PR also fixes the handling of UnifiedUploads / UnifiedDownloads client options to avoid complicated restart logic.

### DIFF
--- a/go/pkg/client/cas_upload.go
+++ b/go/pkg/client/cas_upload.go
@@ -246,9 +246,28 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 		return nil, 0, err
 	}
 	wait := make(chan *uploadResponse, uploads)
-	var missing []digest.Digest
+	var dgs []digest.Digest
+	dedupDgs := make(map[digest.Digest]bool, uploads)
+	for _, ue := range entries {
+		if _, ok := dedupDgs[ue.Digest]; !ok {
+			dgs = append(dgs, ue.Digest)
+			dedupDgs[ue.Digest] = true
+		}
+	}
+	missing, err := c.MissingBlobs(ctx, dgs)
+	if err != nil {
+		return nil, 0, err
+	}
+	missingDgs := make(map[digest.Digest]bool)
+	for _, dg := range missing {
+		missingDgs[dg] = true
+	}
 	var reqs []*uploadRequest
 	for _, ue := range entries {
+		if _, ok := missingDgs[ue.Digest]; !ok {
+			uploads--
+			continue
+		}
 		if ue.Digest.IsEmpty() {
 			uploads--
 			LogContextInfof(ctx, log.Level(2), "Skipping upload of empty entry %s", ue.Digest)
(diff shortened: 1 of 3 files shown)


go/pkg/client/cas_upload.go
```diff
@@ -246,9 +246,28 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
...
+		return nil, 0, err
+	}
+	missingDgs := make(map[digest.Digest]bool)
+	for _, dg := range missing {
+		missingDgs[dg] = true
+	}
 	var reqs []*uploadRequest
 	for _, ue := range entries {
+		if _, ok := missingDgs[ue.Digest]; !ok {
+			uploads--
```
mrahs, line 268: It looks like this counter is redundant. The receiver loop below can be guarded using `len(reqs)` since that is effectively the number of sent requests and consequently the number of expected responses.

go/pkg/client/cas_upload.go
```diff
@@ -246,9 +246,28 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
...
 	wait := make(chan *uploadResponse, uploads)
-	var missing []digest.Digest
+	var dgs []digest.Digest
+	dedupDgs := make(map[digest.Digest]bool, uploads)
+	for _, ue := range entries {
+		if _, ok := dedupDgs[ue.Digest]; !ok {
+			dgs = append(dgs, ue.Digest)
+			dedupDgs[ue.Digest] = true
+		}
+	}
```
mrahs, line 256: Is this deduplication necessary? It used to be so when it happened after aggregating digests from multiple calls, but since this is now here, it doesn't seem likely to be duplicates from a single caller. Even if there was, checking for non-missing within the upload loop below renders this loop even less useful.

go/pkg/client/cas_upload.go
ramymedhat: Good catch. Done.

mrahs (approved):

go/pkg/client/cas_upload.go
```diff
@@ -280,13 +297,12 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 				return nil, 0, resp.err
 			}
 			if resp.missing {
```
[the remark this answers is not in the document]
ramymedhat: This should be preserved because updateAndNotify will only notify the first client that the blob was missing to avoid double counting. https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/client/cas_upload.go#L614.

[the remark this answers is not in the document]
mrahs: So this flag is also used to mark a cache hit/miss. Got it.

PR #440, opened 2023-04-24
introduce the file system walker package

mrahs:
The package provides an API to walk the file system in depth-first style traversal. This encapsulates IO and implementation naunces and exposes a concise interface for clients.

The DFS traversal and the deferring behaviour is leveraged in the CAS upload implementation that will posted later.

### DIFF
--- a/go/pkg/io/walker/stack.go
+++ b/go/pkg/io/walker/stack.go
@@ -1,0 +1,49 @@
+package walker
+
+import "fmt"
+
+type stack struct {
+	stack []any
+}
+
+func (s *stack) push(items ...any) {
+	s.stack = append(s.stack, items...)
+}
+
+func (s *stack) insert(item any, i int) {
+	if i < 0 || i > len(s.stack) {
+		panic(fmt.Errorf("stack index %d out of range with stack length %d", i, len(s.stack)))
+	}
+
+	if i == len(s.stack) {
+		s.stack = append(s.stack, item)
+		return
+	}
+
+	// Duplicate item at index i and shift subsequent items right.
+	s.stack = append(s.stack[:i+1], s.stack[i:]...)
+	// Replace the duplicated item with the specified one.
+	s.stack[i] = item
+}
+
+func (s *stack) pop() any {
+	p := s.peek()
+	if p == nil {
+		return nil
+	}
+	s.stack = s.stack[:len(s.stack)-1]
+	return p
+}
+
+func (s *stack) peek() any {
+	i := len(s.stack) - 1
+	if i < 0 {
+		return nil
+	}
+	p := s.stack[i]
+	return p
+}
+
+func (s *stack) len() int {
+	return len(s.stack)
+}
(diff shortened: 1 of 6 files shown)


go/pkg/io/walker/stack.go
```diff
@@ -0,0 +1,49 @@
...
+	// Replace the duplicated item with the specified one.
+	s.stack[i] = item
+}
+
+func (s *stack) pop() any {
+	p := s.peek()
+	if p == nil {
+		return nil
+	}
+	s.stack = s.stack[:len(s.stack)-1]
```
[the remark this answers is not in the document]
bozydarsz: Good to know, shouldn't be a concern then. I'll have better understanding on large the stack can grow during the DepthFirst execution, but I don't think having `O(len(s.stack))` worst case is a problem as far as it's `O(1)` on average (which I believe it is).

go/pkg/io/walker/walker.go
```diff
@@ -0,0 +1,327 @@
...
+			}
+			return deferred, next, nil
+		}
+
+		// Iterate before checking for EOF to avoid missing the last batch.
+		for _, name := range names {
+			// name is relative.
+			relName := impath.MustRel(name)
+			rp := e.realPath.Append(relName)
+			dp := e.desiredPath.Append(relName)
```
bozydarsz, line 276: if desiredPath is nil it will fail

go/pkg/io/walker/walker.go
```diff
@@ -0,0 +1,327 @@
...
+// Replace is only valid for symlinks.
+// An unexpected NextStep value cancels the entire walk.
+//
+// If a symlink is encountered, the callback must return Skip to avoid triggering a recursive walk on the symlink target.
+// Alternatively, returning Replace will instruct the walker to traverse the target as if its path is that of the symlink itself.
+//
+// At any point, the client may return Cancel to cancel the entire walk.
+//
+// The error returned by the walk is either nil or ErrBadNextStep.
+func DepthFirst(root impath.Absolute, exclude Filter, fn WalkFunc) error {
```
bozydarsz, line 78: Does the future use of the walker justify requiring to provide WalkFunc as a parameter instead of defining its behavior within the walker package?

I didn't see the code that uses the walker yet, but looking at the current implementation of visit and processDir I've noticed lots of extra code that handles errors caused by invalid results returned by WalkFunc. Likewise, because of the fact that WalkFunc is called multiple times for the same path and due to the fact that WalkFunc needs to handle various errors, it seems that implementation of the WalkFunc is not trivial and my require deep knowledge of DeepFirst internals to make it right, making the usage of the package cumbersome. 

If it's possible, maybe it'd be simpler from the perspective the caller, but also from the perspective of the walker package to define, if not all, at least some of these behaviors within the package instead of deferring decisions to WalkFunc

mrahs: I have revised the implementation to eliminate runtime errors that can be captured at compile time. I like it better now. Thanks for the comment!

bozydarsz (approved):

PR #441, opened 2023-04-26
Handle symlinks in packageDirectories

peterebden:
Very similar to the handling of `t.files`, seems to have just been missed in this case.


mrahs (approved):
Thanks for the PR, Peter.

PR #442, opened 2023-04-26
Add option for determining compression of blob uploads

peterebden:
The current implementation determines compression solely on blob size, which is useful at the bottom end (compressing small blobs is never useful) but less so for larger blobs, where it's not uncommon to encounter a 50MB gzipped tarball that can't be further compressed.

This allows the possibility of a function which can make more nuanced decisions based on the `uploadinfo` structures. I've also added a best-effort implementation based on some code we had already - happy to discuss whether that should be something this module provides or not.

### DIFF
--- a/go/pkg/client/cas.go
+++ b/go/pkg/client/cas.go
@@ -61,10 +61,10 @@ func (c *Client) shouldCompress(sizeBytes int64) bool {
 func (c *Client) shouldCompressEntry(ue *uploadinfo.Entry) bool {
 	if !c.shouldCompress(ue.Digest.Size) {
 		return false
-	} else if c.UploadCompressionClassifier == nil {
+	} else if c.UploadCompressionPredicate == nil {
 		return true
 	}
-	return c.UploadCompressionClassifier(ue)
+	return c.UploadCompressionPredicate(ue)
 }
 
 // makeBatches splits a list of digests into batches of size no more than the maximum.
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -134,8 +134,8 @@ type Client struct {
 	// compressed. Use 0 for all writes being compressed, and a negative number for all operations being
 	// uncompressed.
 	CompressedBytestreamThreshold CompressedBytestreamThreshold
-	// UploadCompressionClassifier is a function called to decide whether a blob should be compressed for upload.
-	UploadCompressionClassifier UploadCompressionClassifier
+	// UploadCompressionPredicate is a function called to decide whether a blob should be compressed for upload.
+	UploadCompressionPredicate UploadCompressionPredicate
 	// MaxBatchDigests is maximum amount of digests to batch in upload and download operations.
 	MaxBatchDigests MaxBatchDigests
 	// MaxQueryBatchDigests is maximum amount of digests to batch in CAS query operations.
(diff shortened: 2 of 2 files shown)


go/pkg/client/client.go
```diff
@@ -234,6 +237,16 @@ func (s CompressedBytestreamThreshold) Apply(c *Client) {
 	c.CompressedBytestreamThreshold = s
 }
 
+// An UploadCompressionClassifier determines what kind of compression to use for a blob on upload.
+// Note that the CompressedBytestreamThreshold takes priority over this (i.e. if the blob to be uploaded
+// is smaller than the threshold, this will not be called to classify it).
+type UploadCompressionClassifier func(*uploadinfo.Entry) repb.Compressor_Value
```
mrahs, line 243: I think this could simply return a boolean rather than a compressor. I believe the choice of the compression algorithm is mostly influenced by compute resources rather than the content itself. Another reason is that right now only ZSTD is supported by the SDK so the generalization adds complexity without immediate benefits.

go/pkg/client/cas_upload.go
```diff
@@ -607,3 +608,33 @@ func updateAndNotify(st *uploadState, bytesMoved int64, err error, missing bool)
 	st.clients = nil
 	st.ue = nil
 }
+
+// DetectArchiveUploads is an UploadCompressionClassifier that attempts to detect whether
+// uploads are already archives, either based on their initial few bytes or known file extensions.
+// It is a best-effort classifier that may not recognise every possible type.
+func DetectArchiveUploads(ue *uploadinfo.Entry) repb.Compressor_Value {
```
mrahs, line 615: I think this is very useful, but probably doesn't make a lot of sense to include it in the SDK. Different users might have different criteria and/or different performance requirements for this detection. Accommodating such variety becomes a bottleneck at the SDK level and is best left to the clients.

go/pkg/client/cas.go
```diff
@@ -57,6 +59,15 @@ func (c *Client) shouldCompress(sizeBytes int64) bool {
 	return int64(c.CompressedBytestreamThreshold) >= 0 && int64(c.CompressedBytestreamThreshold) <= sizeBytes
 }
 
+func (c *Client) shouldCompressUpload(ue *uploadinfo.Entry) bool {
```
mrahs, line 62: Nit: I think `shouldCompressEntry` is more readable.

mrahs (commented):
Thank you for implementing a more flexible compression option. I think the choice of forwarding the compression decision to the user is a great one!

go/pkg/client/client.go
peterebden: Yeah, I did wonder about this. I was aiming for a more flexible interface that wouldn't need to break compatibility in future if you did want to support other compressors. I agree that's a little theoretical though so I'll swap it to a bool.

go/pkg/client/cas_upload.go
peterebden: I thought you might think that. I'll leave it out, although I feel like clients will need to reinvent this kind of logic in a few places. Still, maybe better not to add this prematurely.

go/pkg/client/client.go
```diff
@@ -234,6 +237,16 @@ func (s CompressedBytestreamThreshold) Apply(c *Client) {
 	c.CompressedBytestreamThreshold = s
 }
 
+// An UploadCompressionClassifier determines what kind of compression to use for a blob on upload.
```
mrahs, line 240: This comment line is out of date. Also, I wonder if `UploadCompressionPredicate` is a more suitable name now (which would also leave the "classifier" suffix available for a future option).

go/pkg/client/client.go
```diff
@@ -133,6 +134,8 @@ type Client struct {
 	// compressed. Use 0 for all writes being compressed, and a negative number for all operations being
 	// uncompressed.
 	CompressedBytestreamThreshold CompressedBytestreamThreshold
+	// UploadCompressionClassifier is a function called to decide how blobs for upload should be compressed.
```
mrahs, line 137: Same here.
```suggestion
	// UploadCompressionClassifier is a function called to decide whether a blob should be compressed for upload.
```

go/pkg/client/cas_upload.go
mrahs: Were I not a maintainer I would definitely want this in the SDK, especially if someone smart already did the thinking for me :)
As a maintainer, I should anticipate this being a hot spot for PRs due to the large scope of implementation details.

go/pkg/client/client.go
peterebden: Yup, makes sense. Done.

mrahs (approved):

PR #443, opened 2023-04-26
Adding compression to batched blob actions

peterebden:
This updates the RE proto to latest and implements support for compression in BatchUpdateBlobs and BatchReadBlobs.


go/pkg/client/cas_upload.go
```diff
@@ -128,10 +131,15 @@ func (c *Client) BatchWriteBlobs(ctx context.Context, blobs map[digest.Digest][]
 	var sz int64
 	for k, b := range blobs {
 		sz += int64(k.Size)
```
mrahs, line 133: Optional: This size might not be as accurate as before if compression is enabled. This means that the request might carry more or less bytes that it would otherwise be able to fit. This would cause a problem if the blobs are incompressible in which case `sz` would not account for the compression headers causing the size guard below to not trigger and the gRPC call might fail. On the other hand, `sz` may count more than the size of the compressed bytes in which case the size guard below might incorrectly cancel the call.

That said, I do not see correctness issues in either case.

mrahs (commented):
Thanks for the PR, Peter.

It looks like you might need to run `bazel run //:gazzle` to update bazel's build files.

go/pkg/client/cas_upload.go
peterebden: I've changed it to count the compressed size when compressing, which should account for it as best we can.
It does seem a little bit best-effort since the implementation doesn't accurately measure the size of the encoded proto (e.g. overhead from proto encoding, and the contents of the other fields, notably digests), but I imagine it's not often an issue in practice.

peterebden:
Thanks for the review!

> It looks like you might need to run bazel run //:gazzle to update bazel's build files.

Running gazelle didn't seem to change any files? 
I've attempted to just do the update by hand which hopefully will make it happier - I'm not very familiar with Bazel or Gazelle day-to-day, and my memories of Blaze are not very useful at this point 😅

go/pkg/client/cas_upload.go
```diff
@@ -127,11 +131,18 @@ func (c *Client) BatchWriteBlobs(ctx context.Context, blobs map[digest.Digest][]
...
 			Digest: k.ToProto(),
 			Data:   b,
-		})
+		}
+		if c.batchCompression && c.shouldCompress(k.Size) {
+			r.Data = zstdEncoder.EncodeAll(r.Data, nil)
+			r.Compressor = repb.Compressor_ZSTD
+			sz += int64(len(r.Data))
+		} else {
+			sz += int64(k.Size)
```
mrahs, line 143: It's tempting to use `len(r.Data)` here as well, but this keeps this path identical to what it was before this change. Good call!

go/pkg/client/cas_upload.go
mrahs: Good observation! That's why I phrased it as "not as accurate as before". There is an additional cost to measuring the size the encoded proto, but I don't have a concrete idea of its real world impact yet. I figured it's outside the scope of this PR so what you did is a net improvement.

mrahs (approved):
LGTM. Thanks, Peter!

PR #446, opened 2023-05-09
add gazelle rule for external deps

mrahs:
This replaces the hard coded list by using go.mod as the source for gazelle.

Depends on https://github.com/bazelbuild/remote-apis-sdks/pull/445


ramymedhat (commented):
Can you remove the hardcoded list?

mrahs:
> Can you remove the hardcoded list?

I removed the file. If you are referring to the deps listed in the workspace file, those override what gazelle generates to ensure some extra arguments are present, such as disabling generating proto files.

PR #447, opened 2023-05-09
extract context metadata into its own package

mrahs:
I want to reuse the code in the `cas` package.

Depends on https://github.com/bazelbuild/remote-apis-sdks/pull/446


go/pkg/context/context.go
```diff
@@ -0,0 +1,201 @@
+// Package context allows attaching metadata to the context of RPC calls.
+package context
```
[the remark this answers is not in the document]
ramymedhat: sgtm

go/pkg/client/cas.go
```diff
@@ -6,8 +6,8 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strings"
 
+	ctxmd "github.com/bazelbuild/remote-apis-sdks/go/pkg/contextmd"
```
ramymedhat, line 10: Do we need ctxmd? Can we directly use the package name?

mrahs: Hmm, I prefer the shorter form, but sure.

ramymedhat (approved):

PR #448, opened 2023-05-10
introduce symlink options package

mrahs:
The package allows defining a set of options that cover all supported
cases by the SDK. The options instance is immutable and can only be created using
predefined set to ensure every instance has a valid set of options.

### DIFF
--- a/go/pkg/symlinkopts/opts.go
+++ b/go/pkg/symlinkopts/opts.go
@@ -1,0 +1,148 @@
+// Package symlinkopts provides an efficient interface to create unambiguous symlink options.
+package symlinkopts
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Options represents a set of options for handling symlinks. The zero value is equivalent to skipping all symlinks.
+type Options struct {
+	preserve        bool
+	noDangling      bool
+	includeTarget   bool
+	resolve         bool
+	resolveExternal bool
+}
+
+// String returns a string representation of the options.
+//
+// The string has at most five letters:
+//
+//	P for Preserve
+//	N for No Dangling
+//	I for Include Target
+//	R for Replace
+//	E for Replace External
+func (o Options) String() string {
+	var b strings.Builder
+	if o.preserve {
+		fmt.Fprint(&b, "P")
+	}
+	if o.noDangling {
+		fmt.Fprint(&b, "N")
+	}
+	if o.includeTarget {
+		fmt.Fprint(&b, "I")
+	}
+	if o.resolve {
+		fmt.Fprint(&b, "R")
+	}
+	if o.resolveExternal {
+		fmt.Fprint(&b, "E")
+	}
+	return b.String()
+}
+
+// Preserve returns true if the options include the corresponding property.
+func (o Options) Preserve() bool {
+	return o.preserve
+}
+
+// NoDangling returns true if the options include the corresponding property.
+func (o Options) NoDangling() bool {
+	return o.noDangling
+}
+
+// IncludeTarget returns true if the options include the corresponding property.
+func (o Options) IncludeTarget() bool {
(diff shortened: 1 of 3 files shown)


go/pkg/symlinkopts/opts.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"strings"
+)
+
+// Options represents a set of options for handling symlinks.
+// To ensure a valid set of options, use one of the functions provided in this package.
+type Options uint32
+
+const (
+	Preserve Options = 1 << (32 - 1 - iota)
```
ramymedhat, line 14: why not 1 << iota?

mrahs: It's MSB vs LSB. No particular reason to choosing one over the other. The `fs` package uses MSB for `FileInfo`, but that's so the lower bits can match POSIX file permissions: https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/io/fs/fs.go;l=175;drc=aedf298daf508b564e4dddc7687fff8822315a5e

go/pkg/symlinkopts/opts.go
```diff
@@ -0,0 +1,119 @@
...
+
+// ResolveExternal returns true if the options include the corresponding property.
+func (o Options) ResolveExternal() bool {
+	return o&ResolveExternal == ResolveExternal
+}
+
+// ResolveAlways return the correct set of options to always resolve symlinks.
+//
+// This implies that symlinks are followed and no dangling symlinks are allowed.
+// Each target will have the path of the symlink.
```
ramymedhat, line 78: Not sure I understand what this means.

go/pkg/symlinkopts/opts.go
```diff
@@ -0,0 +1,119 @@
...
+	return Resolve | ResolveExternal | IncludeTarget | NoDangling
+}
+
+// ResolveExternalOnly returns the correct set of options to only resolve symlinks
+// if the target is outside the execution root. Otherwise, the symlink is preserved.
+//
+// This implies that all symlinks are followed, therefore, no dangling links are allowed.
+// Otherwise, it's not possible to guarantee that all required files are under the execution root.
+// Targets of non-external symlinks are not included.
+func ResolveExternalOnly() Options {
```
ramymedhat, line 89: What is the general use case for this to make it a preset?

go/pkg/symlinkopts/opts.go
```diff
@@ -0,0 +1,119 @@
...
+	return Preserve | ResolveExternal | NoDangling
+}
+
+// ResolveExternalOnlyWithTarget is like ResolveExternalOnly but targets of non-external symlinks are included.
+func ResolveExternalOnlyWithTarget() Options {
```
ramymedhat, line 94: ditto?

ramymedhat (commented):
Not sure what the use case is for the various presets in the new package. Might be better to add presets when you use them in subsequent CLs?

go/pkg/symlinkopts/opts.go
mrahs: Same thing. It allows the user to resolve external links, but include the targets of internal ones.

mrahs:
> Not sure what the use case is for the various presets in the new package. Might be better to add presets when you use them in subsequent CLs?

The intent is to provide the valid combination as part of the package. For example, combining `Preserve` with `Resolve` doesn't make sense. The presets cover the valid combinations.

go/pkg/symlinkopts/opts.go
ramymedhat: Can you perhaps rephrase it so its more clear? Something like: Each symlink will be replaced by its target. For example, ....

ramymedhat (approved):

PR #456, opened 2023-05-26
add stderrDg and stdoutDg in MetaData

ywmei-brt1:
Introducing two digest fields in Metadata for stderr and stdout. Modified the Set method, so that instead of returning cmdDg, and acDg, it also returns stderrDg and stdoutDG. All the other unit tests are modified accordingly.

Bug: b/172888682
Tested-by: run unit test locally

### DIFF
--- a/go/pkg/fakes/server.go
+++ b/go/pkg/fakes/server.go
@@ -137,7 +137,7 @@ func timeToProto(t time.Time) *tspb.Timestamp {
 
 // Set sets up the fake to return the given result on the given command execution.
 // It is not possible to make the fake result in a LocalErrorResultStatus or an InterruptedResultStatus.
-func (e *TestEnv) Set(cmd *command.Command, opt *command.ExecutionOptions, res *command.Result, opts ...option) (cmdDg, acDg digest.Digest) {
+func (e *TestEnv) Set(cmd *command.Command, opt *command.ExecutionOptions, res *command.Result, opts ...option) (cmdDg, acDg, stderrDg, stdoutDg digest.Digest) {
 	e.t.Helper()
 	cmd.FillDefaultFieldValues()
 	if err := cmd.Validate(); err != nil {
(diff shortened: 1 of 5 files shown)


go/pkg/fakes/server.go
```diff
@@ -200,6 +200,12 @@ func (e *TestEnv) Set(cmd *command.Command, opt *command.ExecutionOptions, res *
 		ac.Timeout = dpb.New(cmd.Timeout)
 	}
 	acDg = digest.TestNewFromMessage(ac)
+	if ar.StderrDigest != nil {
+		stderrDg = digest.Digest{Hash: ar.StderrDigest.Hash, Size: ar.StderrDigest.SizeBytes}
```
mrahs, line 204: There is a convenient function for that: `digest.NewFromProtoUnvalidated(ar.StderrDigest)`.

go/pkg/rexec/rexec_test.go
```diff
@@ -235,7 +239,7 @@ func TestExecDoNotCache_NotAcceptCached(t *testing.T) {
 	// DoNotCache true implies in particular that we also skip action cache lookups, local or remote.
 	opt := &command.ExecutionOptions{DoNotCache: true, DownloadOutputs: true, DownloadOutErr: true}
 	wantRes := &command.Result{Status: command.SuccessResultStatus}
-	_, acDg := e.Set(cmd, opt, wantRes, fakes.StdOutRaw("not cached"))
+	_, _, _, acDg := e.Set(cmd, opt, wantRes, fakes.StdOutRaw("not cached"))
```
mrahs, line 242: Wrong order of variables on the LHS?

go/pkg/rexec/rexec.go
```diff
@@ -120,11 +120,19 @@ func (ec *Context) setOutputMetadata() {
 		ec.Metadata.TotalOutputBytes += int64(len(ec.resPb.StdoutRaw))
 	} else if ec.resPb.StdoutDigest != nil {
```
mrahs, line 121: The digest is always set whereas the raw content may not always be set. In other words, this branch will be skipped if both are set. However, we always want to capture the digest.

mrahs (changes requested):

mrahs (approved):

PR #457, opened 2023-05-30
support filedigest input with xattr

ywmei-brt1:
This change allows user to use a slash separated xattrValue string, i.e. "ed968e840d10d2d313a870bc131a4e2c311d7ad09bdf32b3418147221f51a6e2/5" to tell the RBE both the hash and the size of the input file digest. The string before the slash is treated as the hash value for the input file digest, and the string after the slash is treated as the size of the digest.

Currently, When user run reproxy and provide the xattrValue with a input file, we are using this xattrValue as the hash value for the input file digest, and actively calculating the actual size of the digest via file.Size(). Customer will have to fill the stub file with dummy content to the actually size to make the RBE believes the local copy (the stub file) is the same as the one RBE has .

### DIFF
--- a/go/pkg/filemetadata/filemetadata.go
+++ b/go/pkg/filemetadata/filemetadata.go
@@ -4,6 +4,7 @@ package filemetadata
 import (
 	"errors"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
@@ -112,8 +113,14 @@ func Compute(filename string) *Metadata {
 		}
 		xattrValue, err := XattrAccess.getXAttr(filename, XattrDigestName)
 		if err == nil {
+			xattrStr := string(xattrValue)
+
+			if strings.Contains(xattrStr, "/") {
+				md.Digest, md.Err = digest.NewFromString(xattrStr)
+				return md
+			}
 			md.Digest = digest.Digest{
-				Hash: string(xattrValue),
+				Hash: xattrStr,
 				Size: file.Size(),
 			}
 			return md


go/pkg/filemetadata/filemetadata.go
```diff
@@ -112,8 +113,14 @@ func Compute(filename string) *Metadata {
 		}
 		xattrValue, err := XattrAccess.getXAttr(filename, XattrDigestName)
 		if err == nil {
+			xattrStr := string(xattrValue)
+
+			if strings.Contains(xattrStr, "/") {
+				md.Digest, md.Err = digest.NewFromString(xattrStr)
+				return md
```
ramymedhat, line 120: Can you add a test for this scenario?

ramymedhat:
nit: PR title should refer to the file size since we already support reading the digest?

go/pkg/filemetadata/filemetadata.go
```diff
@@ -112,10 +114,12 @@ func Compute(filename string) *Metadata {
...
-			md.Digest = digest.Digest{
-				Hash: string(xattrValue),
-				Size: file.Size(),
+			xattrStr := string(xattrValue)
+			if strings.Contains(xattrStr, "/") {
+				md.Digest, md.Err = digest.NewFromString(xattrStr)
+				return md
 			}
+			md.Digest, md.Err = digest.NewFromString(fmt.Sprintf("%s/%d", xattrStr, file.Size()))
 			return md
```
ywmei-brt1, line 123: The previous code with `digest.Digest{}` will create digest with a invalid hash, i.e. "abc/100" (hash length should be 64);
I changed to `digest.NewFromString()` so that if user pass a invalid hash string via xatrr (say, "abc"), md.Err can capture this error.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -100,15 +103,17 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
 			defer os.RemoveAll(filename)
 			got := Compute(filename)
+			wantDigest, wantErr := digest.NewFromString(fmt.Sprintf("%s/%d", tc.name, len(tc.contents)))
 			if got.Err != nil {
-				t.Errorf("Compute(%v) failed. Got error: %v", filename, got.Err)
+				if diff := cmp.Diff(wantErr.Error(), got.Err.Error()); diff != "" {
+					t.Errorf("Compute(%v) returned diff Err. (-want +got)\n%s", filename, diff)
+				}
 			}
-			wantDigest, _ := digest.NewFromString(fmt.Sprintf("%s/%d", tc.name, len(tc.contents)))
```
ywmei-brt1, line 106: This wantErr was not captured before, I am asserting it at line 106 on the right hand side now.

go/pkg/filemetadata/filemetadata.go
```diff
@@ -112,10 +114,12 @@ func Compute(filename string) *Metadata {
 		}
 		xattrValue, err := XattrAccess.getXAttr(filename, XattrDigestName)
 		if err == nil {
-			md.Digest = digest.Digest{
-				Hash: string(xattrValue),
-				Size: file.Size(),
+			xattrStr := string(xattrValue)
+			if strings.Contains(xattrStr, "/") {
+				md.Digest, md.Err = digest.NewFromString(xattrStr)
```
ramymedhat, line 119: Can you make sure md.Err is a FileError{} similar to other code paths above?

go/pkg/filemetadata/filemetadata.go
ramymedhat: Good catch

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +123,90 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+		{
+			filename:     "provide_invalid_full_digest_(extra_slash)_will_set_md.Err",
+			contents:     "123456",
+			xattrDgStr:   "///666",
+			expectErrStr: "expected digest in the form hash/size, got ///666",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.filename, func(t *testing.T) {
+			file, err := os.Create(tc.filename)
```
ramymedhat, line 174: better to create the file under https://pkg.go.dev/testing#B.TempDir and cleanup with t.Cleanup

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +123,90 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Failed to create file: %s", err)
+			}
+			path := file.Name()
+			defer os.Remove(path)
+			defer file.Close()
```
ramymedhat, line 180: ideally the file should be closed after writing

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +123,90 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+			_, err = file.WriteString(tc.contents)
```
ramymedhat, line 181: You can also use https://pkg.go.dev/os#WriteFile

go/pkg/filemetadata/filemetadata.go
ywmei-brt1: I have changed all the errors below as FileError{}
1. compute from hash/size xattr
2. compute from hash_only xattr
3. compute directly from file without xattr

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +124,91 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+				t.Fatalf("Failed to write to file: %v\n", err)
+			}
+			if tc.xattrDgStr != "" {
+				if err = xattr.Set(path, xattrDgName, []byte(tc.xattrDgStr)); err != nil {
+					t.Fatalf("Failed to set xattr to file: %v\n", err)
+				}
+			}
+			md := Compute(path)
+			if md.Err != nil {
+				if !errors.As(md.Err, &tc.wantErr) {
```
ywmei-brt1, line 198: I try to use errors.Is(md.Err, tc.wantErr) here but does not work.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +124,91 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+			contents:   "",
+			xattrDgStr: "1111111111111111111111111111111111111111111111111111111111111111/666",
+			wantDgStr:  "1111111111111111111111111111111111111111111111111111111111111111/666",
+		},
+		{
+			filename:   "provide_invalid_digest_(hash_missing_digits)_will_set_md.Err",
+			contents:   "123456",
+			xattrDgStr: "abc",
+			wantDgStr:  "abc/6",
+			wantErr:    &FileError{Err: errors.New("valid hash length is 64, got length 3 (abc)")},
```
mrahs, line 159: Matching exact error messages will make the test case fragile. It looks like in these cases just checking for an error is sufficient.
```suggestion
			wantErr:    true,
```
And then later:
`if !test.wantErr && err != nil { t.Fatalf("Unexpected error: %v", err) }`

ywmei-brt1: Thank you. My original plan is to use the error message to deliver some extra information to the users who are reading the compute() method and help them to understand the logic. Yeah, you are right, the error msg string will make the test fragile and depend on the implementation of digest package. I have change to this to bool now.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +123,84 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+	overwriteXattrDgName(t, xattrDgName)
+	tests := []struct {
+		filename   string
+		contents   string
+		xattrDgStr string
+		wantDgStr  string
+		wantErr    bool
+	}{
+		{
+			filename:  "provide_no_xattr_will_generate_real_digest(sha256sum+real_size)",
```
ramymedhat, line 137: you can add a `name` property with a more readable short name that's not necessarily the file name. This is the convention in the rest of the code for the test name.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +123,84 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.filename, func(t *testing.T) {
+			path := tc.filename
+			//In the Github pipeline, set xatrr for files created under /tmp folder (by running t.tempDir()) will raise an error for operation not supported
+			//But local run of the unit test will pass. So we have to create these test files under the PWD.
+			err := os.WriteFile(path, []byte(tc.contents), 0666)
+			if err != nil {
+				t.Fatalf("Failed to write to file: %v\n", err)
+			}
+			defer os.RemoveAll(path)
```
ramymedhat, line 185: t.Cleanup is the recommended pattern for tests.

ywmei-brt1: I am using t.Cleanup() to delete test files now. Also changed this for all the other unit tests inside of this file for consistency.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -98,17 +101,19 @@ func TestComputeFilesWithXattr(t *testing.T) {
 				t.Fatalf("Failed to create tmp file for testing digests: %v", err)
 			}
 			after := time.Now().Truncate(time.Second).Add(time.Second)
-			defer os.RemoveAll(filename)
+			t.Cleanup(func() { os.RemoveAll(filename) })
 			got := Compute(filename)
+			wantDigest, wantErr := digest.NewFromString(fmt.Sprintf("%s/%d", tc.name, len(tc.contents)))
```
ramymedhat, line 106: Isn't an err here a failure in the setup of the test? We should never except an error here?

ywmei-brt1: The reason why I have this wantErr here is because the current `getXAttrMock()` in this test will try to return filename itself as the digest_hash, which is less than 64 bytes, and it actually raise an error after I changed the filemetadata.go code.

I got the point you mentioned here, so I have changed the code for this test so that `getXAttrMock()` will just return a valid 64 bytes long const hash string (we will only check happy path here for this existing test).

The error path which will raise an error when generating digest for metadata is covered by the new test I added below.

I think the testing logic is more clear now, with minimum change to our existing code.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -98,12 +104,12 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
 			}
 			after := time.Now().Truncate(time.Second).Add(time.Second)
-			defer os.RemoveAll(filename)
+			t.Cleanup(func() { os.RemoveAll(filename) })
 			got := Compute(filename)
 			if got.Err != nil {
 				t.Errorf("Compute(%v) failed. Got error: %v", filename, got.Err)
 			}
-			wantDigest, _ := digest.NewFromString(fmt.Sprintf("%s/%d", tc.name, len(tc.contents)))
+			wantDigest, _ := digest.NewFromString(fmt.Sprintf("%s/%d", mockedHash, len(tc.contents)))
```
ramymedhat, line 112: nit: you can still handle the error here, just that it would be a Fatal error in test setup.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +124,87 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+			contents:   "123456",
+			xattrDgStr: "///666",
+			wantDgStr:  "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855/0",
+			wantErr:    true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			filePath := tc.name
```
ramymedhat, line 179: nit: Does the file path need to be the test name? That adds the constraint that the test name can be a valid file path. You can just write to a constant or random file path.

go/pkg/filemetadata/filemetadata_test.go
```diff
@@ -118,6 +124,87 @@ func TestComputeFilesWithXattr(t *testing.T) {
...
+			//In the Github pipeline, set xatrr for files created under /tmp folder (by running t.tempDir()) will raise an error for operation not supported
```
ramymedhat, line 180: Do you know why this happens?

ywmei-brt1: The symptom is when I try to set xattr for a file created under `t.TempDir()`, I will get an error message for `operation not support`. This is the same error message as what I will get if I use something not starting with `user` for the xattr key, and this only happens on the pipeline machine, I cannot locally reproduce it. 

But I see the re-client code is following the same pattern, it is using the `t.TempDir()` for log. related files(line 67), but generate the file to set xattr to only under the working directory (line 80); 

https://source.corp.google.com/foundry-x-re-client/tests/integ/xattr/xattr_digest_test.go;l=6?q=xattr&sq=&ss=piper%2FGoogle%2Ffoundry-x-re-client#:~:text=integ.SetServerAddressEnvVariable(serverAddress)-,tmpDir%20%3A%3D%20t.TempDir(),%7D,-defer%20os.RemoveAll

go/pkg/filemetadata/filemetadata_test.go
ywmei-brt1: changed to a const named "test.txt"

go/pkg/filemetadata/filemetadata_test.go
ywmei-brt1: Act

ramymedhat (approved):

PR #459, opened 2023-06-08
digest.NewFromString() return Empty on err

ywmei-brt1:
Currently, digest.NewFromString() returns Empty, err when 1) the number of `/` in string is not equal to 1, 2) the size string cannot be parsed into int, BUT, if input is "abc/123", it will return an error for "abc" is less than 64 bytes as a valid hash, as well as a digest{"abc", 123}. To make the behavior consistent, this change makes this method always return an Empty digest if err present.


go/pkg/digest/digest.go
```diff
@@ -137,7 +137,12 @@ func NewFromString(s string) (Digest, error) {
 	if err != nil {
 		return Empty, fmt.Errorf("invalid size in digest %s: %s", s, err)
 	}
-	return New(pair[0], size)
+	if dg, err := New(pair[0], size); err != nil {
```
mrahs, line 140: Looks like `New` and `NewFromProto` would still return an invalid digest with an error. Can you update them as well? Or maybe just those two since this one is delegating to `New` anyways.

mrahs (approved):

PR #464, opened 2023-06-16
Add support for streaming stdout/stderr.

bwkimmel:
I'm adding support for stdout/stderr log streaming to RBE. This change allows the client to receive stdout/stderr streamed in near-real-time while the action is running, rather than waiting until the action is complete and then downloading stdout/stderr from CAS.

The initial motivation for this PR is that our integration tests and other diagnostic tools use this SDK. So I need support for log streaming here to support those uses.

### DIFF
--- a/go/pkg/outerr/outerr.go
+++ b/go/pkg/outerr/outerr.go
@@ -64,3 +64,39 @@ func (s *RecordingOutErr) Stdout() []byte {
 func (s *RecordingOutErr) Stderr() []byte {
 	return s.err.Bytes()
 }
+
+// outWriter is a Writer that writes to the stdout stream of an OutErr.
+type outWriter struct {
+	OutErr
+}
+
+// NewOutWriter provides an io.Writer implementation for writing to the stdout
+// stream of an OutErr.
+func NewOutWriter(oe OutErr) io.Writer {
+	return &outWriter{OutErr: oe}
+}
+
+// Write writes to the stdout stream of the OutErr. This method always succeeds,
+// returning len(p), nil.
+func (o *outWriter) Write(p []byte) (int, error) {
+	o.WriteOut(p)
+	return len(p), nil
+}
+
+// errWriter is a Writer that writes to the stderr stream of an OutErr.
+type errWriter struct {
+	OutErr
+}
+
+// NewErrWriter provides an io.Writer implementation for writing to the stderr
+// stream of an OutErr.
+func NewErrWriter(oe OutErr) io.Writer {
+	return &errWriter{OutErr: oe}
+}
+
+// Write writes to the stderr stream of the OutErr. This method always succeeds,
+// returning len(p), nil.
+func (e *errWriter) Write(p []byte) (int, error) {
+	e.WriteErr(p)
+	return len(p), nil
+}
--- a/go/pkg/outerr/outerr.go
+++ b/go/pkg/outerr/outerr.go
@@ -65,37 +65,35 @@ func (s *RecordingOutErr) Stderr() []byte {
 	return s.err.Bytes()
 }
 
-// outWriter is a Writer that writes to the stdout stream of an OutErr.
+// outWriter is a Writer that writes to the out stream of an OutErr.
 type outWriter struct {
 	OutErr
 }
 
-// NewOutWriter provides an io.Writer implementation for writing to the stdout
+// NewOutWriter provides an io.Writer implementation for writing to the out
 // stream of an OutErr.
 func NewOutWriter(oe OutErr) io.Writer {
 	return &outWriter{OutErr: oe}
 }
 
-// Write writes to the stdout stream of the OutErr. This method always succeeds,
-// returning len(p), nil.
+// Write writes to the out stream of the OutErr. This method always returns len(p), nil.
 func (o *outWriter) Write(p []byte) (int, error) {
 	o.WriteOut(p)
 	return len(p), nil
 }
 
-// errWriter is a Writer that writes to the stderr stream of an OutErr.
+// errWriter is a Writer that writes to the err stream of an OutErr.
 type errWriter struct {
 	OutErr
 }
 
-// NewErrWriter provides an io.Writer implementation for writing to the stderr
+// NewErrWriter provides an io.Writer implementation for writing to the err
 // stream of an OutErr.
 func NewErrWriter(oe OutErr) io.Writer {
 	return &errWriter{OutErr: oe}
 }
 
-// Write writes to the stderr stream of the OutErr. This method always succeeds,
-// returning len(p), nil.
+// Write writes to the stderr stream of the OutErr. This method always returns len(p), nil.
 func (e *errWriter) Write(p []byte) (int, error) {
 	e.WriteErr(p)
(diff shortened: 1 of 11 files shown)


mrahs:
Thank you for submitting this PR.

Can you please explain the motivation behind it and the intended use-case?
I'm particularly curious how it differentiates from the existing download behaviour.

bwkimmel:
I'm adding support for stdout/stderr log streaming to RBE. This change allows the client to receive stdout/stderr streamed in near-real-time while the action is running, rather than waiting until the action is complete and then downloading stdout/stderr from CAS.

The initial motivation for this PR is that our integration tests and other diagnostic tools use this SDK. So I need support for log streaming here to support those uses.

go/pkg/outerr/outerr.go
```diff
@@ -64,3 +64,39 @@ func (s *RecordingOutErr) Stdout() []byte {
...
+func NewErrWriter(oe OutErr) io.Writer {
+	return &errWriter{OutErr: oe}
+}
+
+// Write writes to the stderr stream of the OutErr. This method always succeeds,
+// returning len(p), nil.
+func (e *errWriter) Write(p []byte) (int, error) {
+	e.WriteErr(p)
+	return len(p), nil
+}
```
mrahs, line 102: I must admit, this package makes things more complicated than needs be.. but that's a tangent.
I think this is a reasonable compromise at this point. I just have a few comments about comments in this block.

go/pkg/outerr/outerr.go
```diff
@@ -64,3 +64,39 @@ func (s *RecordingOutErr) Stdout() []byte {
...
 	return s.err.Bytes()
 }
+
+// outWriter is a Writer that writes to the stdout stream of an OutErr.
+type outWriter struct {
+	OutErr
+}
+
+// NewOutWriter provides an io.Writer implementation for writing to the stdout
+// stream of an OutErr.
```
mrahs, line 74:
```suggestion
// NewOutWriter provides an io.Writer implementation for writing to the out
// stream of an OutErr.
```

go/pkg/outerr/outerr.go
```diff
@@ -64,3 +64,39 @@ func (s *RecordingOutErr) Stdout() []byte {
...
+}
+
+// Write writes to the stdout stream of the OutErr. This method always succeeds,
+// returning len(p), nil.
+func (o *outWriter) Write(p []byte) (int, error) {
+	o.WriteOut(p)
+	return len(p), nil
+}
+
+// errWriter is a Writer that writes to the stderr stream of an OutErr.
```
mrahs, line 86:
```suggestion
// errWriter is a Writer that writes to the err stream of an OutErr.
```

go/pkg/outerr/outerr.go
```diff
@@ -64,3 +64,39 @@ func (s *RecordingOutErr) Stdout() []byte {
...
+type errWriter struct {
+	OutErr
+}
+
+// NewErrWriter provides an io.Writer implementation for writing to the stderr
+// stream of an OutErr.
```
mrahs, line 92:
```suggestion
// NewErrWriter provides an io.Writer implementation for writing to the err
// stream of an OutErr.
```

go/pkg/outerr/outerr.go
```diff
@@ -64,3 +64,39 @@ func (s *RecordingOutErr) Stdout() []byte {
...
+func NewErrWriter(oe OutErr) io.Writer {
+	return &errWriter{OutErr: oe}
+}
+
+// Write writes to the stderr stream of the OutErr. This method always succeeds,
+// returning len(p), nil.
```
mrahs, line 98: Same here.
```suggestion
// Write writes to the stderr stream of the OutErr. This method always returns len(p), nil.
```

go/pkg/rexec/rexec.go
```diff
@@ -330,17 +332,48 @@ func (ec *Context) ExecuteRemotely() {
 	ec.Metadata.RealBytesUploaded = bytesMoved
 	log.V(1).Infof("%s %s> Executing remotely...\n%s", cmdID, executionID, strings.Join(ec.cmd.Args, " "))
 	ec.Metadata.EventTimes[command.EventExecuteRemotely] = &command.TimeInterval{From: time.Now()}
-	op, err := ec.client.GrpcClient.ExecuteAndWait(ec.ctx, &repb.ExecuteRequest{
+	eg, ctx := errgroup.WithContext(ec.ctx)
```
mrahs, line 335: This will cancel both streams if either of them fails. Is that intended?

go/pkg/rexec/rexec.go
```diff
@@ -364,7 +397,16 @@ func (ec *Context) ExecuteRemotely() {
 		ec.setOutputMetadata()
 		ec.Result = command.NewResultFromExitCode((int)(ec.resPb.ExitCode))
 		if ec.opt.DownloadOutErr {
-			ec.Result = ec.downloadOutErr()
```
mrahs, line 367: Why not just add the new option to the boolean guard on the previous line?

go/pkg/rexec/rexec.go
```diff
@@ -330,17 +332,48 @@ func (ec *Context) ExecuteRemotely() {
...
+		}
 	})
 	ec.Metadata.EventTimes[command.EventExecuteRemotely].To = time.Now()
 	if err != nil {
 		ec.Result = command.NewRemoteErrorResult(err)
 		return
 	}
 
+	if err := eg.Wait(); err != nil {
+		ec.Result = command.NewRemoteErrorResult(fmt.Errorf("failure writing output streams: %v", err))
```
mrahs, line 373: Should this be a fatal error? Why not fallback to downloading the streams?

go/pkg/outerr/outerr.go
bwkimmel: Yeah, I agree. I was thinking `OutErr` doesn't really seem to add much, but I also didn't want to consider removing it as in-scope for this PR.

go/pkg/rexec/rexec.go
mrahs: Sounds like from the user's perspective, streaming overrides downloading, with a fallback to downloading if the server doesn't provide a stream name.

If that's the case, I'd update godocs for both options to clarify this behaviour. Essentially, the godoc should indicate that the download option is ignored if streaming is on, and the streaming option godoc should indicate that the user should expect a delay due to a download fallback if the server doesn't respond to the streaming request.

go/pkg/rexec/rexec.go
mrahs: I agree. The complexity of the solution requires a sufficient demand for it which is currently not there or we don't know about it.
Failing if either stream fails sounds like it would be similar to failing if downloading either stream fails, which is the current behaviour.

go/pkg/rexec/rexec.go
bwkimmel: Done.

mrahs (approved):

go/pkg/rexec/rexec.go
bwkimmel: Done.

Issue #460, opened 2023-06-08
Support node properties in the remote tool


ola-rozenfeld:
## Problem

[node properties](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L852) are a (relatively recent) part of the RE-API which is relied upon by Bazel's persistent remote workers feature: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/merkletree/MerkleTree.java#L398
Bazel uses this to "mark" some files as tools that should persist on the remote worker. This is used by RE implementations (definitely by [EngFlow](https://docs.engflow.com/re/client/remote-persistent-workers.html?gclid=CjwKCAjw-IWkBhBTEiwA2exyO76Skm_Tj7kSkiY5suIa8kHH3pflVwcP95Lqm-I6hahzXw2RGo7v5hoCnVsQAvD_BwE#client-side-requirements), unsure about Google's RBE).
However, https://github.com/bazelbuild/remote-apis-sdks doesn't support node properties at the moment, which, for example, makes it impossible to use the [remotetool](https://github.com/bazelbuild/remote-apis-sdks/tree/master/go/cmd/remotetool) to debug actions that use remote persistent workers.

## Suggested fix

Introduce optional node properties specification throughout the rexec client layer, starting with adding it to the [command proto](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/api/command/command.proto):

```
message InputSpec {
  ...
  map<string,NodeProperties> input_node_properties = 7;
}

// Either copied from [RE-API](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L852) or introduce a proto dependency on it
message NodeProperties {
...
}
```

and then reading these in [GetDirectoryTree](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/client/cas_download.go#L412), passing them through to rexec [here](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/tool/tool.go#L301), [here](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/tool/tool.go#L590) and [here](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/tool/tool.go#L82).

Please let me know if you're okay with this plan, and I'll try to send a PR shortly.

Thank you!

[...]

PR #466, opened 2023-06-18
noop/cleanup: use fileSysNode in treeNode

ola-rozenfeld:
Tiny cleanup in preparation for fixing https://github.com/bazelbuild/remote-apis-sdks/issues/460

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -23,9 +23,8 @@ import (
 // tree later. It corresponds roughly to a *repb.Directory, but with pointers, not digests, used to
 // refer to other nodes.
 type treeNode struct {
-	files    map[string]*fileNode
-	dirs     map[string]*treeNode
-	symlinks map[string]*symlinkNode
+	leaves   map[string]*fileSysNode
+	children map[string]*treeNode
 }
 
 type fileNode struct {
@@ -327,37 +326,30 @@ func buildTree(files map[string]*fileSysNode) (*treeNode, error) {
 
 		node := root
 		for _, s := range segs {
-			if node.dirs == nil {
-				node.dirs = make(map[string]*treeNode)
+			if node.children == nil {
+				node.children = make(map[string]*treeNode)
 			}
-			child := node.dirs[s]
+			child := node.children[s]
 			if child == nil {
 				child = &treeNode{}
-				node.dirs[s] = child
+				node.children[s] = child
 			}
 			node = child
 		}
 
 		if fn.emptyDirectoryMarker {
-			if node.dirs == nil {
-				node.dirs = make(map[string]*treeNode)
+			if node.children == nil {
+				node.children = make(map[string]*treeNode)
 			}
-			if node.dirs[base] == nil {
-				node.dirs[base] = &treeNode{}
+			if node.children[base] == nil {
+				node.children[base] = &treeNode{}
 			}
 			continue
 		}
-		if fn.file != nil {
-			if node.files == nil {
-				node.files = make(map[string]*fileNode)
-			}
-			node.files[base] = fn.file
-		} else {
-			if node.symlinks == nil {
-				node.symlinks = make(map[string]*symlinkNode)
-			}
-			node.symlinks[base] = fn.symlink
+		if node.leaves == nil {
+			node.leaves = make(map[string]*fileSysNode)
 		}
+		node.leaves[base] = fn
 	}
 	return root, nil
 }
@@ -366,7 +358,7 @@ func packageTree(t *treeNode, stats *TreeStats) (root digest.Digest, blobs map[d
 	dir := &repb.Directory{}
 	blobs = make(map[digest.Digest]*uploadinfo.Entry)
 
-	for name, child := range t.dirs {
+	for name, child := range t.children {
 		dg, childBlobs, err := packageTree(child, stats)
 		if err != nil {
 			return digest.Empty, nil, err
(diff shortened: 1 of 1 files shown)


go/pkg/client/tree.go
```diff
@@ -378,19 +370,21 @@ func packageTree(t *treeNode, stats *TreeStats) (root digest.Digest, blobs map[d
...
-		stats.TotalInputBytes += dg.Size
+	for name, n := range t.leaves {
+		if n.file != nil {
+			dg := n.file.ue.Digest
+			dir.Files = append(dir.Files, &repb.FileNode{Name: name, Digest: dg.ToProto(), IsExecutable: n.file.isExecutable})
+			blobs[dg] = n.file.ue
+			stats.InputFiles++
+			stats.TotalInputBytes += dg.Size
+		}
+		if n.symlink != nil {
```
mrahs, line 381: Can a node be both a file and a symlink? If not, that should be an error to avoid a tree with unexpected nodes.

However, the previous code gave precedence to the file. Maintaining that is also reasonable at this point.

go/pkg/client/tree.go
```diff
@@ -523,16 +517,17 @@ func packageDirectories(t *treeNode) (root *repb.Directory, files map[digest.Dig
...
-		dg := fn.ue.Digest
-		root.Files = append(root.Files, &repb.FileNode{Name: name, Digest: dg.ToProto(), IsExecutable: fn.isExecutable})
-		files[dg] = fn.ue
+	for name, n := range t.leaves {
+		if n.file != nil {
+			dg := n.file.ue.Digest
+			root.Files = append(root.Files, &repb.FileNode{Name: name, Digest: dg.ToProto(), IsExecutable: n.file.isExecutable})
+			files[dg] = n.file.ue
+		}
+		if n.symlink != nil {
```
mrahs, line 526: Same comment here about precedence and unexpected nodes.

mrahs (changes requested):

go/pkg/client/tree.go
ola-rozenfeld: PTAL: Added an enforced consistency check to the internal fileSysNode type (should be exactly one of file/symlink/directory). Cannot unit test because it is an internal type.

go/pkg/client/tree.go
ola-rozenfeld: Added a consistency check. I think erroring out is better here, because this should never happen (and if it does, we have a bug somewhere in this file, because it's an internal type). Would have been easier to write in a language where we could enforce immutability and constructors (because then we could just make sure that all fileSysNodes are created with exactly one of these variables set).

mrahs: Hmm, I'm having a second thought since this is a internal error that the user has no way of handling. I'm leaning towards maintaining existing semantics by just letting a file node take precedence. Thoughts?

ola-rozenfeld: Note that the current code maintains the existing semantics precisely -- we used to keep the file and symlink nodes in separate arrays and go over all of them independently. So there was no "precedence" -- we would go over all of them, and if there was an inconsistency, we would encode both the file and the symlink (creating an invalid tree proto).

I think: either we want to check internal consistency or we trust our current code and don't check it. I'm leaning towards checking/enforcing, since "trust our current code" is famous last words :-)
Then, the only question is how to make the existing code more readable (to make it clear that a fileSysNode is a union). How about this:
- add a `type` field in `fileSysNode` which is one of:

```
type NodeType int

const (
    File NodeType = iota
    Symlink
    Directory
)
```

And then instead of the if cases, do a `switch` over the `type` field. This would change the existing semantics, but it would be a bit clearer what's going on here. WDYT?

mrahs: I thought the file node was taking precedence over the symlink in this if-else statement: https://github.com/bazelbuild/remote-apis-sdks/pull/466/files#diff-45c6218ebcf3e4b4c948697c226c2d8d9ef4a3f8b1e22d420cb4dc655e74e143L350

Isn't the marker field just a more verbose way of preferring a field over another? It's still possible to have multiple fields set, but now the node type field determines which one is actually used.

I think a robust solution would need to involve generics or an interface-based polymorphism. Both of which sound out of scope for your purposes.

I'm happy to approve a compromise where we maintain existing semantics.

mrahs: I think just adding a `continue` above this line would do it.

ola-rozenfeld: > I thought the file node was taking precedence over the symlink in this if-else statement:

Yes, but then we trust the arrays under the treeNode to be correct in a different function. The code is internally consistent and fileSysNode is always created and used as a union. Yeah, maybe since it's internal we don't want to enforce it with a user-visible error...

> I think a robust solution would need to involve generics or an interface-based polymorphism.

Yeah, there is no truly robust solution in Go (need to enforce immutability and constructor usage), and trying to add one would go against the grain of the language, I think (Go is, like, "it's okay, we all know what we're doing"...)

Sure, I added an else case and a comment instead, PTAL.

mrahs: Sounds good.

One last nit if you don't mind: [`continue` rather than `else`](https://google.github.io/styleguide/go/decisions#indent-error-flow)

ola-rozenfeld: Done, although here I disagree:
1. This is not error flow (so the style guideline does not apply).
2. The indent is the same either way (so continue will not make the rest of the code easier to read).
3. `else` makes it a clearer, imo, that the cases are intended as disjoint. This is semantically a `switch` statement.

Up to you.

Thank you!

mrahs: Thanks.

I think this is more consistent with the style guide, which cannot be comprehensive. Code readability is highly subjective and I try to keep it as close to the style guide as possible, even when I disagree myself, which helps with consistency across code bases.

PR #471, opened 2023-06-30
update go to 1.20.7

atetubou:


mrahs:
Thanks for the PR. We are trying to stick to the minimum supported version to avoid forcing users to upgrade if they don't want to. [Go 1.21 is expected to be released this August](https://tip.golang.org/doc/go1.21#:~:text=Go%201.21%20is%20expected%20to%20be%20released%20in%20August%202023.), after which this PR would be appreciated.

If there is a specific reason to having to upgrade now, perhaps we can find a short-term workaround, which we [did before](https://github.com/bazelbuild/remote-apis-sdks/blob/6c9ff4481d5c86436e5272eed87d6ffc981da557/go/pkg/errors/errors.go#L2).

atetubou:
I see. I can wait for release of Go 1.21 then.

.bazelci/presubmit.yml
```diff
@@ -3,12 +3,12 @@ validate_config: 1
 tasks:
   ubuntu1804:
     environment:
-      GO_HOME: "$HOME/go-1.19.5"
+      GO_HOME: "$HOME/go-1.20.5"
```
mrahs, line 6: Can you update the [patch to the latest](http://go/go-release-history#go1.20.7) one as well?

WORKSPACE
```diff
@@ -4,10 +4,10 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "6b65cb7917b4d1709f9410ffe00ecf3e160edf674b78c54a894471320862184f",
+    sha256 = "51dc53293afe317d2696d4d6433a4c33feedb7748a9e352072e2ec3c0dafd2c6",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.39.0/rules_go-v0.39.0.zip",
-        "https://github.com/bazelbuild/rules_go/releases/download/v0.39.0/rules_go-v0.39.0.zip",
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.40.1/rules_go-v0.40.1.zip",
+        "https://github.com/bazelbuild/rules_go/releases/download/v0.40.1/rules_go-v0.40.1.zip",
```
mrahs, line 10: Can you update this one to the [latest](https://github.com/bazelbuild/rules_go/releases/tag/v0.41.0) version?

mrahs (changes requested):

.bazelci/presubmit.yml
atetubou: Done

WORKSPACE
atetubou: I excluded rules_go version update as it requires much more work and investigation of build failure which I'm not so familiar to.

PR #472, opened 2023-06-30
fix regression in downloading out/err streams

mrahs:
The server may terminate the streaming request before sending available
bytes, which prevents the download fallback from getting them.

This change implements a resuming fallback to download the remainder of
the streams.

Additionally, the digests of the out and err streams will be printed in
the output of `execution_action` command to allow the user to access
them even if the remote execution fails, in which case they will not be
retrievable by `show_action`.

### DIFF
--- a/go/pkg/rexec/rexec.go
+++ b/go/pkg/rexec/rexec.go
@@ -14,7 +14,6 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/filemetadata"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/outerr"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
-	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/prototext"
@@ -76,15 +75,20 @@ func (c *Client) NewContext(ctx context.Context, cmd *command.Command, opt *comm
 	}, nil
 }
 
-func (ec *Context) downloadStream(raw []byte, dgPb *repb.Digest, write func([]byte)) error {
+// downloadStream reads the blob for the digest dgPb into memory and forwards the bytes to the write function.
+func (ec *Context) downloadStream(raw []byte, dgPb *repb.Digest, offset int64, write func([]byte)) error {
 	if raw != nil {
-		write(raw)
+		o := int(offset)
+		if int64(o) != offset || o > len(raw) {
+			return fmt.Errorf("offset %d is out of range for length %d", offset, len(raw))
+		}
+		write(raw[o:])
 	} else if dgPb != nil {
 		dg, err := digest.NewFromProto(dgPb)
 		if err != nil {
 			return err
 		}
-		bytes, stats, err := ec.client.GrpcClient.ReadBlob(ec.ctx, dg)
+		bytes, stats, err := ec.client.GrpcClient.ReadBlobRange(ec.ctx, dg, offset, 0)
 		if err != nil {
 			return err
 		}
@@ -137,10 +141,10 @@ func (ec *Context) setOutputMetadata() {
 }
 
 func (ec *Context) downloadOutErr() *command.Result {
-	if err := ec.downloadStream(ec.resPb.StdoutRaw, ec.resPb.StdoutDigest, ec.oe.WriteOut); err != nil {
+	if err := ec.downloadStream(ec.resPb.StdoutRaw, ec.resPb.StdoutDigest, 0, ec.oe.WriteOut); err != nil {
 		return command.NewRemoteErrorResult(err)
 	}
-	if err := ec.downloadStream(ec.resPb.StderrRaw, ec.resPb.StderrDigest, ec.oe.WriteErr); err != nil {
+	if err := ec.downloadStream(ec.resPb.StderrRaw, ec.resPb.StderrDigest, 0, ec.oe.WriteErr); err != nil {
 		return command.NewRemoteErrorResult(err)
 	}
 	return command.NewResultFromExitCode((int)(ec.resPb.ExitCode))
@@ -332,9 +336,12 @@ func (ec *Context) ExecuteRemotely() {
 	ec.Metadata.RealBytesUploaded = bytesMoved
 	log.V(1).Infof("%s %s> Executing remotely...\n%s", cmdID, executionID, strings.Join(ec.cmd.Args, " "))
 	ec.Metadata.EventTimes[command.EventExecuteRemotely] = &command.TimeInterval{From: time.Now()}
-	eg, ctx := errgroup.WithContext(ec.ctx)
+	// Initiate each streaming request once at most.
 	var streamOut, streamErr sync.Once
-	op, err := ec.client.GrpcClient.ExecuteAndWaitProgress(ctx, &repb.ExecuteRequest{
+	var streamWg sync.WaitGroup
+	// These variables are owned by the progress callback (which is async but not concurrent) until the execution returns.
+	var nOutStreamed, nErrStreamed int64
+	op, err := ec.client.GrpcClient.ExecuteAndWaitProgress(ec.ctx, &repb.ExecuteRequest{
 		InstanceName:    ec.client.GrpcClient.InstanceName,
 		SkipCacheLookup: !ec.opt.AcceptCached || ec.opt.DoNotCache,
 		ActionDigest:    ec.Metadata.ActionDigest.ToProto(),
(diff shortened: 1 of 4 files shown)


mrahs:
@bwkimmel can you ptal and make sure it still works for your purposes?

go/pkg/rexec/rexec.go
```diff
@@ -332,35 +341,31 @@ func (ec *Context) ExecuteRemotely() {
 	ec.Metadata.RealBytesUploaded = bytesMoved
 	log.V(1).Infof("%s %s> Executing remotely...\n%s", cmdID, executionID, strings.Join(ec.cmd.Args, " "))
 	ec.Metadata.EventTimes[command.EventExecuteRemotely] = &command.TimeInterval{From: time.Now()}
-	eg, ctx := errgroup.WithContext(ec.ctx)
-	var streamOut, streamErr sync.Once
-	op, err := ec.client.GrpcClient.ExecuteAndWaitProgress(ctx, &repb.ExecuteRequest{
+	// These variables are owned by the progress callback (which is async but not conrrent) until the execution returns.
```
bwkimmel, line 344: concurrent?

go/pkg/rexec/rexec.go
```diff
@@ -332,35 +341,31 @@ func (ec *Context) ExecuteRemotely() {
...
+	op, err := ec.client.GrpcClient.ExecuteAndWaitProgress(ec.ctx, &repb.ExecuteRequest{
 		InstanceName:    ec.client.GrpcClient.InstanceName,
 		SkipCacheLookup: !ec.opt.AcceptCached || ec.opt.DoNotCache,
 		ActionDigest:    ec.Metadata.ActionDigest.ToProto(),
 	}, func(md *repb.ExecuteOperationMetadata) {
-		if !ec.opt.StreamOutErr {
+		if streaming || !ec.opt.StreamOutErr {
 			return
 		}
+		streaming = true
```
bwkimmel, line 355: This assumes that if the stdout/stderr stream names are provided, they are always provided on the first callback. I think RBE does always do this, but is this a requirement of the API? Or would it be technically valid to provide progress callbacks for a while before you end up sending a callback that includes the stream names?

go/pkg/rexec/rexec.go
```diff
@@ -332,35 +341,31 @@ func (ec *Context) ExecuteRemotely() {
...
-					path := fmt.Sprintf("%s/logstreams/%s", ec.client.GrpcClient.InstanceName, name)
-					log.V(1).Infof("%s %s> Streaming to stdout from %q", cmdID, executionID, path)
-					_, err = ec.client.GrpcClient.ReadResourceTo(ctx, path, outerr.NewOutWriter(ec.oe))
-					return err
-				})
-			})
+			path := fmt.Sprintf("%s/logstreams/%s", ec.client.GrpcClient.InstanceName, name)
+			log.V(1).Infof("%s %s> Streaming to stdout from %q", cmdID, executionID, path)
+			// Ignoring the error here since the net result is downloading the full stream after the fact.
+			n, _ := ec.client.GrpcClient.ReadResourceTo(ec.ctx, path, outerr.NewOutWriter(ec.oe))
```
bwkimmel, line 360: This will try to download the whole stdout stream synchronously while the first callback is being invoked (and then synchronously download the entire stderr stream below). I don't think this is what we want here. We still need to start a goroutine for each stream.

Note that this callback function will typically be called multiple times during execution to update the caller on the progress of the action. We don't want to hold up the first such call trying to download the whole of both stream synchronously, one after the other.

Since we don't care about handling the errors below (we only need to know how much was successfully streamed), perhaps use a sync.WaitGroup instead of an errgroup?

go/pkg/rexec/rexec.go
```diff
@@ -332,35 +341,31 @@ func (ec *Context) ExecuteRemotely() {
...
-					path := fmt.Sprintf("%s/logstreams/%s", ec.client.GrpcClient.InstanceName, name)
-					log.V(1).Infof("%s %s> Streaming to stdout from %q", cmdID, executionID, path)
-					_, err = ec.client.GrpcClient.ReadResourceTo(ctx, path, outerr.NewErrWriter(ec.oe))
-					return err
-				})
-			})
+			path := fmt.Sprintf("%s/logstreams/%s", ec.client.GrpcClient.InstanceName, name)
+			log.V(1).Infof("%s %s> Streaming to stdout from %q", cmdID, executionID, path)
+			// Ignoring the error here since the net result is downloading the full stream after the fact.
+			n, _ := ec.client.GrpcClient.ReadResourceTo(ec.ctx, path, outerr.NewErrWriter(ec.oe))
```
bwkimmel, line 367: I think the net effect here is that we won't even try to download stderr until the action is finished, since we won't get here until stdout is finished being read from above.

go/pkg/rexec/rexec.go
```diff
@@ -397,16 +397,24 @@ func (ec *Context) ExecuteRemotely() {
...
-					ec.Result = command.NewRemoteErrorResult(err)
-				}
-			})
+			offset := int(nOutStreamed)
+			if int64(offset) != nOutStreamed {
+				// Offset is larger than math.MaxInt, the stream cannot be read anyways.
+				// Resetting to 0 here to allow the error from downstream to bubble back up here.
+				offset = 0
+			}
+			if err := ec.downloadStream(ec.resPb.StdoutRaw, ec.resPb.StdoutDigest, offset, ec.oe.WriteOut); err != nil {
```
bwkimmel, line 406: Should we check whether streaming already succeeded -- perhaps if offset == digest.SizeBytes -- and not download if it has?

go/pkg/rexec/rexec.go
```diff
@@ -88,9 +94,12 @@ func (ec *Context) downloadStream(raw []byte, dgPb *repb.Digest, write func([]by
 		if err != nil {
```
bwkimmel, line 94: If we're going to support reading from an offset here, perhaps we should use the corresponding [ByteStream feature](https://github.com/googleapis/googleapis/blob/511319c6af9172bf12de3e80672b8109c49efd29/google/bytestream/bytestream.proto#L107), rather than reading the entire blob and then filtering it.

mrahs: Good point.

go/pkg/rexec/rexec.go
mrahs: Nothing is specified in [the API](https://github.com/bazelbuild/remote-apis/blob/068363a3625e166056c155f6441cfb35ca8dfbf2/build/bazel/remote/execution/v2/remote_execution.proto#L118), which means the server may return either, neither or both, and not necessarily in the same or first call.

Good call!

go/pkg/rexec/rexec.go
mrahs: Good catch! I was preoccupied with the fact that the callback is not concurrent that I forgot this call is synchronous.

go/pkg/rexec/rexec.go
```diff
@@ -369,11 +374,6 @@ func (ec *Context) ExecuteRemotely() {
 		return
 	}
 
-	if err := eg.Wait(); err != nil {
```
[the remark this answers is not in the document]
mrahs: Good catch.

go/pkg/rexec/rexec.go
mrahs: There is not such optimization downstream, which is not what I expected. I suppose it's good to add one here.

go/pkg/rexec/rexec.go
mrahs: I'm not following you here. The download below always happens after the execution call has returned.

bwkimmel: Sorry for the ambiguous wording. :) I was referring to the streaming download that happens on this line (not the blob download that takes place after the action has completed if necessary). Because you had removed the goroutines and called ReadResourceTo for both stdout and stderr, one after the other, and synchronously, the first ReadResourceTo call (for stdout) wouldn't finish until the action completes (because the API server won't close the stdout stream until the action completes), and only then would we start "streaming" (not really streaming) stderr. The intent was to be streaming both concurrently while the action is still running.

Now that you've added goroutines back, this comment is resolved.

go/pkg/rexec/rexec.go
```diff
@@ -332,48 +336,65 @@ func (ec *Context) ExecuteRemotely() {
...
-					_, err = ec.client.GrpcClient.ReadResourceTo(ctx, path, outerr.NewOutWriter(ec.oe))
-					return err
-				})
+					// Ignoring the error here since the net result is downloading the full stream after the fact.
+					n, err := ec.client.GrpcClient.ReadResourceTo(ec.ctx, path, outerr.NewOutWriter(ec.oe))
+					if err != nil {
+						log.Errorf("%s %s> error streaming stdout: %v", cmdID, executionID, err)
+					}
+					nOutStreamed += n
+					streamWg.Done()
```
bwkimmel, line 366: Nit: it's a bit more robust if you use `defer wg.Done()` as the first statement in the goroutine. That way some later change that exits the goroutine early won't result in Done() not being called (which would cause a deadlock).

bozydarsz (approved):

PR #473, opened 2023-07-01
Fixing bug in fake and integration test

ola-rozenfeld:
Bug was introduced in https://github.com/bazelbuild/remote-apis-sdks/pull/161.
Fake input files were never actually written to disk, like they were supposed to.

### DIFF
--- a/go/pkg/tool/tool_test.go
+++ b/go/pkg/tool/tool_test.go
@@ -83,8 +83,8 @@ Platform
 
 Inputs
 ======
-[Root directory digest: e23e10be0d14b5b2b1b7af32de78dea554a74df5bb22b31ae6c49583c1a8aa0e/75]
-a/b/input.txt: [File digest: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855/0]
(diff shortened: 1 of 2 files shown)


go/pkg/tool/tool_test.go
```diff
@@ -83,8 +83,8 @@ Platform
 
 Inputs
 ======
-[Root directory digest: e23e10be0d14b5b2b1b7af32de78dea554a74df5bb22b31ae6c49583c1a8aa0e/75]
-a/b/input.txt: [File digest: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855/0]
+[Root directory digest: 1b1d5b9e3eb97866805deca901299fcdde98b5747e147f35187e12c16e120186/75]
+a/b/input.txt: [File digest: c96c6d5be8d08a12e7b5cdc1b207fa6b2430974c86803d8891675e76fd992c20/5]
```
ola-rozenfeld, line 87: Size 5 is now correct, because we set the contents to "input" in line 69 above.

mrahs (approved):

go/pkg/tool/tool_test.go
mrahs: I wonder if it was intentional to minimize IO overhead, but I can see it doesn't generalize for other test scenarios.

PR #475, opened 2023-07-05
Node properties: support for leaf nodes

ola-rozenfeld:
Progress on https://github.com/bazelbuild/remote-apis-sdks/issues/460 (95% of the fix).

In this PR, only support node properties for leaf nodes (files and empty directories); the reason is that to do it for all nodes will require me to change a current invariant / output in tree.go, and I'd like to split it to a separate PR, as it is a bit more invasive change. Also, in practice support for non-leaf nodes is much less urgent, because the main goal here is to support Bazel's use case, which is to mark [files](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/merkletree/MerkleTree.java#L398) as "bazel_tool_input" in order to implement persistent remote workers.


go/api/command/BUILD.bazel
```diff
@@ -22,10 +26,3 @@ go_library(
     importpath = "github.com/bazelbuild/remote-apis-sdks/go/api/command",
     visibility = ["//visibility:public"],
 )
-
-proto_library(
```
ola-rozenfeld, line 26: Not sure why there were two copies of this target.

go/pkg/command/command.go
```diff
@@ -585,6 +588,35 @@ func (c *Command) ToREProto(useOutputPathsField bool) *repb.Command {
 	return cmdPb
 }
 
+func FromREProto(cmdPb *repb.Command) *Command {
```
ola-rozenfeld, line 591: This was moved here from tool.go, because it thematically belongs here, imo, next to the `ToREProto` method.

go/pkg/tool/tool.go
```diff
@@ -99,18 +101,43 @@ func (c *Client) prepCommand(ctx context.Context, client *rexec.Client, actionDi
...
+		nodeProperties = make(map[string]*repb.NodeProperties)
+		for path, t := range ts {
+			if t.NodeProperties != nil {
+				nodeProperties[path] = t.NodeProperties
+			}
+		}
+	} else {
+		inputRoot = filepath.Join(actionRoot, "input")
+		npPath := filepath.Join(actionRoot, "input_node_properties.textproto")
+		if _, err := os.Stat(npPath); err == nil {
```
ola-rozenfeld, line 128: I tried to rewrite this with early return, but failed -- only way I saw to return early is to duplicate one of the if clauses, which was, imo, worse.

go/pkg/tool/tool.go
```diff
@@ -362,8 +371,20 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
 	if err != nil {
 		return err
 	}
-	_, _, err = c.GrpcClient.DownloadDirectory(ctx, rDg, rootPath, filemetadata.NewNoopCache())
-	return err
+	is := &cpb.InputSpec{InputNodeProperties: make(map[string]*repb.NodeProperties)}
```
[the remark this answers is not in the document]

go/pkg/tool/tool.go
```diff
@@ -121,35 +136,32 @@ func (c *Client) prepCommand(ctx context.Context, client *rexec.Client, actionDi
...
-		OutputFiles: cmdPb.OutputFiles,
-		OutputDirs:  cmdPb.OutputDirectories,
-		Platform:    make(map[string]string),
-		Args:        cmdPb.Arguments,
-	}
-
-	for _, ev := range cmdPb.EnvironmentVariables {
-		cmd.InputSpec.EnvironmentVariables[ev.Name] = ev.Value
+func readNodePropertiesFromFile(path string) (nps map[string]*repb.NodeProperties, err error) {
+	if _, err = os.Stat(path); err == nil {
```
mrahs, line 150: Can you rearrange the conditions such that [the terminal one appears first](https://google.github.io/styleguide/go/decisions#indent-error-flow:~:text=terminal%20condition%20is%20not%20met%20should%20appear%20after%20the%20if%20block)?

go/pkg/command/command.go
```diff
@@ -585,6 +588,35 @@ func (c *Command) ToREProto(useOutputPathsField bool) *repb.Command {
...
+		Identifiers: &Identifiers{},
+		WorkingDir:  cmdPb.WorkingDirectory,
+		OutputFiles: cmdPb.OutputFiles,
+		OutputDirs:  cmdPb.OutputDirectories,
+		Platform:    make(map[string]string),
+		Args:        cmdPb.Arguments,
+	}
+
+	// In v2.1 of the RE API the `output_{files, directories}` fields were
+	// replaced by a single field: `output_paths`.
```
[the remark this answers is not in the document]
mrahs: Thanks for the well laid-out comment.

My concern was specifically around losing the information about which paths are directories. As you mentioned, this is only a problem when passing a command from 2.1+ server to a 2.0- one, in which case the information (directories) wasn't or shouldn't be there in the first place.

mrahs (approved):

PR #479, opened 2023-07-11
improve log tracing and filtering

mrahs:
No functional changes.
Added a internal field to trace a request as it propagates throughout the system.
Updated the log message format for easier filtering using grep.

### DIFF
--- a/go/pkg/casng/batching.go
+++ b/go/pkg/casng/batching.go
@@ -27,7 +27,7 @@ import (
 // In other words, if an error is returned, any digest that is not in the returned slice is not missing.
 // If no error is returned, the returned slice contains all the missing digests.
 func (u *BatchingUploader) MissingBlobs(ctx context.Context, digests []digest.Digest) ([]digest.Digest, error) {
-	contextmd.Infof(ctx, log.Level(1), "[casng] batch.query: len=%d", len(digests))
+	contextmd.Infof(ctx, log.Level(1), "[casng] batch.query; len=%d", len(digests))
 	if len(digests) == 0 {
 		return nil, nil
 	}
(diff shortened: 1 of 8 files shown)


go/pkg/casng/pubsub.go
```diff
@@ -8,13 +8,9 @@ import (
...
 )
 
-// tag identifies a pubsub channel for routing purposes.
-// Producers tag messages and consumers subscribe to tags.
-type tag string
-
 // pubsub provides a simple pubsub implementation to route messages and wait for them.
 type pubsub struct {
-	subs    map[tag]chan any
+	subs    map[string]chan any
```
bentekkie, line 13: What was the motivation to move from the `tag` alias to just `string`. Wouldnt the alias help ensure that the tags are valid?

mrahs: In a subsequent PR, I'll have code like `strings.Join(r.tags, "|")` in a log message. The custom types forces me to reallocate the array with an explicit string type. That or the use of generics sounds too much for a simple log message. The value of the custom type is entirely for readability as I don't expect it to have any special behaviour attached to it. It's treated as a string everywhere in this package.

go/pkg/casng/batching.go
```diff
@@ -101,29 +101,33 @@ func (u *BatchingUploader) MissingBlobs(ctx context.Context, digests []digest.Di
 // The errors returned are either from the context, ErrGRPC, ErrIO, or ErrCompression. More errors may be wrapped inside.
 // If an error was returned, the returned stats may indicate that all the bytes were sent, but that does not guarantee that the server committed all of them.
 func (u *BatchingUploader) WriteBytes(ctx context.Context, name string, r io.Reader, size, offset int64) (Stats, error) {
+	startTime := time.Now()
 	if !u.streamThrottle.acquire(ctx) {
 		return Stats{}, ctx.Err()
 	}
 	defer u.streamThrottle.release()
+	log.V(3).Infof("[casng] upload.write_bytes.throttle.duration; start=%d, end=%d", startTime.UnixNano(), time.Now().UnixNano())
```
bozydarsz, line 109: optional: maybe calculate duration instead of providing start end end. Also: wouldn't end be already present in log's timestamp?

bozydarsz (approved):

go/pkg/casng/batching.go
mrahs: The lifetime duration is the interesting value. To get it, the durations need to be merged into the longest non-overlapping interval to exclude redundant ranges when goroutines were running in parallel. Logging durations would truncate the information necessary to calculate that. The log timestamp is controlled by the logger which may be delayed.

This should ideally be a metric measurement rather than a log message, but that's for another day.

PR #480, opened 2023-07-12
generic filter for walker and faster impath

mrahs:
I originally opted for a deterministic ID for the filter by using a regexp. This turned out to be too restrictive and forces users to be creative with regular expressions where simpler Go expressions would suffice.

This delegates the responsibility of using a unique filter to the user, which is required for proper merkle tree processing in the casng uploader.

A quick profiling for revealed a non-trival amount of CPU time was spent on `filepath.Clean`, which is called by various functions in the same package. Building on the fact that paths constructed by the impath package are already clean, I have updated the package to try to avoid unnecessary cleaning.

A subsequent quick profiling shows much less time spent on cleaning paths. However, the cost might have been masked rather than eliminated, but that is a question for antother day.


go/pkg/io/impath/impath.go
```diff
@@ -141,22 +145,78 @@ func MustRel(elements ...string) Relative {
...
-	}
-	if strings.HasPrefix(p, "..") {
 		return zeroRel, errors.Join(ErrNotDescendant, fmt.Errorf("target %q is not a descendant of base %q", target, base))
 	}
-	return Relative{path: p}, nil
+	return Relative{path: rel}, nil
+}
+
+// fastDir assumes all paths are clean and avoids calling filepath.Clean.
+// Code is taken from filepath.Dir of go1.19.
```
bozydarsz, line 168: Cannot we reuse the code instead of copying the implementation?

go/pkg/io/impath/impath.go
```diff
@@ -141,22 +145,78 @@ func MustRel(elements ...string) Relative {
...
+	if dirty(path) {
+		path = filepath.Clean(path)
+	}
+	return path
+}
+
+// dirty return true if path contains consecutive separators or a dot elements (e.g. . or ..).
+func dirty(path string) bool {
+	dotsElm := true
+	lastRune := '\000'
```
bozydarsz, line 208: optional: maybe `prevRune` instead. One could think that last refers to the last character in path, not the one previously encountered.

go/pkg/io/impath/impath.go
```diff
@@ -141,22 +145,78 @@ func MustRel(elements ...string) Relative {
...
+		paths = append(paths, p.String())
+	}
+	path := strings.Join(paths, string(os.PathSeparator))
+	if dirty(path) {
+		path = filepath.Clean(path)
+	}
+	return path
+}
+
+// dirty return true if path contains consecutive separators or a dot elements (e.g. . or ..).
```
bozydarsz, line 205: nit: or dot elements (without `a`)

go/pkg/io/impath/impath.go
```diff
@@ -141,22 +145,78 @@ func MustRel(elements ...string) Relative {
...
+		return vol
+	}
+	if os.IsPathSeparator(dir[len(dir)-1]) {
+		dir = dir[:len(dir)-1]
+	}
+	return vol + dir
+}
+
+// clean returns a clean path from provided parts.
+func clean(base string, elements []Relative) string {
```
bozydarsz, line 187: why not just use `filepath.Clean`? Is it a perf optimization technique?

bozydarsz: Never mind, I've noticed that you already provided the answer in the PR comment.
Can you add the comment next to the function definition, mentioning that it's supposed to be lightweight version of `filepath.Clean`?

go/pkg/io/impath/impath.go
mrahs: I'm excluding the line that calls to `filepath.Clean`, which is the point of this optimization.

bozydarsz (approved):

PR #482, opened 2023-07-13
casng, scaffold upload path

mrahs:
This change is the first in a series to break down the upload code path. Subsequent changes will implement the methods in the order of the data flow.

### DIFF
--- a/go/pkg/casng/batching.go
+++ b/go/pkg/casng/batching.go
@@ -10,12 +10,21 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/contextmd"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/errors"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/io/impath"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/io/walker"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/symlinkopts"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 	log "github.com/golang/glog"
 	"github.com/klauspost/compress/zstd"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 )
+
+// namedDigest is used to conveniently extract the base name and its digest from repb.FileNode, repb.DirectoryNode and repb.SymlinkNode.
+type namedDigest interface {
+	GetName() string
+	GetDigest() *repb.Digest
+}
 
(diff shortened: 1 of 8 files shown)


go/pkg/casng/batching.go
```diff
@@ -10,13 +10,22 @@ import (
...
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/symlinkopts"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 	log "github.com/golang/glog"
 	"github.com/klauspost/compress/zstd"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
 )
 
+// namedDigest is used to conveniently extract the base name and its digest from repb.FileNode, repb.DirectoryNode and repb.SymlinkNode.
+type namedDigest interface {
+	GetName() string
```
bentekkie, line 25: Nit: `Name() string` https://google.github.io/styleguide/go/decisions#getters

go/pkg/casng/batching.go
```diff
@@ -10,13 +10,22 @@ import (
...
+	GetDigest() *repb.Digest
```
bentekkie, line 26: Nit: `Digest() *repb.Digest` https://google.github.io/styleguide/go/decisions#getters

go/pkg/casng/uploader.go
```diff
@@ -343,5 +424,40 @@ func (u *uploader) close() {
...
+		}
+
+		i++
+		log.Infof("[casng] beat; #%d, upload_subs=%d, query_subs=%d, walkers=%d, batching=%d, streaming=%d, querying=%d, open_files=%d, large_open_files=%d",
+			i, u.uploadPubSub.len(), u.queryPubSub.len(), u.walkThrottler.len(), u.uploadThrottler.len(), u.streamThrottle.len(), u.queryThrottler.len(), u.ioThrottler.len(), u.ioLargeThrottler.len())
+	}
+}
+
+func isExec(mode fs.FileMode) bool {
+	return mode&0100 != 0
```
bentekkie, line 462: Does this work on windows?

bentekkie (commented):
This is a fairly large cl I will finish reviewing tomorrow morning

bentekkie:
Generally it looks good though

go/pkg/casng/batching.go
mrahs: This has to match the name from the types I'm asserting to this interface, e.g. [`repb.FileNode`](https://github.com/bazelbuild/remote-apis/blob/39c174e10d224c46b556d8d4615863804d5b2ff6/build/bazel/remote/execution/v2/remote_execution.pb.go#L1158).

go/pkg/casng/uploader.go
mrahs: Probably not. However, if the upstream call is coming from the server, it should work since the bits are constructed from a boolean. If the upstream call is coming from stating the filesystem, the value doesn't matter because a subsequent client cannot (or shouldn't) make use of this bit because Windows doesn't have an executable bit.

PR #483, opened 2023-07-14
casng, implement public upload methods

mrahs:
Upload is the main method used to upload a single request, which may include a tree of files to upload.
The actual upload flow is defined in another file which will be submitted later.

DigestTree skips uploading to the server and returns a merkle tree root which can be useful for clients that wish to build upon that.

UploadTree covers the use case of having a list of inputs to upload which requires replicating a filesystem tree rooted at the common ancestor.
Using Upload with the root forces the user to craft an exclusion filter that ensures only desired files are included. However, in build systems the list of included files is readily available.
This method reconstructs the tree in memory without forcing the user to recreate it on the filesystem (i.e. copy the files into an isoalted root).


go/pkg/casng/batching.go
```diff
@@ -314,12 +318,113 @@ func (u *uploader) writeBytes(ctx context.Context, name string, r io.Reader, siz
...
+		return digest.Digest{}, Stats{}, ctx.Err()
+	}
+
+	var stats Stats
+	var err error
+	for r := range resCh {
+		stats.Add(r.Stats)
+		err = errors.Join(r.Err, err)
+	}
+	rootNode := u.Node(req)
```
bentekkie, line 421: `u.Node(req)` returns the interface `proto.Message` which will never be nil right? You probably need to do `!u.Node(req).ProtoReflect().IsValid()` instead of `rootNode == nil` or try to cast to each of the possible concrete types?

go/pkg/casng/batching.go
```diff
@@ -329,5 +434,216 @@ func (u *BatchingUploader) DigestTree(ctx context.Context, root impath.Absolute,
...
+	// 3rd, Compute the shared ancestor nodes and upload them.
+
+	// This block creates a flattened tree of the paths in reqs rooted at execRoot.
+	// Each key is an absolute path to a node in the tree and its value is a list of absolute paths that any of them can be a key as well.
+	// Example: /a: [/a/b /a/c], /a/b: [/a/b/foo.go], /a/c: [/a/c/bar.go]
+	dirChildren := make(map[impath.Absolute]map[impath.Absolute]proto.Message, len(pathSeen))
+	for _, r := range reqs {
+		// Each request in reqs must correspond to a cached node.
+		node := u.Node(r)
+		if node == nil {
```
bentekkie, line 519: Same as above about the nil check

go/pkg/casng/batching.go
```diff
@@ -329,5 +434,216 @@ func (u *BatchingUploader) DigestTree(ctx context.Context, root impath.Absolute,
...
+	}
+	for len(stack) > 0 {
+		// Peek.
+		dir := stack[len(stack)-1]
+
+		// Depth first.
+		children := dirChildren[dir]
+		pending := false
+		for child, node := range children {
+			if node == nil {
```
bentekkie, line 569: Same as above regarding the nil check

go/pkg/casng/batching.go
```diff
@@ -329,5 +434,216 @@ func (u *BatchingUploader) DigestTree(ctx context.Context, root impath.Absolute,
...
+			rp = parent
+			parent = parent.Dir()
+			children := dirChildren[parent]
+			if children == nil {
+				children = make(map[impath.Absolute]proto.Message, len(reqs)) // over-allocating to avoid copying on growth.
+				dirChildren[parent] = children
+			}
+			// If the parent already has this child, then no need to traverse up the tree again.
+			_, ancestrySeen := children[rp]
+			children[rp] = node
```
bentekkie, line 542: Is it intended that this is nil for every loop after the first one?

go/pkg/casng/batching.go
mrahs: Since the method is called from a sibling one on the same struct, I'm trusting the implementation to return either `nil` (not found) or a valid struct. Verbose guards seem an overkill considering this is technically internal (to the struct) code.

go/pkg/casng/batching.go
```diff
@@ -329,5 +434,216 @@ func (u *BatchingUploader) DigestTree(ctx context.Context, root impath.Absolute,
...
+			return
+		}
+
+		// Add this leaf node to its ancestors.
+		rpStr := strings.TrimPrefix(r.Path.String(), execRoot.String()+string(filepath.Separator))
+		if strings.HasPrefix(rpStr, workingDir.String()) {
+			rpStr = strings.Replace(rpStr, workingDir.String(), remoteWorkingDir.String(), 1)
+		}
+		rp := execRoot.Append(impath.MustRel(rpStr))
+		// rp := remotePath[r.Path]
```
mrahs, line 530: Remove commented code.

go/pkg/casng/batching.go
mrahs: Yes. This looks more concise than using a boolean.

go/pkg/casng/batching.go
bentekkie: That is reasonable. Maybe add to the docs on Node() that it will either return a valid struct or nil. That can be in one of the follow up changes though

mrahs: I thought I already did: https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/casng/uploader.go#L224-L227
What can I add to improve it?

go/pkg/casng/batching.go
```diff
@@ -314,12 +318,113 @@ func (u *uploader) writeBytes(ctx context.Context, name string, r io.Reader, siz
...
+			select {
+			case ch <- r:
+			case <-u.ctx.Done():
+				return
+			}
+		}
+	}()
+
+	var uploaded []digest.Digest
+	for r := range resCh {
```
bozydarsz, line 385: How is `resCh` closed?

go/pkg/casng/batching.go
```diff
@@ -314,12 +318,113 @@ func (u *uploader) writeBytes(ctx context.Context, name string, r io.Reader, siz
...
+		if r.Err != nil {
+			err = errors.Join(r.Err, err)
```
bozydarsz, line 387: There's no streamPipe implementation yet, so I'm not sure whether it's doable or even make sense, but wouldn't it be better to return an error as soon as we got one instead of processing the rest of the requests?

My concern is, if RBE starts to timeout (e.g. because it's not reachable or overloaded, or because of misconfigured endpoint) we might end up waiting `request_timeout * len(reqs)` to get the response instead of `request_timeout` if we returned the error just after one failed request.

go/pkg/casng/batching.go
```diff
@@ -329,5 +434,215 @@ func (u *BatchingUploader) DigestTree(ctx context.Context, root impath.Absolute,
...
+	// Updating the filterID for the set to a deterministic one ensures it gets its unique keys that are still shared between identical sets.
+	// The deterministic ID is the digest of the sorted list of paths concatenated with the ID of the original filter.
+	filter := reqs[0].Exclude
+	filterID := filter.String()
+	paths := make([]string, 0, len(reqs)+1)
+	paths = append(paths, filterID)
+
+	// An ancestor directory takes precedence over all of its descendants to ensure unlisted files are also included in traversal.
+	// Sorting the requests by path length ensures ancestors appear before descendants.
+	sort.Slice(reqs, func(i, j int) bool { return len(reqs[i].Path.String()) < len(reqs[j].Path.String()) })
```
bozydarsz, line 456: just checking: are the paths normalized? e.g. is there no chance that the request would contain `..` component in the path?

go/pkg/casng/batching.go
mrahs: `streamPipe` takes an input channel `ch` and returns an output channel `resCh`. The caller (here) owns the input channel while `streamPipe` owns the output channel. When `ch` is closed (line 369), `streamPipe` will close `resCh` after it's done processing messages received on `ch`.

go/pkg/casng/batching.go
mrahs: The requests go through multiple workers downstream. Such errors would be somewhere in the middle and by the time the request arrives here, potentially multiple requests would already be within the pipeline. If the server is not responding leaving the connection to timeout, that would be considered a retryable error. Right now the retry logic is per-call (may be multiple calls per request), which is not too bad.

If a broadcast cancellation is desired, that could be implemented using the streaming interface of this package which I'm going to submit later. Using that interface, the user gets responses as soon as they are received without internal batching, allowing the user to cancel the call and abort in-flight requests.

go/pkg/casng/batching.go
```diff
@@ -314,12 +318,113 @@ func (u *uploader) writeBytes(ctx context.Context, name string, r io.Reader, siz
...
+	case <-ctx.Done():
+		close(ch)
+		return digest.Digest{}, Stats{}, ctx.Err()
+	}
+
+	var stats Stats
+	var err error
+	for r := range resCh {
+		stats.Add(r.Stats)
+		err = errors.Join(r.Err, err)
```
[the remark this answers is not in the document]
mrahs: Good catch! I must have had those two versions in-between modifications to the `errors` package (this method is more recent than the one with the guard). `errors.Join` already excludes `nil` errors. I shall update the other site.

go/pkg/casng/batching.go
mrahs: Yes. `impath` package ensures every path it creates is clean and immutable.

go/pkg/casng/batching.go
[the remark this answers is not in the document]
mrahs: On second thought, I did a sweep and noticed I'm guarding `Join` almost everywhere. In most sites, it makes the code easier to read than without the guard (which makes it look like an error is expected).

bozydarsz (approved):

PR #484, opened 2023-07-18
Add KeepAlive support

bozydarsz:
KeepAlive functionality sends periodic pings to the server over the transport layer.

### DIFF
--- a/go/pkg/flags/flags.go
+++ b/go/pkg/flags/flags.go
@@ -9,6 +9,10 @@ import (
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/balancer"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/client"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/moreflag"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/keepalive"
+
+	log "github.com/golang/glog"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


go/pkg/flags/flags.go
```diff
@@ -64,6 +68,12 @@ var (
 	StartupCapabilities = flag.Bool("startup_capabilities", true, "Whether to self-configure based on remote server capabilities on startup.")
 	// RPCTimeouts stores the per-RPC timeout values.
 	RPCTimeouts map[string]string
+	// KeepAliveTime specifies gRPCs keepalive time parameter.
+	KeepAliveTime = flag.Duration("grpc_keepalive_time", 0*time.Second, "After a duration of this time if the client doesn't see any activity it pings the server to see if the transport is still alive. If zero or not set, the mechanism is off.")
```
mrahs, line 72: Is the default value set like that to emphasize the unit? If so, I think that would be more useful in the help text where it would displayed to the user and the developer can still read it here.

Also, worth noting that the default it 0.

go/pkg/flags/flags.go
```diff
@@ -103,12 +113,23 @@ func NewClientFromFlags(ctx context.Context, opts ...client.Opt) (*client.Client
 		}
 	}
 
+	dialOpts := make([]grpc.DialOption, 0)
+	if *KeepAliveTime > 0*time.Second {
```
mrahs, line 117: Same here. Is it to show the unit? Comparison with 0 is unit agnostic though.

mrahs (approved):

PR #486, opened 2023-07-19
casng, implement digestion methods

mrahs:
Each method accepts a filepath and produces a digested node, which the worker then forwards to the dispatcher to be dispatched for uploading.

The file digestion method also handles acquiring and releasing IO tokens to respect resource quotas. One particular nuance here is the handling of tokens for large files. Such tokens are acquired in the digester here but are not released until later in another worker. This leverages file locality when processing many large files, such that the file is opened once for both digestion and streaming to the remote server.

The tradeoff here is more susceptibility to bugs where the tokens are not released correctly. For example, adding a return statement and forgetting to update the position of the release statement.

To mitigate such risks, log messages are added to report on the total number of tokens periodically (logBeat) and per request.


go/pkg/casng/upload_digester.go
```diff
@@ -129,7 +138,220 @@ func (u *uploader) digester() {
...
+				return walker.SkipPath, false
+			// This is intended to short-circuit a cancelled uploader. It is not intended to abort by cancelling the uploader's context.
+			case <-u.ctx.Done():
+				log.V(3).Infof("upload.digester.cancel; tag=%s, walk=%s", req.tag, walkID)
+				return walker.SkipPath, false
+			default:
+			}
+
+			// A cache hit here indicates a cyclic symlink with the same requester or multiple requesters attempting to upload the exact same path with an identical filter.
+			// In both cases, deferring is the right call. Once the requset is processed, all requestters will revisit the path to get the digestion result.
```
bozydarsz, line 174: a typo in request and requesters

go/pkg/casng/upload_digester.go
```diff
@@ -129,7 +138,220 @@ func (u *uploader) digester() {
...
+			// If the path was not cached before, claim it by makring it as in-flight.
```
bozydarsz, line 175: a typo in `marking`?

go/pkg/casng/upload_digester.go
```diff
@@ -129,7 +138,220 @@ func (u *uploader) digester() {
...
+				}
+				log.V(3).Infof("[casng] upload.digester.visit.wait; path=%s, real_path=%s, req=%s, tag=%s, walk=%s", path, realPath, req.id, req.tag, walkID)
+				startTime := time.Now()
+				wg.Wait()
+				log.V(3).Infof("[casng] upload.digester.visit.wait.duration; start=%d, end=%d, path=%s, real_path=%s, req=%s, tag=%s, walk=%s", startTime.UnixNano(), time.Now().UnixNano(), path, realPath, req.id, req.tag, walkID)
+				delete(deferredWg, key)
+				return walker.Defer, true
+			}
+
+			node := m.(proto.Message) // Guaranteed assertion because the cache is an internal field.
```
bozydarsz, line 201: optional: is it possible to avoid reusing the cache for different value types? It should make it easier to grasp and make more maintainable

mrahs: I assume you are referring to `sync.WaitGroup` and `proto.Message`. I opted to reusing the cache given it already has a generic type for the value. For readers, it's either the desired value, a empty marker, or a reserved marker. In the last two cases, the reader can reserve the key or try to read again later, respectively.

It's possible to have a separate cache for reservations, but I think it's going to increase the amount of code and memory since there would be two nil checks and two placeholders: one for the reservation cache to say this is computed, and another in the node cache for the actual value. This would also introduce an edge case (internal bug) where the reservation cache says it's computed but the node cache says it's nil.

I think it might be worth exploring further as a possible improvement in another PR.

bozydarsz (approved):

PR #488, opened 2023-07-20
casng, implemented gRPC batching worker

mrahs:
The gRPC batcher handles the batching API of the CAS. It unifies duplicate blobs from multiple requesters into a single request to the remote server, and returns the response to all related requesters.

The unification parameters are configurable and described in the configs.go file as well as in the godocs of public methods.

Dispatching based on byte-size attempts to be as accurate as possible. The bytes are padded by the overhead of serialized proto structures rather than using an arbitrary approximation.

The retry loop is per batch, and only retries failed requests within a batch.

### DIFF
--- a/go/pkg/casng/streaming_upload.go
+++ b/go/pkg/casng/streaming_upload.go
@@ -2,14 +2,21 @@ package casng
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"io/fs"
+	"os"
+	"strings"
+	"time"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/contextmd"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/errors"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/io/impath"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/io/walker"
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/retry"
 	"github.com/pborman/uuid"
+	"google.golang.org/grpc/status"
 
 	slo "github.com/bazelbuild/remote-apis-sdks/go/pkg/symlinkopts"
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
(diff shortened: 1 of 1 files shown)


go/pkg/casng/streaming_upload.go
```diff
@@ -188,10 +195,278 @@ func (u *uploader) streamPipe(ctx context.Context, in <-chan UploadRequest) <-ch
...
+		go func(ctx context.Context, b uploadRequestBundle) {
+			defer u.workerWg.Done()
+			defer u.uploadThrottler.release()
+			// TODO: cancel ctx if all requesters have cancelled their contexts.
+			u.callBatchUpload(ctx, b)
+		}(ctx, bundle)
+
+		bundle = make(uploadRequestBundle)
+		bundleSize = u.uploadRequestBaseSize
+		ctx = u.ctx
```
bozydarsz, line 236: why assigning to `u.ctx` again?

go/pkg/casng/streaming_upload.go
```diff
@@ -188,10 +195,278 @@ func (u *uploader) streamPipe(ctx context.Context, in <-chan UploadRequest) <-ch
...
 // uploadBatcher handles files that can fit into a batching request.
 func (u *uploader) batcher() {
+	log.V(1).Info("[casng] upload.batcher.start")
+	defer log.V(1).Info("[casng] uploader.batcher.stop")
+
+	bundle := make(uploadRequestBundle)
+	bundleSize := u.uploadRequestBaseSize
+	ctx := u.ctx // context with unified metadata.
+
+	handle := func() {
```
bozydarsz, line 205: maybe extract as a separate function for readability and future maintainability. same would apply to other anonymous functions within `batcher` function

go/pkg/casng/streaming_upload.go
mrahs: To reset `ctx` now that the associated bundle has been dispatched with a copy of it.

go/pkg/casng/streaming_upload.go
mrahs: Defining this function as a closure allows sharing read/write access to the three variables above. This allows me to reset the bundle as part of handling it. If this was a separate function, I'd have to repeat the reset code after every call to this function, which introduces another point of failure if a future change missed keeping the two calls tightly coupled everywhere.

I'll add a comment above this function.

bozydarsz (approved):

PR #495, opened 2023-08-02
casng, glue into the client package

mrahs:
This PR also includes some changes for debugging purposes. Mainly, capturing the tree from the client package to help compare it with the one generated by casng. However, these are guarded behind verbosity level 5 to avoid the compute and memory cost.


go/pkg/client/tree.go
```diff
@@ -360,15 +379,27 @@ func buildTree(files map[string]*fileSysNode) (*treeNode, error) {
 	return root, nil
 }
 
-func packageTree(t *treeNode, stats *TreeStats) (root digest.Digest, blobs map[digest.Digest]*uploadinfo.Entry, err error) {
+// If tree is not nil, it will be populated with a flattened tree of path->digest.
+// prefix should always be provided as an empty string which will be used to accumolate path prefixes during recursion.
```
bentekkie, line 383: Wouldnt a better pattern be to have a non recursive top level 
`func packageTree(t *treeNode, stats *TreeStats, tree map[string]digest.Digest) (root digest.Digest, blobs map[digest.Digest]*uploadinfo.Entry, err error)`
that calls a recursive
`func packageTreeWithPrefix(t *treeNode, stats *TreeStats, prefix string, tree map[string]digest.Digest) (root digest.Digest, blobs map[digest.Digest]*uploadinfo.Entry, err error)`

go/pkg/rexec/rexec.go
```diff
@@ -216,6 +228,271 @@ func (ec *Context) computeInputs() error {
...
+	// Sorting by path length descending is necessary to skip redundant ancestors. Otherwise, the conslidation in casng.UploadTree will skip descendants.
+	sort.Slice(ec.cmd.InputSpec.VirtualInputs, func(i, j int) bool {
+		return len(ec.cmd.InputSpec.VirtualInputs[i].Path) > len(ec.cmd.InputSpec.VirtualInputs[j].Path)
+	})
+	for _, p := range ec.cmd.InputSpec.VirtualInputs {
+		if p.Path == "" {
+			return fmt.Errorf("[casng] ng.req: empty virtual path; cmd_id=%s, exec_id=%s", cmdID, executionID)
+		}
+		// If execRoot is a virtual path, ignore it.
+		if p.Path == "." || p.Path == "" {
```
bentekkie, line 282: This is not necessary because if `p.Path == ""` then it will return an error above

go/pkg/rexec/rexec.go
```diff
@@ -216,6 +228,271 @@ func (ec *Context) computeInputs() error {
...
+	}
+	slPreserve := treeOpts.Preserved
+	switch cmdOpts {
+	case command.ResolveSymlink:
+		slPreserve = false
+	case command.PreserveSymlink:
+		slPreserve = true
+	}
+
+	switch {
```
bentekkie, line 410: Would be cleaner to have return statements inside the cases

go/pkg/client/tree.go
mrahs: It's usually useful to simplify the public interface. Considering this is internal and actually have a single call site, I thought it would be verbose for no real benefit.

bentekkie: That's fair

bentekkie (approved):

PR #503, opened 2023-08-30
Copy NodeProperties instead of importing into command.proto

bentekkie:
On builds with high parallelism the import of remote_execution.proto ends up causing resource exhaustion issues.

The root cause is unknown at this time but modifying the proto to define a copy instead fixes the issues.


go/api/command/command.proto
```diff
@@ -126,7 +126,31 @@ message InputSpec {
   SymlinkBehaviorType.Value symlink_behavior = 6;
 
   // Node properties of inputs.
-  map<string,build.bazel.remote.execution.v2.NodeProperties> input_node_properties = 7;
+  map<string,NodeProperties> input_node_properties = 7;
+}
+
+// A copy of NodeProperties from https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto
+// to importing it as a proto dependency
```
Jwata, line 133: Is it a worth mentioning the issue on the high spec machine? 
Is it ok to put http://b/296368945 here?
At least, this build failure is public. https://luci-milo.appspot.com/ui/p/chromium/builders/ci/linux-build-perf-developer/3291/overview

Jwata (approved):

go/api/command/command.proto
Jwata: This comment has been addressed.

bozydarsz (approved):

PR #505, opened 2023-09-09
casng, fix resolving remote paths

mrahs:
A few corner cases were not handeled properly.
The original one was not removing the leading slash after stripping
workingDir.
However, a more interesting one was that paths being outside workingDir,
but still under root, which could translate to partial replacements of
the workingDir path, which is a valid case, contrary to what I thought.
Reverted to using proper filepath APIs.

### DIFF
--- a/go/pkg/casng/batching.go
+++ b/go/pkg/casng/batching.go
@@ -530,19 +530,11 @@ func (u *BatchingUploader) UploadTree(ctx context.Context, execRoot impath.Absol
 
 		// Add this leaf node to its ancestors.
 		// Start by swapping the working directory with the remote one.
-		// Using strings is faster than using filepath functions which use filepath.Clean which is not cheap.
-		// To properly handle corner cases (no working dir, root is '/', etc), this is done in separate steps.
-		rpStr := strings.TrimPrefix(r.Path.String(), execRoot.String()) // if execRoot=='/', rpStr is now relative
-		rpStr = strings.TrimPrefix(rpStr, string(filepath.Separator))   // if execRoot!='/', remove the path separator
-		if strings.HasPrefix(rpStr, workingDir.String()) {
-			rpStr = strings.Replace(rpStr, workingDir.String(), remoteWorkingDir.String(), 1)
-		}
-		rpStrRel, errIm := impath.Rel(rpStr)
+		rp, errIm := ReplaceWorkingDir(r.Path, execRoot, workingDir, remoteWorkingDir)
 		if errIm != nil {
 			err = errors.Join(fmt.Errorf("[casng] upload.tree; cannot construct the merkle tree with a path outside the root: path=%q, root=%q, working_dir=%q, remote_working_dir=%q", r.Path, execRoot, workingDir, remoteWorkingDir), errIm)
 			return
 		}
-		rp := execRoot.Append(rpStrRel)
 		parent := rp
 		for {
 			rp = parent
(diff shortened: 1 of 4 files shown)


bozydarsz (approved):

go/pkg/casng/batching.go
```diff
@@ -665,3 +657,26 @@ func (u *BatchingUploader) UploadTree(ctx context.Context, execRoot impath.Absol
 
 	return
 }
+
+func ReplaceWorkingDir(path, root impath.Absolute, workingDir, remoteWorkingDir impath.Relative) (impath.Absolute, error) {
```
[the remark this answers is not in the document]
mrahs: I wanted to, but the unit test is in another package. The value of testing this function outweighed the downside of exporting it. I also added godocs and a couple explicit checks and errors.

go/pkg/casng/batching.go
```diff
@@ -665,3 +657,38 @@ func (u *BatchingUploader) UploadTree(ctx context.Context, execRoot impath.Absol
...
+	if !strings.HasPrefix(path.String(), root.String()) {
+		return impath.Absolute{}, fmt.Errorf("cannot replace working dir for path %q because it is not prefixed by %q", path, root)
+	}
+
+	p := strings.TrimPrefix(path.String(), root.String()) // if root == '/', p is now relative
+	// if root isn't the system root, there would be a leading path separator.
+	p = strings.TrimPrefix(p, string(filepath.Separator))
+
+	if !strings.HasPrefix(p, workingDir.String()) {
+		log.Warningf("cannot replace working dir %q for path %q (dervied from %q) because it is not part of it", workingDir, p, path)
```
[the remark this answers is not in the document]
bozydarsz: maybe produce it only with a higher verbosity?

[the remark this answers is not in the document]
mrahs: Thanks for pointing the case where paths need to be under `execRoot`, but not necessarily under `workingDir` or `remoteWorkingDir`.

go/pkg/casng/batching.go
```diff
@@ -665,3 +657,27 @@ func (u *BatchingUploader) UploadTree(ctx context.Context, execRoot impath.Absol
 
 	return
 }
+
+// ReplaceWorkingDir swaps remoteWorkingDir for workingDir in path which must be prefixed by root.
+// workingDir is assumed to be prefixed by root, and the returned path will be a descendent of root, but necessarily a decendent of remoteWorkingDir.
```
bozydarsz, line 662: did you mean "(...) not necessarily a descendent"?

also, I think a noun is descendant (with "a")

go/pkg/casng/batching.go
```diff
@@ -665,3 +657,27 @@ func (u *BatchingUploader) UploadTree(ctx context.Context, execRoot impath.Absol
...
+// Example: path=/root/out/foo.c, root=/root, workdingDir=out/reclient, remoteWorkingDir=set_by_reclient/a/b, result=/root/set_by_reclient/a/foo.c
```
bozydarsz, line 663: Is that fact that  remoteWorkingDir=`set_by_reclient/a/b` but result `/root/set_by_reclient/a/foo.c` (without `/b`) on purpose or is it a typo?

mrahs: On purpose. The unit test has a case for this one. The path resolves to `../foo.c` relative to `out/reclient`, which is then joined to `set_by_reclient_/a/b/../foo.c` which normalizes to `set_by_reclient/a/foo.c`.

go/pkg/casng/batching.go
mrahs: Last I checked, both forms of the word can be used interchangeably. But yes, I missed a "not".

PR #507, opened 2023-09-28
[go] Add support for authentication via Credential Helper

Yannic:
This change adds support for Credential Helpers, similar to Bazel's support for Credential Helpers.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/bazelbuild/remote-apis-sdks
 go 1.20
 
 require (
+	github.com/EngFlow/credential-helper-go v0.0.0-20230928231836-1ae22fe3da4b
 	github.com/bazelbuild/remote-apis v0.0.0-20230411132548-35aee1c4a425
 	github.com/golang/glog v1.1.0
 	github.com/google/go-cmp v0.5.9
@@ -17,7 +18,7 @@ require (
 	google.golang.org/genproto v0.0.0-20230803162519-f966b187b2e5
 	google.golang.org/genproto/googleapis/bytestream v0.0.0-20230807174057-1744710a1577
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230731190214-cbb8c96f2d6d
-	google.golang.org/grpc v1.58.0-dev.0.20230804151048-7aceafcc52f9
+	google.golang.org/grpc v1.58.0
 	google.golang.org/protobuf v1.31.0
 )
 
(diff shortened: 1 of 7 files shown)


Yannic:
cc @ola-rozenfeld

gkousik:
Hi Yannic, thanks for the PR!

Is the goal to eventually use this credentials from reclient? If so, could you instead use the `ExternalAuthToken` option? We would ideally like the SDK interface to remain generic.

Also, as an FYI, we are also working on credentials helper support in reclient and are designing the interface to be generic (should be merged in a few weeks).

Yannic:
Yes, Credential Helper support for `reclient` is the primary motivation here. But also being able to use Credential Helpers for the tools from this repo which we sometimes use for debugging. I think having Credential Helpers here could also allow us to remove support for some other auth mechanisms since credential helpers are more generic (that's also a goal for Bazel to eventually remove them all and provide credential helpers for .netrc, GCP, ...). WDYT?

Looking forward to seeing Credential Helpers in `reclient` :)

gkousik:
Ah goot point about `remotetool`. For usage of creds helper for remotetool, could it be added as a flag in https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/cmd/remotetool/main.go, then passed in via `ExternalAuthToken` flag to the core SDK?

ola-rozenfeld:
I suggest we sync up on the plan for generalizing the credential helper support, because it looks to me like there has been some parallel work happening on this front, and deduplicating it / getting on the same page will be useful. @Yannic 's plan was to use the [Bazel credential helper](https://groups.google.com/g/bazel-discuss/c/mwLotKlERh4) ([design](https://github.com/bazelbuild/proposals/blob/main/designs/2022-06-07-bazel-credential-helpers.md))and basically to bring the SDK to parity with Bazel in that regard. Is that what you've been working on as well?

If you guys decide that a VC will be best to hash this all out, I'd love to take part as well.

gkousik:
An update: I have a doc written and its being reviewed internally. Once we have agreement, I'll clean it up and share the interface externally (so a few more days).

gkousik:
I have opened a discussion proposal here https://github.com/bazelbuild/reclient/discussions/16 for credentials helper support.
The TL;DR of that proposal is we'll add credentials helper flags along with the authentication flags in the SDK and support it very similar to how Bazel does (but with support for two extra fields).

daniel-sudz:
@gkousik what is the current state of this?

gkousik:
@daniel-sudz We are currently implementing this in the SDK repository!

PR #509, opened 2023-10-02
Fix stats in BatchDownloadBlobs call

gkousik:
Previously, BatchDownloadBlobs() wasn't properly recording the compressed download size. To fix it, I've introduced a new public interface called BatchDowloadBlobsWithStats(), which will return statistics on the downloaded blob size in addition to the actual content.

As part of this, I also realized that tests are non-existent for compressed BatchDownloadBlobs() and added them.

In a subsequent change, I'll fix BatchUploadBlobs() as well.

TESTED: Unit test

### DIFF
--- a/go/pkg/client/capabilities.go
+++ b/go/pkg/client/capabilities.go
@@ -46,7 +46,7 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
 		}
 		for _, compressor := range c.serverCaps.CacheCapabilities.SupportedBatchUpdateCompressors {
 			if compressor == repb.Compressor_ZSTD {
-				c.batchCompression = true
+				c.useBatchCompression = UseBatchCompression(true)
 			}
 		}
 	}
--- a/go/pkg/client/cas_download.go
+++ b/go/pkg/client/cas_download.go
@@ -186,16 +186,19 @@ func (c *Client) DownloadDirectory(ctx context.Context, d digest.Digest, outDir 
 // zstdDecoder is a shared instance that should only be used in stateless mode, i.e. only by calling DecodeAll()
 var zstdDecoder, _ = zstd.NewReader(nil)
 
-// BatchDownloadBlobs downloads a number of blobs from the CAS to memory. They must collectively be below the
-// maximum total size for a batch read, which is about 4 MB (see MaxBatchSize). Digests must be
-// computed in advance by the caller. In case multiple errors occur during the blob read, the
-// last error will be returned.
-func (c *Client) BatchDownloadBlobs(ctx context.Context, dgs []digest.Digest) (map[digest.Digest][]byte, error) {
+// CompressedBlobInfo is primarily used to store stats about compressed blob size
+// in addition to the actual blob data.
+type CompressedBlobInfo struct {
+	CompressedSize int64
+	Data           []byte
+}
+
+func (c *Client) BatchDownloadBlobsWithStats(ctx context.Context, dgs []digest.Digest) (map[digest.Digest]CompressedBlobInfo, error) {
 	if len(dgs) > int(c.MaxBatchDigests) {
 		return nil, fmt.Errorf("batch read of %d total blobs exceeds maximum of %d", len(dgs), c.MaxBatchDigests)
 	}
 	req := &repb.BatchReadBlobsRequest{InstanceName: c.InstanceName}
-	if c.batchCompression {
+	if c.useBatchCompression {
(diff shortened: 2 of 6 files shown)


mrahs:
Is the granularity warranted or would following the same pattern in similar methods (returning a `MovedBytesMetadata` instance with aggregate stats) be sufficient?

gkousik:
> Is the granularity warranted or would following the same pattern in similar methods (returning a MovedBytesMetadata instance with aggregate stats) be sufficient?

Unfortunately, it seems like I do need this particular method to return granular stats. The `BatchDownloadBlobsWithStats()` is in turn called by the `downloadBatch()` function of the unified uploader / downloader, which attaches granular stats about compression for each digest and returns it on the response channel (https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/client/cas_download.go#L736).

If I aggregate the stats in `BatchDownloadBlobsWithStats()` function, I won't be able to subsequently provide granular info there. But it does look like all these get aggregated eventually into `MovedBytesMetadata`.

go/pkg/client/capabilities.go
```diff
@@ -46,7 +46,7 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
 		}
 		for _, compressor := range c.serverCaps.CacheCapabilities.SupportedBatchUpdateCompressors {
 			if compressor == repb.Compressor_ZSTD {
-				c.batchCompression = true
+				c.BatchCompression = true
```
mrahs, line 49: It looks like this was made public to be accessible from the test package? If so, the existing patter is define public configs as client options, i.e. https://github.com/bazelbuild/remote-apis-sdks/blob/2ab761126c7ac5a597f5a93acd53acd2c028e97d/go/pkg/client/client.go#L270

go/pkg/client/cas_test.go
```diff
@@ -1867,3 +1867,51 @@ func TestDownloadFilesCancel(t *testing.T) {
...
+
+func TestBatchDownloadBlobsCompressed(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("Cannot listen: %v", err)
+	}
+	fakeCAS := fakes.NewCAS()
+	defer listener.Close()
```
mrahs, line 1879: `t.Cleanup` is probably more suited than `defer` in tests.

go/pkg/client/cas_download.go
```diff
@@ -727,20 +749,23 @@ func (c *Client) download(ctx context.Context, data []*downloadRequest) {
...
+	bchMap, err := c.BatchDownloadBlobsWithStats(ctx, batch)
 	if err != nil {
 		afterDownload(batch, reqs, map[digest.Digest]*MovedBytesMetadata{}, err)
 		return
 	}
 	for _, dg := range batch {
+		bi := bchMap[dg]
+		if bi == nil {
+			bi = &CompressedBlobInfo{}
+		}
```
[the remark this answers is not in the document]
gkousik: Done

go/pkg/client/capabilities.go
gkousik: Thanks for the context too, good point. Done!

go/pkg/client/cas_test.go
gkousik: https://stackoverflow.com/questions/61609085/what-is-useful-for-t-cleanup

Sounds like t.Cleanup is useful when the test calls other functions (which doesn't happen here). In addition, all other tests here also just defer the cleanup function. I would rather it stay like the other tests and fix them all in a separate change.

mrahs: Fair point.

go/pkg/client/capabilities.go
```diff
@@ -46,7 +46,7 @@ func (c *Client) CheckCapabilities(ctx context.Context) (err error) {
 		}
 		for _, compressor := range c.serverCaps.CacheCapabilities.SupportedBatchUpdateCompressors {
 			if compressor == repb.Compressor_ZSTD {
-				c.batchCompression = true
+				UseBatchCompression(true).Apply(c)
```
mrahs, line 49: Nit: `c.batchCompression = UseBatchCompresion(true)` looks more idiomatic.

mrahs (approved):

go/pkg/client/capabilities.go
gkousik: Done.

PR #514, opened 2023-10-24
fix preserving intermediate symlinks

mrahs:
This patch fixes an oversight in the previous PR #511 where an intermediate symlink is processed just like an input which means all of its tree is also included even though it is not part of the specified input.

That not only inflated the Merkle tree, but also included symlinks to targets outside the execution root which caused failures in builds that do not specify the materialization option.

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -234,10 +234,11 @@ func loadFiles(execRoot, localWorkingDir, remoteWorkingDir string, excl []*comma
 		opts = DefaultTreeSymlinkOpts()
 	}
 
-	var seenRelPaths map[string]struct{}
+	var seenAncestorSymlinkRel map[string]bool
 	if opts.Preserved {
-		// Remember loaded files to avoid reprocessing shared symlink ancestors.
-		seenRelPaths = make(map[string]struct{}, len(filesToProcess))
+		// Remember ancestor symlinks to avoid redundant processing for shared ancestors.
+		// This ensures that an ancestor symlink is only ever added once to the list.
+		seenAncestorSymlinkRel = make(map[string]bool, len(filesToProcess))
 	}
 	for len(filesToProcess) != 0 {
 		path := filesToProcess[0]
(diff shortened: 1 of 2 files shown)


go/pkg/client/tree.go
```diff
@@ -308,9 +309,17 @@ func loadFiles(execRoot, localWorkingDir, remoteWorkingDir string, excl []*comma
...
-				filesToProcess = append(filesToProcess, targetExecRoot)
+				// Do not follow target if this symlink was only an ancestor and not an expclicit input.
+				// Zero or actual value translates to: if never seen or seen but already processed.
+				if !seenAncestorSymlinkRel[normPath] {
+					// getTargetRelPath validates this target is under execRoot,
+					// and the iteration loop will get the relative path to execRoot,
+					filesToProcess = append(filesToProcess, targetExecRoot)
+				} else {
+					// If this symlink was also an explicit input, it should appear more than once in the list.
+					// Setting its value to false allows subsequent (unseen) copies to be correctly processed as expclicit inputs.
```
ramymedhat, line 320: explicit

go/pkg/client/tree_test.go
```diff
@@ -1038,7 +1039,7 @@ func TestComputeMerkleTree(t *testing.T) {
 			wantCacheCalls: map[string]int{
 				"foobarDir/foo": 2,
 				"foobarDir/bar": 2,
-				"foobarDir":     5, // 2 as an ancestor to input files, 1 as input, 2 as an ancestor to nested files.
+				"foobarDir":     5, // 2 as input ancestor, 1 as input, 2 as nested input ancestor
```
ramymedhat, line 1042: Not sure I understand why this is not 3?

go/pkg/client/tree_test.go
```diff
@@ -1071,8 +1072,8 @@ func TestComputeMerkleTree(t *testing.T) {
 			wantCacheCalls: map[string]int{
 				"foobarDir/foo": 2, // 1 via the dir and 1 via the symlink
 				"foobarDir/bar": 2,
-				"foobarDir":     3, // 1 as input, 2 as a real ancestor
-				"foobarSymDir":  3, // 1 as input, 2 as a symlink ancestor
+				"foobarDir":     3, // 1 as target of the symlink, 2 as input ancestor
+				"foobarSymDir":  4, // 1 as input, 1 as added symlink ancestor, 2 as input ancestor
```
ramymedhat, line 1076: nit: some tests use foobarSymDir->foobarDir and some use foobarDir->foobarDirOrig. Can we make it consistent?

mrahs: It's a compromise to avoid creating more global digests in this test. Basically this allows `foobarDir` the symlink to materialize as a directory whose digest is already known.

mrahs (commented):
After a couple of attempts to refactor `loadFiles` to reduce its complexity, I didn't arrive at a satisfactory result. I may need to widen the scope of the refactor which I'd like to avoid.

Instead, I made a small update to make it clearer when the symlink target should not be followed regardless of the option. I think it's a good compromise for now.

ramymedhat (approved):

PR #515, opened 2023-10-25
Make download_action command write a script to execute the action loc…

gkousik:
…ally

This will be useful if the developer just wants to run the action within the docker container on their machine after downloading the action inputs

### DIFF
--- a/go/pkg/tool/tool.go
+++ b/go/pkg/tool/tool.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
@@ -404,6 +405,10 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
 		return err
 	}
 
+	if err := c.writeExecScript(ctx, commandProto, filepath.Join(outputPath, "run_locally.sh")); err != nil {
+		return err
+	}
+
 	log.Infof("Fetching input tree from input root digest.. %v", actionProto.GetInputRootDigest())
 	rootPath := filepath.Join(outputPath, "input")
 	os.RemoveAll(rootPath)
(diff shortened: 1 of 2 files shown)


go/pkg/tool/tool.go
```diff
@@ -393,6 +398,55 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
 	return c.writeProto(is, filepath.Join(outputPath, "input_node_properties.textproto"))
 }
 
+func (c *Client) writeExecScript(ctx context.Context, cmd *repb.Command, filename string) error {
+	if cmd == nil {
+		return fmt.Errorf("invalid comment (nil)")
```
mrahs, line 403:
```suggestion
		return fmt.Errorf("invalid command (nil)")
```

go/pkg/tool/tool.go
```diff
@@ -393,6 +398,55 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
...
+	}
+
+	var runActionScript bytes.Buffer
+	runActionFilename := filepath.Join(filepath.Dir(filename), "run_command.sh")
+	wd := cmd.WorkingDirectory
+	execCmd := strings.Join(cmd.GetArguments(), " ")
+	runActionScript.WriteString(fmt.Sprintf("#!/bin/bash\n\n"))
```
mrahs, line 410: Unnecessary `fmt`?

Maybe `fmt.Fprintf(runActionScript, "#!/bin/bash\n\n")`?

go/pkg/tool/tool.go
```diff
@@ -393,6 +398,55 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
...
+	runActionScript.WriteString(fmt.Sprintf("# This script is meant to be called by %v.\n", filename))
```
mrahs, line 411: "by" implies an actor. "from within" is probably clearer?

go/pkg/tool/tool.go
gkousik: Done.

go/pkg/tool/tool.go
gkousik: Done.

go/pkg/tool/tool.go
gkousik: Done.

mrahs (approved):
It looks like there are some hashes in the test that need to be updated now that there is an extra file in the output.

PR #521, opened 2023-10-27
Add missing formatting directives

Yannic:
```
compilepkg: nogo: errors found by nogo during build-time code analysis:
go/pkg/tool/tool.go:384:11: github.com/google/glog.Infof call has arguments but no formatting directives (printf)
```


mrahs (approved):
Thank you, Yannic. The CI here could use better configs for linters.

PR #523, opened 2023-11-07
Save show_action result while download action

ywmei-brt1:
Test:Manual Test

### DIFF
--- a/go/pkg/tool/tool.go
+++ b/go/pkg/tool/tool.go
@@ -428,13 +428,13 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
 			return err
 		}
 	}
-	res, err := c.formatActionResStr(ctx, actionProto, resPb, commandProto, cmdDg)
+	res, err := c.formatAction(ctx, actionProto, resPb, commandProto, cmdDg)
 	if err != nil {
-		return fmt.Errorf("error showing action %v: %v", outputPath, err)
+		return fmt.Errorf("error formatting action %v", err)
 	}
-	err = os.WriteFile(filepath.Join(outputPath, "show_action.txt"), []byte(res), 0644)
+	err = os.WriteFile(filepath.Join(outputPath, "action.txt"), []byte(res), 0644)
 	if err != nil {
-		return fmt.Errorf("error dumping to show_action.txt] %v: %v", outputPath, err)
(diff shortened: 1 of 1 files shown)


mrahs:
The data `ShowAction` fetches and prints is already downloaded. The user can `cat` the same information from files without having to call the server twice. Wouldn't that achieve what this PR is trying to do?

go/pkg/tool/tool.go
```diff
@@ -418,10 +422,21 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
...
 	}
-	if len(is.InputNodeProperties) == 0 {
-		return nil
+	if len(is.InputNodeProperties) != 0 {
+		err = c.writeProto(is, filepath.Join(outputPath, "input_node_properties.textproto"))
+		if err != nil {
+			return err
+		}
+	}
+	res, err := c.formatActionResStr(ctx, actionProto, resPb, commandProto, cmdDg)
```
mrahs, line 431: Nit: `formatAction` sounds descriptive enough.

go/pkg/tool/tool.go
```diff
@@ -418,10 +422,21 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
...
+	if err != nil {
+		return fmt.Errorf("error showing action %v: %v", outputPath, err)
```
mrahs, line 433: Is the outputPath relevant in this error?
```suggestion
		return fmt.Errorf("error formatting action: %v", err)
```

go/pkg/tool/tool.go
```diff
@@ -418,10 +422,21 @@ func (c *Client) DownloadAction(ctx context.Context, actionDigest, outputPath st
...
 	}
-	return c.writeProto(is, filepath.Join(outputPath, "input_node_properties.textproto"))
+	err = os.WriteFile(filepath.Join(outputPath, "show_action.txt"), []byte(res), 0644)
+	if err != nil {
+		return fmt.Errorf("error dumping to show_action.txt] %v: %v", outputPath, err)
```
mrahs, line 437: Nit: 
```suggestion
		return fmt.Errorf("error dumping to action.txt] %v: %v", outputPath, err)
```

mrahs (approved):
Thanks. This patch looks much better.

PR #524, opened 2023-11-07
fix preserving intermediate symlinks

mrahs:
This patch handles virtual inputs. The code that handles virtual inputs skips loadFiles entirely which effectively skips the previous patches.

This patch removes the redundant map by leveraging the existing one, fs. The tradeoff is less allocation from the removed map, but more calls to filepath.Rel and filepath.Join to get the key for the existing map.

Tested on a relevant action locally and confirmed it succeeds. Also ran a full build with no issues.

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -152,7 +152,10 @@ func getRemotePath(path, workingDir, remoteWorkingDir string) (string, error) {
 }
 
 // getExecRootRelPaths returns local and remote exec-root-relative paths for a given local absolute path
-func getExecRootRelPaths(absPath, execRoot, workingDir, remoteWorkingDir string) (relPath string, remoteRelPath string, err error) {
+// path may be relative or absolute. In both cases it's joined to and relativised to the execRoot.
+// This has unintuitive implications. For example, execRoot=/root and path=/foo, returns relPath=foo.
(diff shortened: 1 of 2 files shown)


go/pkg/client/tree.go
```diff
@@ -227,45 +230,65 @@ func evalParentSymlinks(execRoot, relPath string, materializeOutsideExecRoot boo
 	return evaledPathBuilder.String(), symlinks, nil
 }
 
+// loadIntermediateSymlinks inserts symlink nodes into fs.
+// If the symlink path already exists in fs, it is not overwritten, which means the first entry wins.
```
gkousik, line 234: nit: Can this be even clarified as "If the source path of the symlink already exists (e.g., "a" in a-->../afile)"

go/pkg/client/tree.go
```diff
@@ -227,45 +230,65 @@ func evalParentSymlinks(execRoot, relPath string, materializeOutsideExecRoot boo
...
+// This helps avoid redundant allocations from shared ancestors.
```
gkousik, line 235: Do you mind adding an example here? it helps with reasoning :)

gkousik (approved):

go/pkg/client/tree.go
mrahs: Done

go/pkg/client/tree.go
mrahs: Done

PR #526, opened 2023-11-08
Ignore files without read permissions

andusy:
We want to ignore files that don't have read permissions rather than returning an error, as it is likely that the file is not needed if it can't be read.

Bug: b/308481142

### DIFF
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -111,7 +111,7 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 
 // shouldIgnoreErr returns whether a given error should be ignored.
 func shouldIgnoreErr(err error) bool {
-	// We should skip files without read permissions. If the user doesn't have read permissions, 
+	// We should skip files without read permissions. If the user doesn't have read permissions,
 	// the file is unlikely to be used in the build in the first place.
 	if e, ok := err.(*filemetadata.FileError); ok {
 		return os.IsPermission(e.Err)
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -109,6 +109,16 @@ func shouldIgnore(inp string, t command.InputType, excl []*command.InputExclusio
 	return false
 }
 
+// shouldIgnoreErr returns whether a given error should be ignored.
+func shouldIgnoreErr(err error) bool {
+	// We should skip files without read permissions. If the user doesn't have read permissions, 
+	// the file is unlikely to be used in the build in the first place.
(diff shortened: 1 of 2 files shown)


go/pkg/client/tree.go
```diff
@@ -331,6 +331,10 @@ func loadFiles(execRoot, localWorkingDir, remoteWorkingDir string, excl []*comma
 			if shouldIgnore(absPath, command.DirectoryInputType, excl) {
 				continue
 			} else if meta.Err != nil {
+				// Skip files with read permission issues.
```
mrahs, line 334: Nit: expand this description to clarify why this is a reasonable thing to do.

go/pkg/client/tree.go
```diff
@@ -358,6 +362,10 @@ func loadFiles(execRoot, localWorkingDir, remoteWorkingDir string, excl []*comma
 			if shouldIgnore(absPath, command.FileInputType, excl) {
 				continue
 			} else if meta.Err != nil {
+				// Skip files with read permission issues.
+				if e, ok := meta.Err.(*filemetadata.FileError); ok && os.IsPermission(e.Err) {
+					continue
+				}
```
mrahs, line 368: It looks like this is worth extracting into a helper function, say `shouldIgnoreErr`.

go/pkg/client/tree_test.go
```diff
@@ -2117,6 +2117,88 @@ func TestComputeOutputsToUploadDirectories(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		root := t.TempDir()
+		if err := construct(root, tc.input); err != nil {
+			t.Fatalf("failed to construct input dir structure: %v", err)
+		}
+
+		if err := os.Chmod(filepath.Join(root, "bar"), 0100); err != nil {
```
mrahs, line 2159: This is specific to the one test case. If one is sufficient, it's not worth generalizing with test tables here.

go/pkg/client/tree_test.go
```diff
@@ -2117,6 +2117,73 @@ func TestComputeOutputsToUploadDirectories(t *testing.T) {
...
+	root := t.TempDir()
+	if err := construct(root, input); err != nil {
+		t.Fatalf("failed to construct input dir structure: %v", err)
+	}
+
+	if err := os.Chmod(filepath.Join(root, "bar"), 0100); err != nil {
+		t.Fatalf("failed to set permissions of bar: %v", err)
+	}
+
+	t.Run(desc, func(t *testing.T) {
```
mrahs, line 2149: Looks like an unnecessary sub-test?

mrahs (approved):

PR #529, opened 2024-01-24
Local disk CAS

ola-rozenfeld:
Implements an optional local disk cache (LRU, CAS) used for remote action outputs.
Goma has this feature, and many of our Reclient customers requested it since (usually because of slow internet connections).

In my benchmarks on Linux with a relatively good internet connection this was not able to reduce the runtime of a fully cached build by a statistically significant amount. But it didn't hurt either.


ola-rozenfeld:
Hang on, tests pass on Mac but not on Linux, checking...

ola-rozenfeld:
Update: Brave have tried this change and see no difference, because we still need to do an Action Cache RPC. I will now add optional Action Cache to this change to see how much improvement they get with it.

gkousik:
Thanks for the PR Ola! I generally like and agree with having a local disk cache feature, but I want to ask some due diligence questions.

1. In slow internet connection scenarios, has racing exec strategy proven to be helpful? Is local disk cache faster than racing in those scenarios? 
2. > In my benchmarks on Linux with a relatively good internet connection this was not able to reduce the runtime of a fully cached build by a statistically significant amount. But it didn't hurt either.

    This is very interesting - what was the size of your disk cache and what was the total output downloaded by the build? I am assuming this is a Chrome build? Its surprising that `network transfer + write` ~= `local_disk_copy` for a full build.
3. I think `StoreCas` also could suffer from a race condition where the local build modifies the file after writing it out. Perhaps we need to validate digest after `cp` is finished?
4. Finally a request, assuming we go ahead (I just want to know answers to (1) and (2)), could this be split up into smaller CLs? I have some readability comments that maybe easier to do with smaller changes :)

ola-rozenfeld:
Hey Kousik and Anas! Good questions! 1 - @goodov from Brave has tried this out on a slow connection -- maybe Aleksey can comment on how it compares to racing, because I know he has tried that as well (although it's not yet the default setting we recommend to customers). I'm wondering if there's a way for me to artificially create a VM with a slow connection to try it out -- maybe get a VM in the farthest possible region from our backend?...

2 - I tried a Chromium build on Linux with 50gb cache size, and the actual cache size was 26GB. Which, btw, implies that RBE download metrics that it displays at the end of the build are waaaay off (it usually says `RBE Stats: down 5.78 GB`).

3 - Good point! I didn't think the output files would be locally modified (do they? Do you know which actions do that?). I'll add handling of this case and a unit test. But I'm struggling to imagine this being a race condition unless it's literally two concurrent actions that output the same file -- and if that's the case, isn't it already a race condition in the build itself?...

4 - Yes, agreed, I did it in one go because I wanted @goodov to be able to give it a try easily. I think this can naturally break into 3 PRs -- first, introduce the diskcache with CAS-only plus unit tests, then second PR adds Action Cache functionality plus unit tests, then 3rd PR introduces the client flags plus instrumentation to actually use the diskcache in the client. Does that sound good?

Thank you!

ola-rozenfeld:
Updates:
- changed the directory structure following Marc-Antoine's (@maru3l) suggestion (based on Git and Isolate implementations) -- am now splitting the cache directory into 256 subdirectories based on 2-byte prefix of the digest, and using the other bytes for the name. Also moved the "_ac" to suffix. Both of these should improve performance, especially on Windows.
- Split this PR out into https://github.com/bazelbuild/remote-apis-sdks/pull/530 (will send out part 2 after part 1 is in).

I'm leaving this PR here for reference to the whole thing.

ola-rozenfeld:
@mrahs Friendly ping! :-) Let me know if you prefer to review the whole thing or in chunks (e.g. https://github.com/bazelbuild/remote-apis-sdks/pull/530).

Thank you!

mrahs:
> @mrahs Friendly ping! :-) Let me know if you prefer to review the whole thing or in chunks (e.g. #530).
> 
> Thank you!

Sorry I'm late. I'm trying to understand the context of such cache within past discussions and future requirements.

>I tried a Chromium build on Linux with 50gb cache size, and the actual cache size was 26GB. Which, btw, implies that RBE download metrics that it displays at the end of the build are waaaay off (it usually says RBE Stats: down 5.78 GB).

Ah, interesting. Any chance you've been able to spot the cause of this mismatch?

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,467 @@
...
+				if err != nil {
+					return fmt.Errorf("error parsing cached file name %s: %w", path, err)
+				}
+				info, err := d.Info()
+				if err != nil {
+					return fmt.Errorf("error getting file info of %s: %w", path, err)
+				}
+				it := &qitem{
+					key: k,
+					lat: fileInfoToAccessTime(info),
```
maruel, line 174: Why use the last access time instead of last modified time?
I don't believe that using the file system to track this information is a good idea. My implementations saved this as a dumb json file and that was very simple, I don't care about the actual time stamp, just the ordering; thus the file is a flat list of files in LRU for eviction order. This also removes the need for the priorityQueue type.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,467 @@
...
+		case <-d.ctx.Done():
+			return
+		case <-d.gcReq:
+			start := time.Now()
+			// Evict old entries until total size is below cap.
+			for uint64(atomic.LoadInt64(&d.stats.TotalSizeBytes)) > d.maxCapacityBytes {
+				d.mu.Lock()
+				it := heap.Pop(d.queue).(*qitem)
+				d.mu.Unlock()
+				size, err := d.getItemSize(it.key)
```
maruel, line 344: I don't understand why this does a stat on the file when the size is already in the digest.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,467 @@
...
+				}
+				atomic.AddInt64(&d.stats.TotalSizeBytes, -size)
+				atomic.AddInt64(&d.stats.TotalNumFiles, -1)
+				atomic.AddInt64(&d.stats.TotalGCedSizeBytes, size)
+				atomic.AddInt64(&d.stats.NumFilesGCed, 1)
+				it.mu.Lock()
+				diskOpsStart := time.Now()
+				// We only delete the files, and not the prefix directories, because the prefixes are not worth worrying about.
+				if err := os.Remove(d.getPath(it.key)); err != nil {
+					log.Errorf("Error removing file: %w", err)
```
maruel, line 357: It'd be good if the error would be accumulated, so that Shutdown() could be renamed to Close() and return the accumulated errors.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+	hash, isAc := strings.CutSuffix(pair[0], "_ac")
+	dg, err := digest.New(hash, size)
+	if err != nil {
+		return key{}, fmt.Errorf("invalid digest from file name %s: %w", fname, err)
+	}
+	return key{digest: dg, isCas: !isAc}, nil
+}
+
+func (d *DiskCache) getPath(k key) string {
```
maruel, line 252: Optional nit: I would have preferred files to have a uniform file name "%s-%d" % (hash, size), instead of having the size as an extension. Then you could use ".cas" as the marker for cas files. IMHO it would be a bit cleaner.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,467 @@
...
+	q.items[i], q.items[j] = q.items[j], q.items[i]
+	q.items[i].index = i
+	q.items[j].index = j
+}
+
+func (q *priorityQueue) Push(x any) {
+	if q.n == cap(q.items) {
+		// Resize the queue
+		old := q.items
+		q.items = make([]*qitem, 2*cap(old)) // Initial capacity needs to be > 0.
```
maruel, line 64:
```
if old == 0 {
  old = 256
}
```
then you can remove the comment

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,467 @@
...
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/protobuf/proto"
+
+	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	log "github.com/golang/glog"
+)
+
+type key struct {
+	digest digest.Digest
+	isCas  bool
```
maruel, line 28: Is isCas meant to do a performance optimization later? I don't see any behavior change based on this member.

ola-rozenfeld:
Thank you, Mark-Antoine -- I'm still tinkering with the whole thing, I just added stats and benchmark tests (not pushed yet), and I'm trying out different variations to see what's fastest. I hope I can get it all done by Monday.

go/pkg/client/cas_download.go
```diff
@@ -129,6 +140,11 @@ func (c *Client) DownloadOutputs(ctx context.Context, outs map[string]*TreeOutpu
 		if err := cache.Update(absPath, md); err != nil {
 			return fullStats, err
 		}
+		if c.diskCache != nil {
+			if err := c.diskCache.StoreCas(output.Digest, absPath); err != nil {
```
maruel, line 144: Do you think it could be done as part of DownloadFiles() ? Otherwise this adds a serial latency.

ola-rozenfeld:
I'm really sorry, but I have to admit to myself and others that I apparently have zero spare cycles to put into this in the foreseeable future :-(
Interested parties, such as @maruel or @goodov are welcome to pick up this work and go from here, merging it in any form/parts. Again, I'm sorry! In retrospect, I should have tried to merge and iterate rather than address all comments at once -- but you all had extremely helpful comments, which were important to address.

PR #530, opened 2024-01-31
Local disk cache: part 1

ola-rozenfeld:
Introducing the diskcache package, CAS-only (Action Cache to be added in part 2).

Split out from https://github.com/bazelbuild/remote-apis-sdks/pull/529 which has the whole thing, for reference.


go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,314 @@
...
+		shutdown: make(chan bool),
+	}
+	heap.Init(res.queue)
+	_ = os.MkdirAll(root, os.ModePerm)
+	// We use Git's directory/file naming structure as inspiration:
+	// https://git-scm.com/book/en/v2/Git-Internals-Git-Objects#:~:text=The%20subdirectory%20is%20named%20with%20the%20first%202%20characters%20of%20the%20SHA%2D1%2C%20and%20the%20filename%20is%20the%20remaining%2038%20characters.
+	for i := 0; i < 256; i++ {
+		_ = os.MkdirAll(filepath.Join(root, fmt.Sprintf("%02x", i)), os.ModePerm)
+	}
+	_ = filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
```
goodov, line 121: can this be run as a goroutine for each `00-ff` directory inside the for-loop above? Synchronous 256 directory walks doesn't look great, especially when each directory will contain thousands of files.

`DiskCache::New` should be multithreaded and use all available IO as possible. Posix OSes may be pretty fast, but Windows is not so cool when we're talking about directory walking.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,314 @@
...
+	defer in.Close()
+	out, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	if err := out.Chmod(srcInfo.Mode()); err != nil {
+		return err
+	}
+	defer out.Close()
+	_, err = io.Copy(out, in)
```
goodov, line 278: can we move the file copy function into a separate source (similar to `atim_*`) so it can be altered for `darwin` and `windows` to make use of [clonefile](https://github.com/golang/sys/blob/511ec846b66e39447acbd021083b26ca2747a188/unix/syscall_darwin.go#L564C7-L564C16) and upcoming Windows [DevDrive CoW](https://devblogs.microsoft.com/engineering-at-microsoft/copy-on-write-in-win32-api-early-access/)?

go/pkg/diskcache/diskcache.go
ola-rozenfeld: Yep, absolutely. Done.

go/pkg/diskcache/diskcache.go
ola-rozenfeld: Great idea, thank you! I didn't know of clonefile. I renamed the system-specific files to `sys_*` and changed the header to match, so that this change can be nicely done in a followup PR. It is a very contained optimization which doesn't affect the cache layout, so it's backward-compatible, and hence imo it belongs in a folllowup.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return err
+	}
+	// Required sanity check: sometimes the copy pretends to succeed, but doesn't, if
```
maruel, line 291: Can you expand on that? That's news to me. Or you mean in the general case of a race condition?

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+// If the digest exists in the disk cache, copy the file contents to the given path.
+func (d *DiskCache) LoadCas(dg digest.Digest, path string) bool {
+	k := key{digest: dg}
+	iUntyped, loaded := d.store.Load(k)
+	if !loaded {
+		return false
+	}
+	it := iUntyped.(*qitem)
+	it.mu.RLock()
+	if err := copyFile(d.getPath(k), path, dg.Size); err != nil {
```
maruel, line 312: This simplifies the control flow:
```
it.mu.RLock()
err := copyFile(d.getPath(k), path, dg.Size)
it.mu.RUnlock()
if err != nil {
```

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+	_ = os.MkdirAll(root, os.ModePerm)
+	// We use Git's directory/file naming structure as inspiration:
+	// https://git-scm.com/book/en/v2/Git-Internals-Git-Objects#:~:text=The%20subdirectory%20is%20named%20with%20the%20first%202%20characters%20of%20the%20SHA%2D1%2C%20and%20the%20filename%20is%20the%20remaining%2038%20characters.
+	var wg sync.WaitGroup
+	wg.Add(256)
+	for i := 0; i < 256; i++ {
+		prefixDir := filepath.Join(root, fmt.Sprintf("%02x", i))
+		go func() {
+			defer wg.Done()
+			_ = os.MkdirAll(prefixDir, os.ModePerm)
```
maruel, line 124: There are extreme cases where this could happen (e.g. disk full). I'd recommend to use https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext for error management instead of logging as a silent failure.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+	return key{digest: dg}, nil
+}
+
+func (d *DiskCache) getPath(k key) string {
+	return filepath.Join(d.root, k.digest.Hash[:2], fmt.Sprintf("%s.%d", k.digest.Hash[2:], k.digest.Size))
+}
+
+func (d *DiskCache) StoreCas(dg digest.Digest, path string) error {
+	if dg.Size > int64(d.maxCapacityBytes) {
+		return fmt.Errorf("blob size %d exceeds DiskCache capacity %d", dg.Size, d.maxCapacityBytes)
```
maruel, line 203: A more graceful behavior is to try to trigger a synchronous GC first before erroring out.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	it := &qitem{
+		key: key{digest: dg},
+		lat: time.Now(),
+	}
+	it.mu.Lock()
+	defer it.mu.Unlock()
+	_, exists := d.store.LoadOrStore(it.key, it)
```
maruel, line 211: d.store is already synchronized by it.mu. I'd vote to make store a normal map[digest]*qitem instead; it would be faster.

go/pkg/diskcache/sys_windows.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+// This will return correct values only if `fsutil behavior set disablelastaccess 0` is set.
+// Tracking of last access time is disabled by default on Windows.
+func GetLastAccessTime(path string) (time.Time, error) {
```
maruel, line 12: I'd recommend instead `func FileInfoToAccessTime(i *os.FileInfo) time.Time` and keep the os.Stat() call at the call site.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+		return err
+	}
+	if dstInfo.Size() != size {
+		return fmt.Errorf("copy of %s to %s failed: src/dst size mismatch: wanted %d, got %d", src, dst, size, dstInfo.Size())
+	}
+	return nil
+}
+
+// If the digest exists in the disk cache, copy the file contents to the given path.
+func (d *DiskCache) LoadCas(dg digest.Digest, path string) bool {
```
maruel, line 304: I'd propose to rename from LoadCas to CopyFromCache

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,323 @@
...
+				atomic.AddInt64(&res.sizeBytes, size)
+				res.mu.Lock()
+				heap.Push(res.queue, it)
+				res.mu.Unlock()
+				return nil
+			})
+		}()
+	}
+	wg.Wait()
+	go res.gc()
```
maruel, line 164: I think gc should only be called `if len(res.queue) > X && res.sizeBytes > Y && res.queue[0].lat < time.Now().Sub(Z)` to reduce unnecessary processing.

maruel (commented):
Thanks for the implementation!

go/pkg/diskcache/diskcache.go
maruel: I realize the sister function is called StoreCas. I don't know if it's a nomenclature reused in this code base, if so ignore this comment.

ola-rozenfeld: Yeah, Load* and Store* are consistent with sync.Map's naming, we also sometimes use Get/Update. I'll leave the naming decisions to @mrahs.

go/pkg/diskcache/diskcache.go
ola-rozenfeld: Yep, thank you.

go/pkg/diskcache/sys_windows.go
ola-rozenfeld: Sure, although note how the Darwin interface needs to differ from the others. Fortunately, the whole mess compiles...

go/pkg/diskcache/diskcache.go
ola-rozenfeld: In general, if we write to a file while concurrently deleting it, the Copy might not return an error (even if no bytes were actually copied), and neither the Stat. But the Stat will return different size than expected (usually, 0).
But the comment made me notice that `io.Copy` actually returns the number of bytes copied and seems to be correct, so I amended the sanity check to only verify the copy return value to be faster.

goodov:
I'm a little concerned about the whole "scan this huge dir on startup" process. In Chromium Reclient is auto started/stopped during each `autoninja` invocation. This is different compared to how Goma worked: there was a separate `goma_ctl ensure_start` command that kept the proxy in the background until a user terminated it manually, that's why local disk cache re-read wasn't an issue.

Does this mean `reproxy` startup (and `autoninja` call in general) will be delayed by the overall time to scan "local cache" directory? Or is this a separate process right now? @ola-rozenfeld can you please do some tests around this area?

I can think of these options to improve the situation:
1. Run the "local cache" scan in background and just "fail" cache lookups until it's ready (maybe with some separate status). This is not perfect, because the scan will be even slower when the actual build starts.
2. Use some kind of a simple "database" file that can be hot-loaded at startup and then be updated/synced in background (`proto` file for ex. or maybe some Go-specific serialization).

maruel (commented):
> I'm a little concerned about the whole "scan this huge dir on startup" process. In Chromium Reclient is auto started/stopped during each `autoninja` invocation. This is different compared to how Goma worked: there was a separate `goma_ctl ensure_start` command that kept the proxy in the background until a user terminated it manually, that's why local disk cache re-read wasn't an issue.
> 
> Does this mean `reproxy` startup (and `autoninja` call in general) will be delayed by the overall time to scan "local cache" directory? Or is this a separate process right now? @ola-rozenfeld can you please do some tests around this area?
> 
> I can think of these options to improve the situation:
> 
> 1. Run the "local cache" scan in background and just "fail" cache lookups until it's ready (maybe with some separate status). This is not perfect, because the scan will be even slower when the actual build starts.
> 2. Use some kind of a simple "database" file that can be hot-loaded at startup and then be updated/synced in background (`proto` file for ex. or maybe some Go-specific serialization).

I agree and was about to suggest the same thing. In isolated, I used the second option. A very simple json file worked fine, really, just keeping object in a LRU list. Then a priority queue is not needed, just mutate the slice in place. Go is surprisingly fast at this, I would be surprised if it were not faster than using a priority queue. One lock instead of many.

The isolated code soft-failed when a file was missing / corrupted on fetch.
Only "once in a while" would you scan the whole directory to make sure there's no forgotten files.

go/pkg/diskcache/diskcache.go
```diff
@@ -0,0 +1,331 @@
...
+	if newSize > d.maxCapacityBytes {
+		select {
+		case d.gcReq <- atomic.AddUint64(&d.gcTick, 1):
+		default:
+		}
+	}
+	return nil
+}
+
+func (d *DiskCache) gc() {
```
mrahs, line 236: I think this implementation may have the following issues:
- if the capacity is not sufficient for the workload, eviction becomes part of every update.
- having three different locks being acquired and released independently is prune to races; e.g. attempting to load an item after it's been deleted from the ordered list but before it's been deleted from the map, will fail to update the index in the list.
- the client must remember to `Shutdown` the cache to avoid resource leaks.

The capacity issue can be mitigated by evicting a percentage of the cache rather than just enough to meet the threshold. This accommodates several new items before triggering another eviction.

The synchronization issue can be traded off with a simpler implementation that uses a single lock. Updating the list and the map should be an atomic operation in which the entire list is shared and therefore is the bottleneck. Actually, hits are bounded by IO anyways. Misses on the other hand can be faster with less locking. Then again, misses are always followed by updates which are bounded by IO. I think IO is more of a bottleneck than synchronization so it's worth going for a simple locking flow.

How bad is it if eviction events were blocking? I think if eviction saturates IO bandwidth it should be the same whether it's async or sync. Maybe blocking would be faster if we consider disk locality. I also think that most users would prefer using a large cache to avoid eviction anyways. Without a background eviction task, the cache would be simpler to work with.

go/pkg/diskcache/diskcache.go
mrahs: I think the cache can have three main methods: putting bytes, getting bytes, copying bytes to a file path. Since the cache deals with bytes, I'm inclined to like `Write(key, Reader), Read(key, Writer), Copy(key, filepath)`, but `Store, Load, LoadInto/CopyTo` also sound reasonable. I made another comment on generalizing this method which would avoid having to choose suffixes.

go/pkg/diskcache/diskcache.go
mrahs: I'd push error handling on the user by making the `client.Opt` accept a `*DiskCache` instance.

ola-rozenfeld:
> *bump*

Yes, I'm really sorry for disappearing -- I've had very hectic few weeks and didn't have time for this. This effort is something I'm doing voluntarily in my spare time, and it is not currently prioritized by EngFlow. I will be back to working on this in the next week or two!

PR #532, opened 2024-02-06
remotetool: adding option to upload directory

ola-rozenfeld:
Feature request from Marc-Antoine's (@maruel).

### DIFF
--- a/go/pkg/tool/tool.go
+++ b/go/pkg/tool/tool.go
@@ -326,6 +326,20 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
 	return err
 }
 
+// UploadDirectory uploads a directory the specified path as a Merkle-tree to the remote cache.
+func (c *Client) UploadDirectory(ctx context.Context, path string) error {
+	log.Infof("Computing Merkle tree rooted at %s.", path)
+	root, blobs, stats, err := c.GrpcClient.ComputeMerkleTree(ctx, path, "", "", &command.InputSpec{Inputs: []string{"."}}, filemetadata.NewNoopCache())
+	if err != nil {
+		return err
+	}
+	log.Infof("Directory root digest: %v", root)
+	log.Infof("Directory stats: %d files, %d directories, %d symlinks, %d total bytes", stats.InputFiles, stats.InputDirectories, stats.InputSymlinks, stats.TotalInputBytes)
+	log.Infof("Uploading directory %v rooted at %s to CAS.", root, path)
+	_, _, err = c.GrpcClient.UploadIfMissing(ctx, blobs...)
+	return err
+}
+
 func (c *Client) writeProto(m proto.Message, baseName string) error {
 	f, err := os.Create(baseName)
 	if err != nil {
--- a/go/pkg/tool/tool.go
+++ b/go/pkg/tool/tool.go
@@ -326,9 +326,9 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
 	return err
 }
 
-// UploadDirectory uploads a directory the specified path as a Merkle-tree to the remote cache.
+// UploadDirectory uploads a directory from the specified path as a Merkle-tree to the remote cache.
 func (c *Client) UploadDirectory(ctx context.Context, path string) error {
-	log.Infof("Computing Merkle tree rooted at %s.", path)
+	log.Infof("Computing Merkle tree rooted at %s", path)
 	root, blobs, stats, err := c.GrpcClient.ComputeMerkleTree(ctx, path, "", "", &command.InputSpec{Inputs: []string{"."}}, filemetadata.NewNoopCache())
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


maruel:
Thanks a lot, this is really appreciated!

maruel:
I confirmed the code works. Thanks! The only issue I'm seeing is to get the hash of the uploaded directory. I had to look at the logs to get the line `Directory root digest: <hash>`. If it could be communicated back in a more structured way, it would be awesome.

I don't think this is a blocker for this PR, this as is already works, it's just that running a regexp on the output isn't optimal.

ola-rozenfeld:
> I confirmed the code works. Thanks! The only issue I'm seeing is to get the hash of the uploaded directory. I had to look at the logs to get the line `Directory root digest: <hash>`. If it could be communicated back in a more structured way, it would be awesome.
> 
> I don't think this is a blocker for this PR, this as is already works, it's just that running a regexp on the output isn't optimal.

Yeah, agreed, I also was not a fan of that. What would you suggest? Adding another path flag to output the digest to?

maruel:
Given that all all outputs are currently generated by glog, and glog only prints to stderr, I think communicating back to stdout would be totally acceptable. The same for upload_blob.

An extensible form would be print json to stdout, e.g. change UploadBlob() and UploadDirectory() to return the digest, then have main.go do https://go.dev/play/p/6fFjHyDbgbx

ola-rozenfeld:
Hmm, I suspect my future reviewer might prefer that I either just output the digest to stdout as a string, or dump it into an optional file argument. I'll let the reviewer (most likely @mrahs ) decide -- any option will be very simple here. Thank you!

maruel:
Sounds good. The only value proposition for using json for output is that it can be extended in the future with additional data, e.g. cache hit rate and other metrics.

maruel (approved):

go/pkg/tool/tool.go
```diff
@@ -326,6 +326,20 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
 	return err
 }
 
+// UploadDirectory uploads a directory the specified path as a Merkle-tree to the remote cache.
```
mrahs, line 329:
```suggestion
// UploadDirectory uploads a directory from the specified path as a Merkle-tree to the remote cache.
```

go/pkg/tool/tool.go
```diff
@@ -326,6 +326,20 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
...
+func (c *Client) UploadDirectory(ctx context.Context, path string) error {
+	log.Infof("Computing Merkle tree rooted at %s.", path)
```
mrahs, line 331: The period might look like a dot with the extension omitted.
```suggestion
	log.Infof("Computing Merkle tree rooted at %s", path)
```

mrahs (approved):

go/pkg/tool/tool.go

go/pkg/tool/tool.go

ola-rozenfeld:
> Re printing data to stdout: I see that value of being able to use the tool in a pipeline; I've had instances where I wanted to do so. I think it's worth having a structured way of printing and controlling the output of the tool. I'm thinking a `-json` flag that prints to stdout would be reasonable. The list of available keys would be printed in the flag's help string. Perhaps the flag can take a string to specify which keys to include, but that's probably not useful considering the size of the keyset, and it can be extended to do so in the future if needs be without breaking existing workflows.
> 
> If you agree, this fits nicely into a separate PR.

Great, I'll add a JSON flag, then, in a separate PR after this one is merged. Thank you!

ola-rozenfeld:
Json flag: https://github.com/bazelbuild/remote-apis-sdks/pull/533

go/pkg/tool/tool.go
```diff
@@ -326,6 +326,20 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
 	return err
 }
 
+// UploadDirectory uploads a directory from the specified path as a Merkle-tree to the remote cache.
+func (c *Client) UploadDirectory(ctx context.Context, path string) error {
+	log.Infof("Computing Merkle tree rooted at %s", path)
+	root, blobs, stats, err := c.GrpcClient.ComputeMerkleTree(ctx, path, "", "", &command.InputSpec{Inputs: []string{"."}}, filemetadata.NewNoopCache())
```
maruel, line 332: I just realized, the current implementation serializes all hashing, then all uploading. Both should be happening concurrently for effective performance.

mrahs: That's not supported in the `client` package. It's supported in the `cas` package, but that one does not have feature parity with the implementation in the `client` package which is why it's not integrated yet. We are working on converging the two such that streaming (concurrently digest and upload) and batching (digest all, upload all) workflows are possible. No ETAs for the time being though.

maruel: We're observing a good 90s to hash our content so if this can be prioritized it would be very appreciated. Thanks

PR #533, opened 2024-02-12
remotetool: add --json output

ola-rozenfeld:
Followup to https://github.com/bazelbuild/remote-apis-sdks/pull/532

This does the simplest thing possible for now -- we can always add more keys by creating a custom struct that we return from all tool operations, and marshalling that as json.

### DIFF
--- a/go/cmd/remotetool/main.go
+++ b/go/cmd/remotetool/main.go
@@ -69,7 +69,7 @@ var (
 	overwrite    = flag.Bool("overwrite", false, "Overwrite the output path if it already exist.")
 	actionRoot   = flag.String("action_root", "", "For execute_action: the root of the action spec, containing ac.textproto (Action proto), cmd.textproto (Command proto), and input/ (root of the input tree).")
 	execAttempts = flag.Int("exec_attempts", 10, "For check_determinism: the number of times to remotely execute the action and check for mismatches.")
-	jsonOutput   = flag.String("json_output", "", "Path to output operation result as JSON. Currently supported for \"upload_dir\", and includes various upload metadata (see UploadStats).")
+	jsonOutput   = flag.String("json", "", "Path to output operation result as JSON. Currently supported for \"upload_dir\", and includes various upload metadata (see UploadStats).")
 	_            = flag.String("input_root", "", "Deprecated. Use action root instead.")
 )
 
@@ -150,9 +150,13 @@ func main() {
 		us, err := c.UploadDirectory(ctx, getPathFlag())
 		if *jsonOutput != "" {
 			js, _ := json.MarshalIndent(us, "", "  ")
-			log.Infof("Outputting JSON results to %s", *jsonOutput)
-			if err := os.WriteFile(*jsonOutput, []byte(js), 0o666); err != nil {
-				log.Exitf("Error writing JSON output to file: %v", err)
+			if *jsonOutput == "-" {
+				fmt.Printf("%s\n", js)
+			} else {
+				log.Infof("Outputting JSON results to %s", *jsonOutput)
+				if err := os.WriteFile(*jsonOutput, []byte(js), 0o666); err != nil {
+					log.Exitf("Error writing JSON output to file: %v", err)
+				}
 			}
 		}
 		if err != nil {
--- a/go/cmd/remotetool/main.go
+++ b/go/cmd/remotetool/main.go
@@ -69,7 +69,7 @@ var (
 	overwrite    = flag.Bool("overwrite", false, "Overwrite the output path if it already exist.")
 	actionRoot   = flag.String("action_root", "", "For execute_action: the root of the action spec, containing ac.textproto (Action proto), cmd.textproto (Command proto), and input/ (root of the input tree).")
 	execAttempts = flag.Int("exec_attempts", 10, "For check_determinism: the number of times to remotely execute the action and check for mismatches.")
-	jsonOutput   = flag.Bool("json", false, "If set, output operation result as JSON. Currently supported for \"upload_dir\": includes the \"digest\" key for the root digest of the uploaded directory.")
+	jsonOutput   = flag.Bool("json", false, "If set, output operation result as JSON. Currently supported for \"upload_dir\", and includes various upload metadata (see UploadStats).")
 	_            = flag.String("input_root", "", "Deprecated. Use action root instead.")
 )
 
@@ -147,13 +147,13 @@ func main() {
 		}
 
 	case uploadDir:
-		dg, err := c.UploadDirectory(ctx, getPathFlag())
+		us, err := c.UploadDirectory(ctx, getPathFlag())
 		if err != nil {
 			log.Exitf("error uploading directory for path %s: %v", getPathFlag(), err)
 		}
 		if *jsonOutput {
-			d, _ := json.Marshal(map[string]string{"digest": dg.String()})
-			fmt.Printf("%s\n", d)
+			js, _ := json.MarshalIndent(us, "", "  ")
+			fmt.Printf("%s\n", js)
 		}
 
 	default:
--- a/go/cmd/remotetool/main.go
+++ b/go/cmd/remotetool/main.go
@@ -69,7 +69,7 @@ var (
 	overwrite    = flag.Bool("overwrite", false, "Overwrite the output path if it already exist.")
 	actionRoot   = flag.String("action_root", "", "For execute_action: the root of the action spec, containing ac.textproto (Action proto), cmd.textproto (Command proto), and input/ (root of the input tree).")
 	execAttempts = flag.Int("exec_attempts", 10, "For check_determinism: the number of times to remotely execute the action and check for mismatches.")
-	jsonOutput   = flag.Bool("json", false, "If set, output operation result as JSON. Currently supported for \"upload_dir\", and includes various upload metadata (see UploadStats).")
+	jsonOutput   = flag.String("json_output", "", "Path to output operation result as JSON. Currently supported for \"upload_dir\", and includes various upload metadata (see UploadStats).")
 	_            = flag.String("input_root", "", "Deprecated. Use action root instead.")
 )
 
@@ -148,9 +148,12 @@ func main() {
 
 	case uploadDir:
 		us, err := c.UploadDirectory(ctx, getPathFlag())
-		if *jsonOutput {
+		if *jsonOutput != "" {
 			js, _ := json.MarshalIndent(us, "", "  ")
-			fmt.Printf("%s\n", js)
+			log.Infof("Outputting JSON results to %s", *jsonOutput)
+			if err := os.WriteFile(*jsonOutput, []byte(js), 0o666); err != nil {
+				log.Exitf("Error writing JSON output to file: %v", err)
+			}
 		}
 		if err != nil {
 			log.Exitf("error uploading directory for path %s: %v", getPathFlag(), err)
--- a/go/cmd/remotetool/main.go
+++ b/go/cmd/remotetool/main.go
@@ -148,12 +148,12 @@ func main() {
 
 	case uploadDir:
 		us, err := c.UploadDirectory(ctx, getPathFlag())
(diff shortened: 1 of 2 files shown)


go/cmd/remotetool/main.go
```diff
@@ -145,9 +147,14 @@ func main() {
 		}
 
 	case uploadDir:
-		if err := c.UploadDirectory(ctx, getPathFlag()); err != nil {
+		dg, err := c.UploadDirectory(ctx, getPathFlag())
+		if err != nil {
 			log.Exitf("error uploading directory for path %s: %v", getPathFlag(), err)
 		}
+		if *jsonOutput {
+			d, _ := json.Marshal(map[string]string{"digest": dg.String()})
```
maruel, line 155: How hard would it be to add the cache hit rate and more metadata? In isolated I had included the individual file sizes as a delta-coded integer list. It was really sweet to create an histogram while packing the data as efficiently as possible; https://chromium.googlesource.com/infra/luci/luci-go/+/c364d2ae0a1c94014db11ff581ae7c9fa2ba1e3e/common/isolated/large.go
That said, sums and percentiles are probably sufficient.

mrahs (approved):

go/pkg/tool/tool.go
```diff
@@ -326,18 +326,43 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
...
+	var sumMissingBytes int64
+	for _, d := range missing {
+		sumMissingBytes += d.Size
+	}
+	us := &UploadStats{
+		TreeStats:            *stats,
+		RootDigest:           root,
+		TotalNumBlobs:        int64(len(blobs)),
+		NumCacheMisses:       int64(len(missing)),
+		RealBytesUploaded:    n,
```
maruel, line 362: Would you mind using terminology like CacheHits / CacheMisses for sizes too? Real and Logical can be more obscure to folks unfamiliar with the system.

maruel (approved):
Thanks!

go/pkg/tool/tool.go
ola-rozenfeld: Oh, they're not the same thing -- the difference between real and logical is compression only -- TotalInputBytes is the total size of the entire directory + metadata protos. Maybe it would be better if I omit the real/logical distinction altogether and just report the TotalBytesUploaded instead?

maruel: Ah! Maybe use "CompressedBytesUploaded" then instead of RealBytesUploaded?  I think it's really useful data to have! Then I'd advocate for "CacheMissesBytes" for LogicalBytesUploaded. I don't feel "BytesCacheMisses" sounds good but it'd be nice to be coherent with "NumCacheMisses".

ola-rozenfeld: Sure, done. But I worry that CompressedBytesUploaded might sound misleading, because it will only be different from CacheMissesBytes if compression is enabled.

mrahs: I might suggest "BytesTransferred" for actual bytes moved over the wire, "BytesRequested" for total bytes provided in the request, and "BytesProcessed" for actual bytes processed locally (may be less than BytesRequested if processing was interrupted).

maruel: That's a great idea.

ola-rozenfeld: I didn't get it, sorry -- CompressedBytesUploaded->BytesTransferred, and what should I rename CacheMissesBytes to? Also, if any error happened I prefer not to return any stats at all, because then it's not clear from the tool output that an error occurred (lmk if that's wrong and we want to print partial stats).

mrahs: Good question. The output should be explicit about errors. I agree that in this case, an non-zero exit code with no output is sufficient since the errors are being logged to stderr. The "BytesProcessed" is useful internally when it's combined from multiple requests, which is not in scope for this tool.

mrahs: Hmm, I was thinking about it from the perspective of an interactive use-case, where the user can spot-check stderr. If we're going to support scripting workflows, we'd need to make sure that both the exit code and an error field are set since users might use either or both. Also, the error message now needs to be printed on both streams when `--json` is provided, which might call for some utility functions.

ola-rozenfeld: Okay, renamed the field, leaving out BytesProcessed for now, added an Error field, and printed the JSON unconditionally. PTAL.

go/pkg/tool/tool.go
```diff
@@ -329,39 +329,41 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
...
 
 // UploadDirectory uploads a directory from the specified path as a Merkle-tree to the remote cache.
 func (c *Client) UploadDirectory(ctx context.Context, path string) (*UploadStats, error) {
 	log.Infof("Computing Merkle tree rooted at %s", path)
 	root, blobs, stats, err := c.GrpcClient.ComputeMerkleTree(ctx, path, "", "", &command.InputSpec{Inputs: []string{"."}}, filemetadata.NewNoopCache())
 	if err != nil {
-		return nil, err
+		return &UploadStats{Error: err.Error()}, err
+	}
+	us := &UploadStats{
```
maruel, line 347: I'd say to move this before the condition at line 344 so the progress can be registered. It's useful to know if any progress was made or not.

go/pkg/tool/tool.go
```diff
@@ -329,39 +329,41 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
...
 	}
-	us := &UploadStats{
-		TreeStats:               *stats,
-		RootDigest:              root,
-		TotalNumBlobs:           int64(len(blobs)),
-		NumCacheMisses:          int64(len(missing)),
-		CompressedBytesUploaded: n,
-		CacheMissesBytes:        sumMissingBytes,
-	}
+	us.NumCacheMisses = int64(len(missing))
```
maruel, line 364: Similarly, I'd move these before the error checking. It's more useful for the caller.

maruel (approved):
looks great with two minor code movement.

go/pkg/tool/tool.go
maruel: lgtm

go/pkg/tool/tool.go
ola-rozenfeld: That's not how ComputeMerkeTree [works](https://github.com/bazelbuild/remote-apis-sdks/blob/46b7334fa1151e051515d606ad849e2856ea7210/go/pkg/client/tree.go#L412): it follows the pattern that we're breaking here, which is the return value of everything else is empty when a non-nil error is returned. So we don't have any partial progress. Ditto below.
We can change that contract, but I think it's better done separately.

go/pkg/tool/tool.go
ola-rozenfeld: Ditto -- both functions do not currently return partial progress.

go/pkg/tool/tool.go
maruel: Sounds good for separate PR. Thanks!

go/cmd/remotetool/main.go
```diff
@@ -145,7 +147,12 @@ func main() {
 		}
 
 	case uploadDir:
-		if err := c.UploadDirectory(ctx, getPathFlag()); err != nil {
+		us, err := c.UploadDirectory(ctx, getPathFlag())
+		if *jsonOutput {
+			js, _ := json.MarshalIndent(us, "", "  ")
+			fmt.Printf("%s\n", js)
```
maruel, line 153: Did you mean to write to `os.WriteFile(jsonOutput, js, 0o666)` ?

go/pkg/tool/tool.go
```diff
@@ -326,18 +326,45 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
 	return err
 }
 
+// UploadStats contains various metadata of a directory upload.
+type UploadStats struct {
```
maruel, line 330: optional nit: technically since it's a main package, it doesn't need to be exported.

maruel (approved):
still lgtm (I can't resolve comments)

go/pkg/tool/tool.go
```diff
@@ -326,18 +326,45 @@ func (c *Client) DownloadDirectory(ctx context.Context, rootDigest, path string)
...
+	rc.TreeStats
+	RootDigest       digest.Digest
+	TotalNumBlobs    int64
+	NumCacheMisses   int64
+	BytesTransferred int64
+	CacheMissesBytes int64
```
mrahs, line 336: Nit: I find shared prefixes easier to read and helps auto-complete engines with suggestions.

```suggestion
	CountBlobs    int64
	CountCacheMisses  int64
	BytesTransferred int64
	BytesCacheMisses int64
```

mrahs (approved):

go/cmd/remotetool/main.go
ola-rozenfeld: I didn't, but sure, if that's what you prefer :-) Done.

go/pkg/tool/tool.go
ola-rozenfeld: It's not, though. This is the `tool` package.

go/pkg/tool/tool.go

go/cmd/remotetool/main.go
mrahs: Sorry to reopen this, but wouldn't printing to stdout accommodate pipes and files (via redirection) while writing to a file complicates the pipe workflow?

maruel: The precedent in the posix way of doing this is to specific "-" as output to stdout. I think we'd get the best of both worlds.

go/pkg/tool/tool.go
mrahs: I think it'd still work like a factory pattern, allowing the main package to access the struct when returned from here, but disallowing it from creating new instances of this struct.

maruel: Why? It's dead data. What are you trying to achieve?

go/cmd/remotetool/main.go
mrahs: Isn't that usually for tools that naturally expect files for in/out?

go/pkg/tool/tool.go
mrahs: I guess my point was that: not exporting this struct would still work, even if it's used in an external package like `main`.

go/cmd/remotetool/main.go
maruel: Users want a tool to be composable. Since remotetool sends its status update to stderr and/or a file, I agree it is possible to use stdout to get the data. While in many case it's great, sometimes the user may want to not take a chance and use a named file instead. For example curl will output to stdout by default but may redirect to the user's choosing with `--output` and associated flags. Exceptions to that are use case where incrementally of the processing is needed so they do not offer the option to write to a file. A great example is `go test -json` where the user will want to process a list of json objects.

go/pkg/tool/tool.go
maruel: I misread the PR, and I'd recommend to not change anything here. A simple exported struct was the right choice IMO.

go/cmd/remotetool/main.go
maruel: To be clear, especially that I commented a lot here, I don't have a strong opinion and I'm fine with whatever @ola-rozenfeld or @mrahs prefers. If you prefer a bool `-json` flag it's okay too.

mrahs: Thanks for taking the time to bounce ideas! No strong opinions here either. `json=-` (which can be supported later) or `-json` works for me.

Ola, let me know if you're ready to merge and I'll hit the button.

ola-rozenfeld: Okay, added support for "json=-" for stdout. That's it, should be ready to merge.
Thank you Anas and Marc Antoine for the quick reviews!

PR #534, opened 2024-02-16
add a placeholder function for templating shell scripts

mrahs:


gkousik:
Could you add a description of what shellSprintf would eventually look like? LGTM Otherwise.

gkousik (approved):

PR #536, opened 2024-02-22
Pass back auxiliary_metadata as part of Command

gkousik:
This is supported by the API spec, however the API spec suggests an anyproto for the actual metadata. So we cannot use a specific message for the proto. For now, I've defined a fake proto for the purpose of testing.

Tested: Unit tests


gkousik:
@Mikes-rec for review

mrahs (approved):

PR #537, opened 2024-02-29
add dir content cache for ComputeMerkleTree

mrahs:
FMC (file metadata cache) helps avoid IO cost for digesting files. When computing a merkle tree the directories still need to be listed to construct the tree. Caching the listing helps avoid the cost of IO for that.

However, this might break some builds that have directories with varying content throughout the build. For now, the cache is gated by a flag to help guide a future change on how to filter out such directories from the cache.


gkousik:
> When computing a merkle tree the directories still need to be listed to construct the tree. Caching the listing helps avoid the cost of IO for that.

I think this will be problematic in output dirs. Directory contents do change often in output dirs as and when we write output files - did you get a chance to test this turned on in a Chrome build?

mrahs:
> > When computing a merkle tree the directories still need to be listed to construct the tree. Caching the listing helps avoid the cost of IO for that.
> 
> I think this will be problematic in output dirs. Directory contents do change often in output dirs as and when we write output files - did you get a chance to test this turned on in a Chrome build?

Yep. The build succeeded without fallbacks. I'm not sure what other builds might fail, which is I implemented it as an opt-in option for a controlled rollout.

go/pkg/client/tree.go
```diff
@@ -366,15 +366,18 @@ func loadFiles(execRoot, localWorkingDir, remoteWorkingDir string, excl []*comma
 				return meta.Err
 			}
 
-			f, err := os.Open(absPath)
-			if err != nil {
-				return err
-			}
+			files := meta.DirChildren
+			if files == nil {
+				f, err := os.Open(absPath)
```
gkousik, line 371: This code seems repeated, can this be a utility function that tree.go and filemetadata.go uses?

go/pkg/filemetadata/cache.go
```diff
@@ -17,14 +17,31 @@ func ResetGlobalCache() {
 
 // Cache is a store for file digests that supports invalidation.
```
gkousik, line 18: This doc line needs to be updated.

go/pkg/filemetadata/cache.go
```diff
@@ -17,14 +17,31 @@ func ResetGlobalCache() {
...
+	withDirContent bool
+}
+
+type Option func(*fmCache)
+
+// WithDirContent adds the list of files under a directory to the metadata.
+// exclude is currently ignored, but is meant to avoid caching content
+// for directories that are known to change their content during the lifetime
+// of the cache.
+func WithDirContent(exclude string) Option {
```
gkousik, line 32: can we introduce this option when we actually use it?

go/pkg/filemetadata/cache_test.go
```diff
@@ -112,3 +114,56 @@ func TestLoadAfterChangeWithoutValidation(t *testing.T) {
 		t.Errorf("Cache has wrong num of CacheMisses, want 1, got %v", c.GetCacheMisses())
 	}
 }
+
+func TestWithDirContent(t *testing.T) {
```
gkousik, line 118: Do you want to write a test where you call `c.Get()`, add a file, call `c.Get()` again and ensure that the second file does NOT show up? This would confirm the cache actually works.

go/pkg/client/tree.go
mrahs: It's going to be:
```
f, err := os.Open(absPath)
if err != nil {
	return err
}

files, err = f.Readdirnames(-1)
f.Close()
if err != nil {
	return err
}
```
vs
```
import "github.com/bazelbuild/remote-apis-sdks/go/pkg/io"

names, err := io.Readdirnames(absPath)
if err != nil {
	return err
}
files = names
```
but with a new exported package just for the utility function. I'm hesitant about exporting helper functions.

go/pkg/filemetadata/cache.go
mrahs: It's needed to allow reclient to selectively toggle it via a flag. I'll remove the string argument for now and add a note about this being experimental.

go/pkg/client/tree.go
gkousik: I do think the latter impl reduces cognitive load, but upto you though.

go/pkg/filemetadata/cache.go
gkousik: yeah I meant the exclude string arg!

go/pkg/client/tree.go
mrahs: I agree it's cleaner, but I'm not sure about the public helper function in a library. Tbh, many packages should have been internal.

go/pkg/filemetadata/cache_test.go
```diff
@@ -112,3 +114,57 @@ func TestLoadAfterChangeWithoutValidation(t *testing.T) {
...
+	want = &Metadata{
+		Digest:       digest.Empty,
+		IsDirectory:  true,
+		IsExecutable: true,
+		DirChildren:  []string{filepath.Base(filename)},
+	}
+	if diff := cmp.Diff(want, got, ignoreMtime); diff != "" {
+		t.Fatalf("Get(%v) returned diff. (-want +got)\n%s", dir, diff)
+	}
+	_ = createTempFile(t, dir, []byte("content3"))
```
gkousik, line 165: nit: you can ignore the "_ ="

gkousik (approved):

Issue #525, opened 2023-11-08
Allow clients to provide InputDigests to avoid unnecessary transfer


bshashank:
Today the API for the SDK only allows clients to specify `Inputs` (files available to the client) and `VirtualInputs`(binary available to clients) (from [command.go](https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/command/command.go#L102-L105))
But I would like to extend this to support providing digests that are available in the remote server.

I have a [fork with this code](https://github.com/bazelbuild/remote-apis-sdks/commit/7f1ed2182bbba9cb5bcdc8e7e9a26458a38c28d0) that I am able to use for this. But I wanted to ask folks here if it would make sense to add to the SDK?

This would be useful when trying to do "build-without-bytes" for example.

mrahs:
That's sounds reasonable. It looks like `VirtualInputs` can be extended to support that. Currently, the actual bytes are expected, but another field can be added to set the digest instead. Now this of course introduces an inconsistency problem if both the bytes and the digests are set, but that can be discussed if needs be.

My concern with adding a separate field for digests is that it needs to replicate some meta fields to describe the digest, such as the file type and executable bit.

It would be great if you can check if extending `VirtualInputs` in a backwards compatible way can fulfill your additional requirements. I'm happy to review PRs for this.

PR #542, opened 2024-03-14
Extend Virtual Inputs to allow digests

andusy:
Allows virtual inputs to be used with digests. If Contents is provided with InputDigest or the digest does not exist in the CAS, return an error.

Bug: b/324349357, https://github.com/bazelbuild/remote-apis-sdks/issues/525

### DIFF
--- a/go/pkg/client/cas_upload.go
+++ b/go/pkg/client/cas_upload.go
@@ -287,7 +287,7 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 			contextmd.Infof(ctx, log.Level(2), "Skipping upload of empty entry %s", ue.Digest)
 			continue
 		}
-		if ue.IsVirtualInputWithDigest() {
+		if ue.IsVirtualFile() {
 			return nil, 0, fmt.Errorf("virtual input with digest: %s provided, but does not exist in CAS", ue.Digest)
 		}
 		req := &uploadRequest{
--- a/go/pkg/client/cas_upload.go
+++ b/go/pkg/client/cas_upload.go
@@ -287,6 +287,9 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 			contextmd.Infof(ctx, log.Level(2), "Skipping upload of empty entry %s", ue.Digest)
 			continue
 		}
+		if ue.IsVirtualInput && ue.IsFile() {
+			return nil, 0, fmt.Errorf("virtual input with digest: %s provided, but does not exist in CAS", ue.Digest)
+		}
 		req := &uploadRequest{
 			ue:   ue,
 			meta: meta,
--- a/go/pkg/client/cas_upload.go
+++ b/go/pkg/client/cas_upload.go
@@ -288,7 +288,7 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 			continue
 		}
 		if ue.IsVirtualFile() {
-			return nil, 0, fmt.Errorf("virtual input with digest: %s provided, but does not exist in CAS", ue.Digest)
+			return nil, 0, fmt.Errorf("virtual input with digest %q provided, but does not exist in CAS", ue.Digest)
 		}
 		req := &uploadRequest{
 			ue:   ue,
--- a/go/pkg/client/cas_upload.go
+++ b/go/pkg/client/cas_upload.go
@@ -287,7 +287,7 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 			contextmd.Infof(ctx, log.Level(2), "Skipping upload of empty entry %s", ue.Digest)
 			continue
 		}
-		if ue.IsVirtualInput && ue.IsFile() {
+		if ue.IsVirtualInputWithDigest() {
 			return nil, 0, fmt.Errorf("virtual input with digest: %s provided, but does not exist in CAS", ue.Digest)
 		}
 		req := &uploadRequest{
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -445,7 +445,7 @@ func (c *Client) ComputeMerkleTree(ctx context.Context, execRoot, workingDir, re
 		}
 		entry := uploadinfo.EntryFromBlob(i.Contents)
 		if i.InputDigest != "" {
-			dg, err := digest.NewFromString(i.InputDigest);
+			dg, err := digest.NewFromString(i.InputDigest)
 			if err != nil {
 				return digest.Empty, nil, nil, err
 			}
--- a/go/pkg/client/tree.go
+++ b/go/pkg/client/tree.go
@@ -440,16 +440,18 @@ func (c *Client) ComputeMerkleTree(ctx context.Context, execRoot, workingDir, re
 			}
 			continue
 		}
-		if i.InputDigest != "" && i.Contents != nil {
+		if i.InputDigest != "" && len(i.Contents) > 0 {
 			return digest.Empty, nil, nil, errors.New("digest and file content cannot be provided for the same virtual input")
 		}
-		entry := uploadinfo.EntryFromBlob(i.Contents)
+		var entry *uploadinfo.Entry
 		if i.InputDigest != "" {
 			dg, err := digest.NewFromString(i.InputDigest)
 			if err != nil {
 				return digest.Empty, nil, nil, err
 			}
-			entry = uploadinfo.EntryFromFile(dg, normPath)
+			entry = uploadinfo.EntryFromRemoteFile(dg, normPath)
+		} else {
+			entry = uploadinfo.EntryFromBlob(i.Contents)
 		}
 		entry.IsVirtualInput = true
 		fs[remoteNormPath] = &fileSysNode{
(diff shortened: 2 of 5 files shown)


go/pkg/client/tree.go
```diff
@@ -440,9 +440,21 @@ func (c *Client) ComputeMerkleTree(ctx context.Context, execRoot, workingDir, re
 			}
 			continue
 		}
+		if i.InputDigest != "" && i.Contents != nil {
+			return digest.Empty, nil, nil, errors.New("digest and file content cannot be provided for the same virtual input")
+		}
+		entry := uploadinfo.EntryFromBlob(i.Contents)
```
mrahs, line 446: It's worth placing this in an `else` branch to make it a bit clearer that either bytes or hash is used. It also avoids the cost of creating unnecessary structs.

go/pkg/client/tree.go
```diff
@@ -440,9 +440,21 @@ func (c *Client) ComputeMerkleTree(ctx context.Context, execRoot, workingDir, re
...
+		if i.InputDigest != "" {
+			dg, err := digest.NewFromString(i.InputDigest)
+			if err != nil {
+				return digest.Empty, nil, nil, err
+			}
+			entry = uploadinfo.EntryFromFile(dg, normPath)
```
mrahs, line 452: Why mark it as a file blob rather than a bytes blob?

go/pkg/client/cas_upload.go
```diff
@@ -287,6 +287,9 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 			contextmd.Infof(ctx, log.Level(2), "Skipping upload of empty entry %s", ue.Digest)
 			continue
 		}
+		if ue.IsVirtualInput && ue.IsFile() {
+			return nil, 0, fmt.Errorf("virtual input with digest: %s provided, but does not exist in CAS", ue.Digest)
+		}
```
mrahs, line 292: This block is not clear to me. Why is it an error to have a missing virtual input?

go/pkg/client/tree.go
andusy: I wanted to preserve the current behaviour. We don't actually check that the Contents are not empty before creating an entry. It looks like we just check that the digest is empty before downloading https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/client/cas_upload.go#L286

go/pkg/client/cas_upload.go
andusy: VirtualInputs should exist on RBE and if it doesn't then it means something likely already went wrong before reaching this step. I don't think we should be uploading a possibly corrupted digest in this case.

go/pkg/client/tree.go
andusy: I'm not too familiar with specifics within the sdk. Is there a behavioural difference when marking an input as file vs a blob? Which would you suggest to use in this case?

mrahs: The path field should point to the absolute local path. Virtual inputs do not exist locally by definition so they should all be of type blob.

go/pkg/client/tree.go
mrahs: The `if-else` block would only alter the source of the entry. If the digest is not specified, the entry will still be created from the content regardless of its actual value.

go/pkg/client/tree.go
mrahs: Hmm, but now that we are changing that definition to include two more cases:
- files that exist locally but the digest for which is already computed
- files that do not exist locally but we know their digest and it should already exist remotely

I think the implementation should take all three cases into account.

andusy: Going to label the cases you provided to make it easier:
1. files that exist locally but the digest for which is already computed
2. files that do not exist locally but we know their digest and it should already exist remotely

(1) In this case VirtualInputs take precedence, correct? Meaning, since we ignore the local file this would technically just be treated as (2)?

mrahs: I agree. It's simpler to define just two disjoint cases:
1. bytes (no digest) that can be digested and uploaded if necessary
2. a digest (no bytes) that should exist remotely

go/pkg/client/tree.go
```diff
@@ -440,9 +440,23 @@ func (c *Client) ComputeMerkleTree(ctx context.Context, execRoot, workingDir, re
...
+		if i.InputDigest != "" && len(i.Contents) > 0 {
+			return digest.Empty, nil, nil, errors.New("digest and file content cannot be provided for the same virtual input")
+		}
+		var entry *uploadinfo.Entry
+		if i.InputDigest != "" {
+			dg, err := digest.NewFromString(i.InputDigest)
+			if err != nil {
+				return digest.Empty, nil, nil, err
+			}
+			entry = uploadinfo.EntryFromRemoteFile(dg, normPath)
```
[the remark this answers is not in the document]
mrahs: You're right. I read that wrong. It's usually the absolute path if I'm not mistaken.

go/pkg/client/cas_upload.go
```diff
@@ -287,6 +287,9 @@ func (c *Client) uploadUnified(ctx context.Context, entries ...*uploadinfo.Entry
 			contextmd.Infof(ctx, log.Level(2), "Skipping upload of empty entry %s", ue.Digest)
 			continue
 		}
+		if ue.IsVirtualFile() {
+			return nil, 0, fmt.Errorf("virtual input with digest: %s provided, but does not exist in CAS", ue.Digest)
```
mrahs, line 291: nit: 
```suggestion
			return nil, 0, fmt.Errorf("virtual input with digest %q provided, but does not exist in CAS", ue.Digest)
```

go/pkg/uploadinfo/entry.go
```diff
@@ -59,3 +65,14 @@ func EntryFromFile(dg digest.Digest, path string) *Entry {
...
 	}
 }
+
+// EntryFromVirtualFile creates an entry from a file not on disk.
+// The digest is expected to exist in the CAS.
+func EntryFromVirtualFile(dg digest.Digest, path string) *Entry {
+	return &Entry{
+		Digest:      dg,
+		Path:        path,
+		ueType:      ueBlob,
```
mrahs, line 75:
```suggestion
		ueType:      ueFile,
```

go/pkg/client/tree_test.go
```diff
@@ -1565,6 +1590,10 @@ func TestComputeMerkleTree(t *testing.T) {
 				t.Errorf("ComputeMerkleTree(...) = gave error %q, want success", err)
 			}
 			for _, ue := range inputs {
+				if ue.IsVirtualFile() {
+					gotBlobs[ue.Digest] = tc.digesttoBlob[ue.Digest]
```
mrahs, line 1594: Why is this needed?

go/pkg/uploadinfo/entry.go
andusy: Don't we want to use blobs for virtual files?

go/pkg/client/tree_test.go
andusy: chunker.New tries to:
 - read the content for blobs
 - read the file for files
 
Both of which won't be available in this case. So for testing purposes I'm just mapping the virtual inputs' digest back to the corresponding blob.

mrahs: Would it work if you excluded virtual lobs from `additionalBlobs` instead?

go/pkg/uploadinfo/entry.go
mrahs: My suggestion was from a consistency perspective. Since those are not meant to be uploaded, it doesn't matter what the type is.

go/pkg/client/tree_test.go
andusy: yes, but I would still have to add a condition here since `ueFile` would not be (and probably should not be) handled in `chunker.New()`.

mrahs: That's what I had in mind. Thanks.

go/pkg/client/tree_test.go
```diff
@@ -462,6 +462,7 @@ func TestComputeMerkleTree(t *testing.T) {
 		// blobs.
 		rootDir         *repb.Directory
 		additionalBlobs [][]byte
+		digesttoBlob    map[digest.Digest][]byte
```
mrahs, line 465: No longer needed.

go/pkg/client/tree_test.go
```diff
@@ -1462,6 +1463,30 @@ func TestComputeMerkleTree(t *testing.T) {
...
+				InputNodeProperties: map[string]*cpb.NodeProperties{"fooDir/foo": fooProperties},
+			},
+			rootDir: &repb.Directory{Directories: []*repb.DirectoryNode{
+				{Name: "barDir", Digest: barDirDgPb},
+				{Name: "fooDir", Digest: fooDirDgPb},
+			}},
+			additionalBlobs: [][]byte{fooDirBlob, barDirBlob},
+			digesttoBlob: map[digest.Digest][]byte{
+				fooDg: fooBlob,
+				barDg: barBlob,
```
mrahs, line 1482: leftover

go/pkg/uploadinfo/entry.go
```diff
@@ -10,6 +10,7 @@ import (
 const (
 	ueBlob = iota
 	uePath
+	ueFile
```
mrahs, line 13: Ah, my bad. I meant to type `uePath`. No need for a new type.

mrahs (approved):

PR #543, opened 2024-03-26
Update VirtualInputs proto with file properties

andusy:
- Add digest, mtime, and filemode to virtualinputs proto.
- Add mtime and filemode to virtualinputs struct.


go/api/command/command.proto
```diff
@@ -91,6 +91,12 @@ message VirtualInput {
   bool is_executable = 3;
   // Whether the input is an empty directory.
   bool is_empty_directory = 4;
+  // The digest of the virtual input that is expected to exist in the CAS.
+  string digest = 5;
+  // Mtime of the virtual input.
+  int64 mtime = 6;
+  // The virtual inputs' mode and permissions bits.
+  int32 filemode = 7;
```
mrahs, line 99: Should be `uint32` to match `fs.FileMode`.

go/api/command/command.proto
```diff
@@ -91,6 +91,12 @@ message VirtualInput {
   bool is_executable = 3;
   // Whether the input is an empty directory.
   bool is_empty_directory = 4;
+  // The digest of the virtual input that is expected to exist in the CAS.
+  string digest = 5;
+  // Mtime of the virtual input.
+  int64 mtime = 6;
```
mrahs, line 97: This probably needs more specification. Is it seconds millis or nanos? What about corss-filesystem support (e.g. one can do seconds only while another can do fractions)?

mrahs (commented):
Can you update the godocs for these structs? It currently says "don't exist on disk", but the definition now includes "may exist but shouldn't be accessed".

go/pkg/command/command.go
```diff
@@ -666,6 +672,9 @@ func inputSpecFromProto(is *cpb.InputSpec) *InputSpec {
 			Contents:         contents,
 			IsExecutable:     vi.IsExecutable,
 			IsEmptyDirectory: vi.IsEmptyDirectory,
+			Digest:           vi.Digest,
+			Mtime:            time.Unix(vi.Mtime.Seconds, int64(vi.Mtime.Nanos)),
```
mrahs, line 676: Shouldn't be `vi.Mtime.AsTime()`?

mrahs (commented):
Can you a unit test for the conversion between repb and comandpb?

mrahs (approved):

PR #544, opened 2024-03-27
Get input output list with symlinks

ywmei-brt1:
Add a function, GetIO, to return list of inputs and outputs. 

Note that symlinks with new name: my_sym_link (path), point to a old name: my_origional_file (target) will be represent as a string `my_sym_link->my_origional_file` (from path to target)

### DIFF
--- a/go/pkg/fakes/server.go
+++ b/go/pkg/fakes/server.go
@@ -28,7 +28,7 @@ import (
 	repb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
 	bsgrpc "google.golang.org/genproto/googleapis/bytestream"
 	bspb "google.golang.org/genproto/googleapis/bytestream"
-	anypb "google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/anypb"
 	dpb "google.golang.org/protobuf/types/known/durationpb"
 	tspb "google.golang.org/protobuf/types/known/timestamppb"
 )
@@ -302,6 +302,21 @@ func (f *InputFile) apply(ac *repb.ActionResult, s *Server, execRoot string) err
 	}
 	s.CAS.Put(bytes)
 	return nil
+}
+
+// InputSymlink to be made available to the fake action.
+type InputSymlink struct {
+	Path    string // newname
+	Content string
+	Target  string // oldname
+}
+
+// Apply puts the target file in the fake CAS and create a symlink in OS.
+func (ins *InputSymlink) apply(ac *repb.ActionResult, s *Server, execRoot string) error {
+	inf := InputFile{Path: ins.Target, Contents: ins.Content}
+	inf.apply(ac, s, execRoot)
+	// create a symlink from oldname to newname
+	return os.Symlink(filepath.Join(execRoot, ins.Target), filepath.Join(execRoot, ins.Path))
 }
 
 // OutputFile is to be added as an output of the fake action.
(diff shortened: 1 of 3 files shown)


ywmei-brt1:
Will separate the input field and output field into 

- input/output files, 
- input/output files symlinks,  
- input/output directories
- - input/output directories symlinks

ywmei-brt1:
Inputs are separated as 1) Paths, the 2) Symlinks, as there is no such thing as input directories;
Outputs are separated as 1) File 2) Dir, 3) File Symlinks, 4) Dir Symlinks.

go/pkg/tool/tool.go
```diff
@@ -848,3 +858,92 @@ func (c *Client) getActionResult(ctx context.Context, actionDigest string) (*rep
...
+}
+
+// GetIO returns the Inputs and Outputs of an action Digest.
+func (c *Client) GetIO(ctx context.Context, actionDigest string) (*IO, error) {
+	acDg, err := digest.NewFromString(actionDigest)
+	if err != nil {
+		return nil, fmt.Errorf("error creating action digest: %w", err)
+	}
+	actionProto := &repb.Action{}
+	// Get all the Inputs.
```
gkousik, line 892: nit: comments should always be about "why". The what is evident from the code :)

go/pkg/tool/tool.go
```diff
@@ -848,3 +858,92 @@ func (c *Client) getActionResult(ctx context.Context, actionDigest string) (*rep
 	}
 	return resPb, nil
 }
+
+// IO is a collection input root Digest, Inputs, and Outputs for an action.
+type IO struct {
```
gkousik, line 863: naming suggestion: FlatActionIO

Wdyt about having the struct be even more direct? Like:

```
type FlatActionIO {
    RootDg string
    InputPaths map[string]string
    InputPathSymlinks map[string]string
    OutputFiles map[string]string
    OutputDirs map[string]string
    OutputFileSymlinks map[string]string
    OutputDirSymlinks map[string]string
}
```

Why do we need separate structs for Inputs and Outputs?

go/pkg/tool/tool.go
```diff
@@ -848,3 +858,92 @@ func (c *Client) getActionResult(ctx context.Context, actionDigest string) (*rep
...
+// key is `<path>-><target>`, the value is the dg of the target file.
+type Outputs struct {
+	Files        map[string]string
+	Dirs         map[string]string
+	FileSymlinks map[string]string
+	DirSymlinks  map[string]string
+}
+
+// GetIO returns the Inputs and Outputs of an action Digest.
+func (c *Client) GetIO(ctx context.Context, actionDigest string) (*IO, error) {
```
gkousik, line 886: naming nit: FlattenActionIO ?

go/pkg/tool/tool.go
```diff
@@ -848,3 +858,92 @@ func (c *Client) getActionResult(ctx context.Context, actionDigest string) (*rep
...
+		return nil, fmt.Errorf("error getting input root digest: %w", err)
+	}
+	_, inputPaths, inputSymlinks, err := c.getInputTree(ctx, actionProto.GetInputRootDigest())
+	if err != nil {
+		return nil, err
+	}
+	// Get all the Outputs.
+	resPb, err := c.getActionResult(ctx, actionDigest)
+	if err != nil {
+		return nil, err
```
gkousik, line 907: For actions that failed, they wouldn't have an action result. Would it be better to just add a warning in case getting action result fails and leave the outputs as empty?

go/pkg/fakes/server.go
```diff
@@ -304,6 +304,26 @@ func (f *InputFile) apply(ac *repb.ActionResult, s *Server, execRoot string) err
 	return nil
 }
 
+// InputSymlink to be made available to the fake action.
+type InputSymlink struct {
+	Path    string // newname
```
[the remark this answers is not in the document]
ywmei-brt1: As disscussed offline, change field names to: Path, TargetContent, Target

PR #546, opened 2024-04-08
Document some typical usecases for remotetool.

rjobredeaux3:


go/README.md
```diff
@@ -1,9 +1,62 @@
...
+  --operation=execute_action \
+  COMMON_FLAGS
+  [--action_root=PATH|--digest=DIGEST] \
+  --path OUTPUT_PATH
+```
+- For `COMMON_FLAGS`, see the [above section](#common-flags).
+- An action digest formatted as `"{hash}/{size}"` can be provided directly using the `--digest` flag, or a previously downloaded action can be used with `--action_root`. The `--action_root` path should point to the `--path` of a previous `download_action` invocation. Specifically, the `--action_root` folder must contain a `cmd.textproto` and `ac.textproto` files.
+- `--path` is the destination where the outputs of the action will be downloaded
+
+#### Running a modified version of a downloaded action
```
mrahs, line 45: I don't quite remember the details, but I think it should be doable without uploading manually. `execute_action` operation should do that automatically if `action_root` is provided. I think the action is rebuilt in that case and `action_digest` argument is ignored.

mrahs (approved):

PR #552, opened 2024-04-15
remove unmaintained errors pkg

mrahs:
The errors package was deprecated in 2021 https://github.com/pkg/errors

I converted wrapped errors to the standard method assuming the stack trace was not intended to be included in these calls.

I opted to removing the stack trace from the calls that explicitly included it in favour of proper error wrapping and propagation. I haven't needed to look at stack traces so far, but I'm not sure if others rely on it. This can be re-added later if needed.


mrahs:
@danw FYI.
@ola-rozenfeld FYI in case the stack traces are actually needed.
@avidrissman FYI in case the stack traces are needed in the `cas` package.

avidrissman:
(I appreciate you bringing me in, but I made the change to the `cas` package because we were encountering them downstream. I have no other expertise or knowledge about the `cas` package.)

gkousik:
It would perhaps be useful to run a build with reclient, run some failing actions and ensure errors from the SDK are still being propagated as expected :D

gkousik (approved):

PR #556, opened 2024-04-25
Use longer timeouts for downloading and uploading data

ppluciennik:
By default only 20 seconds were used as a timeout.

Read and Write operations should be at least the same as BatchReadBlobs and BatchUpdateBlobs.

In practice it would be best to increase timeouts for Read and Write operations as streaming large objects (like 200-300MB) may take longer on a slow or busy network.

### DIFF
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -798,6 +798,8 @@ var DefaultRPCTimeouts = map[string]time.Duration{
 	"GetCapabilities":  5 * time.Second,
 	"BatchUpdateBlobs": time.Minute,
 	"BatchReadBlobs":   time.Minute,
+	"Read":             time.Minute,
+	"Write":            time.Minute,
 	"GetTree":          time.Minute,
 	// Note: due to an implementation detail, WaitExecution will use the same
 	// per-RPC timeout as Execute. It is extremely ill-advised to set the Execute


ppluciennik:
I'd propose to change given timeouts to 2 minutes or longer.

mrahs:
Thank you for the PR. It looks like the [`Read`](https://github.com/bazelbuild/remote-apis-sdks/blob/88db5fa9e4aecb5d4afabf2964d9e19c2b7b22d2/go/pkg/client/bytestream.go#L148) and [`Write`](https://github.com/bazelbuild/remote-apis-sdks/blob/88db5fa9e4aecb5d4afabf2964d9e19c2b7b22d2/go/pkg/client/bytestream.go#L74) calls are limited to the gRPC message size, which defaults to 4MB. The latency of these calls should not be affected by the blob size being streamed. I think the major factor in latency is how fast the server can prepare a response. Batch calls may take more time on the server side to prepare a response for. However, streaming calls are expected to be faster. The timeout here is multiplied by the number of `stream.Recv` call necessary to complete the streaming. Keeping it low trades throughput for latency because the workload is latency-sensitive (you typically want fast builds and won't tolerate slow execution servers).
I think that was the reasoning behind these defaults.

Can you provide more context around having higher defaults (given that the intended tradeoffs can be customized by overriding the defaults)?

ppluciennik:
You're right I've read code incorrectly, Read timeout is per 4MB blob.
I've increased timeouts on my side and it improved situation, however it's not ideal.

I've double checked metrics on server side and read/write p99 latencies are < 0.5s.
The only thing that comes into my mind is that many Read responses share a few grpc connections and on slower/busy network it times out and there are many retries.

ppluciennik:
I was mislead by p99.  Latency p100 revealed slow responses from backend.

PR #562, opened 2024-05-06
Add credshelper package to support the credentials helper interface

banikharbanda:
This adds all the credentials helper relevant code from
bazelbuild/reclient/internal/pkg/auth to the new credshelper package in
remote-apis-sdks with some changes:
- CredentialsHelper isn't an auth "Mechanism" (since we're only
  supporting one mechanism, we don't need to define a type for that)
- If no cache file is provided then it is assumed that no caching is
  required. If a valid cache file path is provided then it always caches
credentials.
- Updated oauth2 package version to support ReuseTokenWithExpiry - this
  had to be moved earlier in the Workspace file to override older
versions imported from elsewhere.


gkousik (commented):
Can you add a description to the change about "what the current CL adds" and what subsequent CLs will add?

Also, is the code copied over from re-client or has there been some modifications to it?

I guess without the subsequent change that actually uses the creds helper, I'm inclined to rubber-stamp this and review this code in the CL that actually uses it :) Would that be ok?

banikharbanda:
> Can you add a description to the change about "what the current CL adds" and what subsequent CLs will add?
> 
> Also, is the code copied over from re-client or has there been some modifications to it?
> 
> I guess without the subsequent change that actually uses the creds helper, I'm inclined to rubber-stamp this and review this code in the CL that actually uses it :) Would that be ok?

Added more detail to the PR description and commit message. And yes makes sense to review the final version of this which is actually used (and will look a bit different from what it is currently as well)

gkousik (approved):

PR #563, opened 2024-05-08
Add auxiliary metadata for cache hit actions

ywmei-brt1:
Add auxiliary metadata for cache hit actions as well. Previously we only have this for remote execution actions.

Test: run 2 actions with cmake, check the log to see the auxiliary metadata is avaliable for cache hit actions. https://paste.googleplex.com/5312315578908672


gkousik (approved):
Is it possible to add a test case for it?

PR #568, opened 2024-05-13
Split the remotetool into a separate library

gkousik:
In this CL, I'm creating an embeddedtool binary so that I can subsequently use it in reproxytool binary in re-client to implement the log file filtering functionalities. There's no behavior change associated with this CL.

Tested: Ran show_action and show_action_2 (error'ed out) on the new remotetool.

### DIFF
--- a/go/pkg/tool/embeddedtool.go
+++ b/go/pkg/tool/embeddedtool.go
@@ -1,0 +1,159 @@
+// Package tool in this file provides functions that can be embedded into a go binary
+// such that the binary now supports the various operations the remotetool
+// supports (e.g., show_action, upload_blob, etc).
+// A canonical usage of this library is the remotetool binary.
+package tool
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/outerr"
+
+	log "github.com/golang/glog"
+)
+
+var (
+	inputDigest  string
+	pathPrefix   string
+	overwrite    bool
+	actionRoot   string
+	execAttempts int
+	jsonOutput   string
+)
+
+// RegisterFlags registers the flags necessary for the embedded tool to work.
+func RegisterFlags() {
+	flag.StringVar(&inputDigest, "digest", "", "Digest in <digest/size_bytes> format.")
+	flag.StringVar(&pathPrefix, "path", "", "Path to which outputs should be downloaded to.")
+	flag.BoolVar(&overwrite, "overwrite", false, "Overwrite the output path if it already exist.")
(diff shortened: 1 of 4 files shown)


mrahs:
It sounds like you intend to import the new package from outside the `cmd` dir. Last I checked, `go` won't let you do that. A more suitable place would be `go/pkg` directory.

gkousik:
> It sounds like you intend to import the new package from outside the cmd dir. Last I checked, go won't let you do that. A more suitable place would be go/pkg directory.

ah, I intend to import the new package in reclient (`reproxytool`), but that's also inside the `cmd` dir. I tried and it seems to work.

mrahs:
> > It sounds like you intend to import the new package from outside the cmd dir. Last I checked, go won't let you do that. A more suitable place would be go/pkg directory.
> 
> ah, I intend to import the new package in reclient (`reproxytool`), but that's also inside the `cmd` dir. I tried and it seems to work.

Ah, it looks like there is no inherent restrictions on importing from the cmd directory. The tooling only enforces that for internal packages: [go.dev/doc/go1.4#internalpackages](http://go.dev/doc/go1.4#internalpackages)
OTOH, it's recommended for reusable packages to be outside the cmd directory: [go.dev/doc/modules/layout](http://go.dev/doc/modules/layout)

For the flags, it's not recommended to let packages define their own: https://google.github.io/styleguide/go/decisions#flags

go/pkg/tool/embeddedtool.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			log.Exitf("error uploading directory for path %s: %v", getPathFlag(), err)
+		}
+	},
+}
+
+// ValidateFlags validates the command line flags associated with this embedded remote tool.
+func ValidateFlags() {
+	if !flag.Parsed() {
+		flag.Parse()
```
mrahs, line 143: I think should not be done here now that the flags are the responsibility of the user.

mrahs (changes requested):

go/pkg/tool/embeddedtool.go
gkousik: Done

gkousik:
> Ah, it looks like there is no inherent restrictions on importing from the cmd directory. The tooling only enforces that for internal packages: [go.dev/doc/go1.4#internalpackages](http://go.dev/doc/go1.4#internalpackages) OTOH, it's recommended for reusable packages to be outside the cmd directory: [go.dev/doc/modules/layout](http://go.dev/doc/modules/layout)

Moved to pkg/tool as we discussed offline and got rid of the separate `embeddedtool` pkg.

> For the flags, it's not recommended to let packages define their own: https://google.github.io/styleguide/go/decisions#flags

Done!

mrahs (approved):

PR #572, opened 2024-05-16
 Update credshelper package

banikharbanda:
- Made externalTokenSource implement PerRPC creds interface as well
- Added headers and tests related to this

### DIFF
--- a/go/pkg/credshelper/credshelper.go
+++ b/go/pkg/credshelper/credshelper.go
@@ -3,6 +3,7 @@ package credshelper
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -83,9 +84,13 @@ type Credentials struct {
 
 // externaltokenSource uses a credentialsHelper to obtain gcp oauth tokens.
 // This should be wrapped in a "golang.org/x/oauth2".ReuseTokenSource
-// to avoid obtaining new tokens each time.
+// to avoid obtaining new tokens each time. It implements both the
+// oauth2.TokenSource and credentials.PerRPCCredentials interfaces.
 type externalTokenSource struct {
 	credsHelperCmd *reusableCmd
+	headers        map[string]string
+	expiry         time.Time
+	headersLock    sync.RWMutex
 }
 
 func buildExternalCredentials(baseCreds cachedCredentials, credsFile string, credsHelperCmd *reusableCmd) *Credentials {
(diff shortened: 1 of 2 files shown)


go/pkg/credshelper/credshelper.go
```diff
@@ -190,14 +215,14 @@ func NewExternalCredentials(credshelper string, credshelperArgs []string, credsF
...
-	tk, rexp, err := runCredsHelperCmd(credsHelperCmd)
+	_, tk, rexp, err := runCredsHelperCmd(credsHelperCmd)
 	if err != nil {
 		return nil, err
 	}
 	return buildExternalCredentials(cachedCredentials{token: tk, refreshExp: rexp}, credsFile, credsHelperCmd), nil
 }
 
-func runCredsHelperCmd(credsHelperCmd *reusableCmd) (*oauth2.Token, time.Time, error) {
+func runCredsHelperCmd(credsHelperCmd *reusableCmd) (map[string]string, *oauth2.Token, time.Time, error) {
```
gkousik, line 225: Can the `map, token, time` be wrapped into a return struct? Four return arguments from a function starts to get confusing.

banikharbanda: Done! Also addressed a lint warning about the json output structure name.

go/pkg/credshelper/credshelper.go
```diff
@@ -209,53 +240,50 @@ func runCredsHelperCmd(credsHelperCmd *reusableCmd) (*oauth2.Token, time.Time, e
...
 }
 
-// CredsHelperOut is the struct to record the json output from the credshelper.
-type CredsHelperOut struct {
-	Token         string `json:"token"`
-	Expiry        string `json:"expiry"`
-	RefreshExpiry string `json:"refresh_expiry"`
+// JSONOut is the struct to record the json output from the credshelper.
+type JSONOut struct {
+	Token         string            `json:"token"`
```
gkousik, line 250: I think go lint will complain if these fields aren't documented? I'm not sure (but it will only really show up as a problem when we import to Google3, so perhaps just better to document the fields now).

go/pkg/credshelper/credshelper.go
```diff
@@ -209,53 +240,50 @@ func runCredsHelperCmd(credsHelperCmd *reusableCmd) (*oauth2.Token, time.Time, e
...
+	Headers       map[string]string `json:"headers"`
+	Expiry        string            `json:"expiry"`
+	RefreshExpiry string            `json:"refresh_expiry"`
 }
 
-func parseTokenExpiryFromOutput(out string) (string, time.Time, time.Time, error) {
+func parseTokenExpiryFromOutput(out string) (credshelperOutput, error) {
```
gkousik, line 256: You can return a pointer to credsHelperOutput and avoid unnecessary copies, while also returning `nil` in case of errors.

I don't mind the copy too much since we call creds helper sparsely, but it would be ideal to return a `nil` pointer in case of errors than an empty struct.

gkousik (changes requested):

go/pkg/credshelper/credshelper.go
banikharbanda: I don't think it does, since it didn't complain when I imported the first credshelper package change which already had this.

go/pkg/credshelper/credshelper.go
banikharbanda: Done

go/pkg/credshelper/credshelper.go
gkousik: Sg then.

gkousik (approved):

PR #573, opened 2024-05-20
Ignore unexpected file modes

avidrissman:
Correct #549.

The intention was to skip the file with an unexpected mode, but the “nil, nil” return was actually done in a nested function.

Return ErrSkip to escape the nested function, and correctly return “nil, nil” from visitPath, as was intended.


avidrissman:
I was foolish with the original patch, assuming it was trivial enough that I didn’t need to test it. This patch has been tested in-place with [LUCI](https://chromium.googlesource.com/infra/luci/luci-go/), and it works.

avidrissman:
@mrahs Can you ptal? Thanks.

mrahs (approved):

PR #577, opened 2024-05-30
Fix refresh expiry code and add tests

banikharbanda:


go/pkg/credshelper/credshelper.go
```diff
@@ -166,6 +166,14 @@ func (c *Credentials) RemoveFromDisk() {
 	}
 }
 
+// refreshStatus checks refresh expiry of credentials in case a manual refresh is required.
+func (c *Credentials) refreshStatus() error {
+	if !c.refreshExp.IsZero() && c.refreshExp.Before(nowFn()) {
```
gkousik, line 171: The `!c.refreshExp.IsZero()` means that if refresh expiry isn't specified, we would assume the cached credential is valid indefinitely. Should we instead assume that we should always refresh it if refresh expiry isn't specified?

banikharbanda: Discussed offline and will keep 0 as no manual refresh needed indefinitely.

gkousik (approved):

PR #580, opened 2024-06-03
remove dependency on experimental gRPC API

### LINKED ISSUE #499: go/pkg/balancer uses experimental gRPC APIs
These APIs are labeled as [experimental](https://github.com/grpc/grpc-go/blob/87bf02ad24f6cc071d2553eb5d62332194bba1fe/balancer/balancer.go#L20), and they will be changed in the near-ish future.  Details are in https://github.com/grpc/grpc-go/issues/6472.  Note that any package that uses gRPC's experimental APIs should _itself_ be considered experimental, as a change to them would break compilation for the entire package that uses them.  So we do NOT recommend the use of these features in any other libraries or applications that aren't tolerant to breakages like this.

https://github.com/googleapis/google-cloud-go may have another solution to the same type of problem, which is that it creates multiple channels and load balances between them.  See https://github.com/googleapis/google-api-go-client/blob/caea95689f82049822552cd649765335123831e0/transport/grpc/pool.go#L30 for details.

mrahs:
The custom balancer depends on experimental APIs from grpc-go. It has since been replaced with a simple round robin balancer that has been working well in production.

Fixes #499


gkousik (approved):

gkousik:
Would we need any reproxy changes in how the SDK is initialized when importing this new SDK version?

mrahs:
> Would we need any reproxy changes in how the SDK is initialized when importing this new SDK version?

I'm reusing `MaxConcurrentRequests` which was used to set the pool size for the removed balancer. I don't anticipate any changes to make it work. However, I removed `MaxConcurrentStreams` and this could cause breaks if it was provided as an argument. I'll have to check before the next release.

gkousik:
Ok sg, just wanted us to be aware of it (if not, perhaps better to just do the import now in reclient?)

PR #586, opened 2024-07-10
Report an error on unsupported codec or invalid zstd blob

ppluciennik:
Method BatchDownloadBlobsWithStats silently ignores decoding error or unsupported codec due to lack of incrementing numErrs.

### DIFF
--- a/go/pkg/client/cas_download.go
+++ b/go/pkg/client/cas_download.go
@@ -256,12 +256,16 @@ func (c *Client) BatchDownloadBlobsWithStats(ctx context.Context, dgs []digest.D
 					if err != nil {
 						errDg = r.Digest
 						errMsg = err.Error()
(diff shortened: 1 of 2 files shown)


ppluciennik:
I'm also not sure,  but probably `allRetriable` should be also set to false.

mrahs:
Thanks for the PR. I think it's reasonable to not retry decoding errors since they come after a successful transmission. It would be great if you can add a test case.

ppluciennik:
I'm sorry for a late response, but I was on vacation.  I've added *TestBatchDownloadBlobsBrokenCompression* test to PR.

mrahs:
Thank you for updating the PR. Nice work. (Sorry for the late response on my end as well..)

mrahs (approved):

PR #587, opened 2024-07-25
Fix incorrect usage of `binary.Write()` in `TestUpload()`.

ifoox:
The call to `binary.Write()` was previously failing because it was being passed a variable length encoded type (int) rather than a fixed length encoding type. This caused the 2000 blobs tests to actually have 0 blobs. Also fix concurrency expectation for the 2000 blobs test.

Tested:
  Ran unit tests.

### DIFF
--- a/go/pkg/client/cas_test.go
+++ b/go/pkg/client/cas_test.go
@@ -724,12 +724,19 @@ func TestUpload(t *testing.T) {
 	t.Parallel()
 	var twoThousandBlobs [][]byte
 	var thousandBlobs [][]byte
-	for i := 0; i < 2000; i++ {
+	for i := uint16(0); i < 2000; i++ {
 		var buf = new(bytes.Buffer)
-		binary.Write(buf, binary.LittleEndian, i)
+		err := binary.Write(buf, binary.LittleEndian, i)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


mrahs:
Curious how you noticed that the blobs were 0 (assuming you mean `twoThousandBlobs` was empty)? The test was passing before; should it also be updated to capture this issue?

ifoox:
I was looking at the test to trace how uploads work (as part of looking at fixing compression stats, internal bug b/312446641), so I added some printfs to the test and noticed something weird happening with 2000 blobs tests, and started debugging from there.

What ended up happening is that `twoThousandBlobs` was a slice of 2k empty slices ([[],[],[]..., 2k times]), and `thousandBlobs` was a slice of 1k empty slices.

I'm not sure the test needs to be updated, since it just happened that the test setup was incorrect relative to the original intention but still a valid way test. What the test was testing was:
 - try to do an upload of 2k empty blobs, with the empty blob already existing in the CAS
 
This is a valid test, it's just not what the original intention of the test was.

mrahs:
I see. Good catch!

On a second look, you already added error checking which should effectively capture the failure prior to your fix.

mrahs (approved):

PR #588, opened 2024-07-25
Fix typo for CorrelatedInvocationsID

ywmei-brt1:
This cl make correction to the identifier's CorrelatedInvocationsID field name to make it match the proto. It was CorrelatedInvocationID before.

We need to have CorrelatedInvocationsID in the ToProto method to deduplicate builds that have the same invocation id.


go/pkg/contextmd/contextmd.go
```diff
@@ -177,33 +177,33 @@ func mergeSet(set map[string]struct{}) string {
...
-			rem := (excess - diff + 1) / 2
-			m.InvocationID = m.InvocationID[:len(m.InvocationID)-rem]
-			m.ActionID = m.ActionID[:len(m.ActionID)-rem]
+	truncateStrings(limit, &m.ActionID, &m.InvocationID, &m.CorrelatedInvocationsID, &m.ToolName, &m.ToolVersion)
+	return m
+}
+
+// truncateStrings remove one char from the longest str at a time until the
+// total length of all the strings is less than limit value.
+func truncateStrings(limit int, inputs ...*string) {
```
mrahs, line 191: Any reason to extract the code to this helper function?

ywmei-brt1: I have a helper function here because I think 

1) how we truncate strings (say, maybe we want to, if possible, only remove the entire action_id, without touching the invocation_id because  invocation_id seems more important, and action_digest is import, but no one works with action_id), and

 2) what strings need to be truncated, and what strings we will not touch (say, maybe the tool version, we might don't want it to be 0.15[missing digits]),
 
 might subject to change....

I don't have much context about why and in which case we will have this header limit overflow problem, and I am not sure what side effect will be after we truncated the strings....

I think lets have a helper function here for now, and until we see a real break, we will come back here and refactor the truncateString logic.

Basically, the parent function indicates we must truncate String when it overflow (this part is solid)
the helper function decides which strategy we use to truncate strings (this we don't have enough info to make decision now, and might subject to change, so I move this part out.)

ywmei-brt1:
I have tested with a re-client chrome build, this sdk change works as expected, and can pass the RBE_correlated_invocations_id value to the rpl log.

go/pkg/contextmd/contextmd.go
mrahs: Ack

mrahs, line 191: Curious why you're passing strings by reference here? Strings in Go are backed by slices, which are cheap to copy because the underlying array is shared.

go/pkg/contextmd/contextmd.go
```diff
@@ -177,33 +177,33 @@ func mergeSet(set map[string]struct{}) string {
...
+	total := 0
+	for _, s := range inputs {
+		total += len(*s)
+	}
+	if total <= limit {
+		return
+	}
+	for total > limit {
```
mrahs, line 199: This loop might end up being expensive considering it's executed at least once for every action. I think it's safer to limit this PR to what's needed to satisfy its title and leave this refactor to another one.

go/pkg/contextmd/contextmd.go
ywmei-brt1: I want to pass m's fields here, and modify them in place (basically modify m in place).

say, if I pass strings here, and the interface is `func truncateStrings(limit int, inputs ...string) outputs []string{}`

I will end up with very long code on the caller's side. I was keep thinking it over and over for about an hour this morning, didn't find out a better solution than pass in ...*string.

```
outputs := truncateStrings(32, m.a, m.b, m.c, m.d, m.e)
m.a = outputs[0]
m.b = outputs[1]
m.c = outputs[2]
m.d = outputs[3]
m.e = outputs[4]
return m
```

mrahs (approved):

PR #590, opened 2024-08-01
Update doc for how to re-run an action with another RBE instance.

ywmei-brt1:
This CL update the readme file to show how to re-run an action on another RBE instance.

This CL also fix the flag conflict issue, for execute_action, one and only one of these two flags should be set: [--action_root=PATH|--digest=DIGEST].

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -12,6 +12,14 @@ Flags that describe the RBE instance used and the authentication method are comm
 
 `--alsologtostderr` and `-v VERBOSITY_LEVEL` can be used to tune the verbosity and location of the output, e.g. `--alsologtosrderr -v 1`.
 
+For example, the common flags could be:
+
+```
+    --service remotebuildexecution.googleapis.com:443 \
+    --use_application_default_credentials=true \
+    --alsologtostderr \
+    --v 1
+```
 #### Downloading an action's inputs and metadata
 ```
 bazelisk run //go/cmd/remotetool -- \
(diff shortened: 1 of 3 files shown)


go/README.md
```diff
@@ -42,6 +50,36 @@ bazelisk run //go/cmd/remotetool -- \
 - An action digest formatted as `"{hash}/{size}"` can be provided directly using the `--digest` flag, or a previously downloaded action can be used with `--action_root`. The `--action_root` path should point to the `--path` of a previous `download_action` invocation. Specifically, the `--action_root` folder must contain a `cmd.textproto` and `ac.textproto` files.
 - `--path` is the destination where the outputs of the action will be downloaded
 
+#### Re-running a downloaded action remotely on a different RBE instance
+
+If you want to download an action from an RBE instance `FOO`, and remotely 
+execute it on another RBE instance `BAR` (probably because you only have the 
+permission to download actions from `FOO`, but lack of permission to don't have 
+the permission to execute an action with instance `FOO`). In this case, you want
```
mrahs, line 58:
```suggestion
permission to download actions from `FOO`, but don't have 
the permission to execute that instance). In this case, you want
```

go/pkg/tool/embeddedtool.go
```diff
@@ -157,3 +159,13 @@ func getPathFlag() string {
 	}
 	return pathPrefix
 }
+
+func validateActionRootAndDg() (string, string) {
+	if inputDigest != "" && actionRoot != "" {
+		log.Exitf("either specify --digest or --action_root, should not set both of them together.")
+	}
+	if inputDigest == "" && actionRoot == "" {
+		log.Exitf("either specify --digest or --action_root, at least one of these two flag must be set.")
```
mrahs, line 168:
```suggestion
		log.Exitf("either specify --digest or --action_root, one of these two flag must be set, but not both.")
```

go/pkg/tool/embeddedtool.go
```diff
@@ -157,3 +159,13 @@ func getPathFlag() string {
 	}
 	return pathPrefix
 }
+
+func validateActionRootAndDg() (string, string) {
```
mrahs, line 163: nit: `validate` implies no return values, but validation can be expected as a side effect for a function that does return values.
```suggestion
func inputDgAndActionRoot() (string, string) {
```

go/pkg/tool/tool.go
```diff
@@ -656,6 +656,7 @@ func (c *Client) ExecuteAction(ctx context.Context, actionDigest, actionRoot, ou
 	case command.LocalErrorResultStatus:
 		oe.WriteErr([]byte(fmt.Sprintf("Local error: %v.\n", ec.Result.Err)))
 	}
+	fmt.Printf("Result: %+v\n", ec.Result)
```
mrahs, line 659: Is this intended or a leftover from testing?

mrahs (approved):

go/pkg/tool/tool.go
ywmei-brt1: This is intended because it does not show me if the action is `cache hit` or `remote execution` or `remote error` in the terminal when I run the action on another RBE instance. I think we should printout the result here.

BTW, I eventually figure out, by design, this remotetool does not use cache hit, it always remote execute a action without cache hit.

PR #591, opened 2024-08-06
Add flag to set symlink following behavior for uploading directory

amonshiz:
We would like to be able to use `upload_dir` in a situation that makes heavy use of symlinks. However, the previous implementation did not respect symlinks and instead resulted in resolving the symlink to the actual path. This is problematic because it means that duplicated data was going to be recreated on a download.

This is solved by introducing a `symlink_behavior` argument that accepts `unspecified` (default), `preserve`, and `resolve` which map to the `SymlinkBehaviorType` enum. This allows the caller to specify how to handle symlinks and ensure that later downloads match exactly what the upload was. This also appears to reduce the data uploaded.

## Test

```bash
bash-5.2$ pwd
/Users/amonshiz/Developer/remote-apis-sdks/go/cmd/remotetool
bash-5.2$ for kind in none unspecified UnspecifiedSymlinkBehavior resolve ResolveSymlink preserve PreserveSymlink; do
    echo "Symlink behavior: $kind"
    dd if=/dev/urandom of=tmp/versions/random_file bs=1024 count=10 &> /dev/null
    ./remotetool_arm64 \
        --operation=upload_dir \
        -service localhost:8980 \
        -service_no_auth \
        -use_rpc_credentials=false \
        -service_no_security \
        --path $(realpath tmp) \
        $([ "$kind" = "none" ] && echo '' || echo "-symlink_behavior=$kind") \
        --json=output.json &> /dev/null
    cat output.json
    echo ""
done
Symlink behavior: none
{
  "InputFiles": 2,
  "InputDirectories": 2,
  "InputSymlinks": 0,
  "TotalInputBytes": 20737,
  "RootDigest": {
    "Hash": "c12e994e2fa5b6e72fc45d66c8a756ec0ac59fc1468e5348ea9b2a3bdfddcf00",
    "Size": 171
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10497,
  "BytesCacheMisses": 10497,
  "Error": ""
}
Symlink behavior: unspecified
{
  "InputFiles": 2,
  "InputDirectories": 2,
  "InputSymlinks": 0,
  "TotalInputBytes": 20737,
  "RootDigest": {
    "Hash": "8b0d6b9f687b459dd176909ec244a5139028d695a012ecbcffa2f02b317c988d",
    "Size": 171
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10497,
  "BytesCacheMisses": 10497,
  "Error": ""
}
Symlink behavior: UnspecifiedSymlinkBehavior
{
  "InputFiles": 2,
  "InputDirectories": 2,
  "InputSymlinks": 0,
  "TotalInputBytes": 20737,
  "RootDigest": {
    "Hash": "aa29e6cea24a50379fbd03d60f712f1c6124f6ed711b808439f78d6b07197b41",
    "Size": 171
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10497,
  "BytesCacheMisses": 10497,
  "Error": ""
}
Symlink behavior: resolve
{
  "InputFiles": 2,
  "InputDirectories": 2,
  "InputSymlinks": 0,
  "TotalInputBytes": 20737,
  "RootDigest": {
    "Hash": "9c5004ba3ec36e502b79dbcbbcf31a8bcd31d06e0cb41e9d4d6bfa30c4434350",
    "Size": 171
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10497,
  "BytesCacheMisses": 10497,
  "Error": ""
}
Symlink behavior: ResolveSymlink
{
  "InputFiles": 2,
  "InputDirectories": 2,
  "InputSymlinks": 0,
  "TotalInputBytes": 20737,
  "RootDigest": {
    "Hash": "38a6711760acc6933141cc92ad6f8224fa60c0a61b4fa4f796c48c0a330d9d30",
    "Size": 171
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10497,
  "BytesCacheMisses": 10497,
  "Error": ""
}
Symlink behavior: preserve
{
  "InputFiles": 1,
  "InputDirectories": 2,
  "InputSymlinks": 1,
  "TotalInputBytes": 10448,
  "RootDigest": {
    "Hash": "e9cb3d11a60ede62cabaf9dff7e9d9f3817af0d06d3660294504beed1680fcd3",
    "Size": 122
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10448,
  "BytesCacheMisses": 10448,
  "Error": ""
}
Symlink behavior: PreserveSymlink
{
  "InputFiles": 1,
  "InputDirectories": 2,
  "InputSymlinks": 1,
  "TotalInputBytes": 10448,
  "RootDigest": {
    "Hash": "d953e212c7db21b1b2e35a54c2f1de293c7d458b2b9daa075ace209a1563de2e",
    "Size": 122
  },
  "CountBlobs": 3,
  "CountCacheMisses": 3,
  "BytesTransferred": 10448,
  "BytesCacheMisses": 10448,
  "Error": ""
}
```


go/pkg/command/command.go
```diff
@@ -71,6 +71,17 @@ func (s SymlinkBehaviorType) String() string {
 	return fmt.Sprintf("InvalidSymlinkBehaviorType(%d)", s)
 }
 
+func SymlinkBehavior(s string) SymlinkBehaviorType {
+  switch s {
+  case "resolve":
```
ola-rozenfeld, line 76: nit: allow the values specified in `symlinkBehaviorType` as well.

go/pkg/tool/embeddedtool.go
```diff
@@ -29,6 +30,7 @@ var (
 func RegisterFlags() {
 	flag.StringVar(&inputDigest, "digest", "", "Digest in <digest/size_bytes> format.")
 	flag.StringVar(&pathPrefix, "path", "", "Path to which outputs should be downloaded to.")
+  flag.StringVar(&symlinkBehavior, "symlink_behavior", "unspecified", "For upload_dir: how to handle symlinks. One of 'unspecified', 'resolve', 'preserve'.")
```
ola-rozenfeld, line 33: Let's set the current behavior as default.

go/pkg/tool/embeddedtool.go
```diff
@@ -119,7 +121,7 @@ var RemoteToolOperations = map[OpType]func(ctx context.Context, c *Client){
 		}
 	},
 	uploadDir: func(ctx context.Context, c *Client) {
-		us, err := c.UploadDirectory(ctx, getPathFlag())
+		us, err := c.UploadDirectory(ctx, getPathFlag(), getSymlinkBehaviorType())
```
ola-rozenfeld, line 124: nit: use the function you added to Command to parse the string as enum.

go/pkg/tool/tool.go
```diff
@@ -338,9 +338,10 @@ type UploadStats struct {
 }
 
 // UploadDirectory uploads a directory from the specified path as a Merkle-tree to the remote cache.
-func (c *Client) UploadDirectory(ctx context.Context, path string) (*UploadStats, error) {
+func (c *Client) UploadDirectory(ctx context.Context, path string, symlinkBehavior string) (*UploadStats, error) {
```
ola-rozenfeld, line 341: This should get the enum value already.

go/pkg/tool/embeddedtool.go
```diff
@@ -157,3 +159,13 @@ func getPathFlag() string {
...
 	return pathPrefix
 }
+
+func getSymlinkBehaviorType() string {
+  switch symlinkBehavior {
+  case "preserve", "resolve", "unspecified":
+    return symlinkBehavior
+  }
+
+  log.Exitf("--symlink_behavior must be specified.")
```
ola-rozenfeld, line 169: If you want to validate this, it should be added to [validateFlags](https://github.com/bazelbuild/remote-apis-sdks/blob/719a5dd43ab6b3f125211b216c48c1e8ec1b4354/go/pkg/tool/embeddedtool.go#L141) instead.

ola-rozenfeld (commented):
Please check go format when done (this is why presubmits are failing). Thank you!

go/pkg/command/command.go
amonshiz: Will do. Is it fine to leave the shorter strings as acceptable too or only allow the full names?

go/pkg/tool/embeddedtool.go
amonshiz: I think this is the current behavior? The original call in `ComputeMerkleTree` did not pass any value for the `SymlinkBehavior` property of the `command.InputSpec` which I _thnk_ means that the default enum value will be assigned which would be the `iota` case (`UnspecifiedSymlinkBehavior`) here? https://github.com/bazelbuild/remote-apis-sdks/blob/719a5dd43ab6b3f125211b216c48c1e8ec1b4354/go/pkg/command/command.go#L56

go/pkg/tool/embeddedtool.go
amonshiz: Is it okay to add the dependency on `command` here? I didn't want to introduce a new dep unnecessarily and potentially break some boundaries. Will do!

go/pkg/tool/embeddedtool.go
amonshiz: Will do!

go/pkg/tool/embeddedtool.go
ola-rozenfeld: Right, but that value is not handled [specifically](https://github.com/bazelbuild/remote-apis-sdks/blob/719a5dd43ab6b3f125211b216c48c1e8ec1b4354/go/pkg/client/tree.go#L94) by the client, so as is with your new validation this would mean the flag now becomes required, while it didn't use to be. So resolve should be the default to keep the current behavior, I think.

ola-rozenfeld (approved):

go/pkg/tool/embeddedtool.go
amonshiz: Discussed offline and I had misunderstood. The additional of `log.Exitf` was confusing and not correct, so removed.

go/pkg/tool/embeddedtool.go
amonshiz: Decided against validating and allowing anything but the following to result in `UnspecifiedSymlinkBehavior`:
- `resolve`
- `ResolveSymlink`
- `preserve`
- `PreserveSymlink`

go/pkg/command/command.go
```diff
@@ -71,6 +71,17 @@ func (s SymlinkBehaviorType) String() string {
 	return fmt.Sprintf("InvalidSymlinkBehaviorType(%d)", s)
 }
 
+func SymlinkBehavior(s string) SymlinkBehaviorType {
```
mrahs, line 74: Nit: the name sounds more like a type. I'd use `ParseSymlinkBehavior` which is akin to `strconv.ParseInt`. `SymlinkBahaviorFromString` is another reasonable name which is akin to `uuid.FromString`.

mrahs:
Thank you for the PR and thank you @ola-rozenfeld for the review.

go/pkg/command/command.go
mrahs: Ah, I've just noticed there is already a `symlinkBehaviorFromProto` in this file.

mrahs (approved):

PR #601, opened 2024-10-29
Support adding gRPC metadata to outgoing RPCs

### LINKED ISSUE #599: Feature: additional HTTP headers for remote requests
This is a feature request for something equivalent to Bazel's `--remote_header`. Something like:

```
--remote_header=x-engflow-resource-tag=cost_center=moonshot
```

When the SDK is configured with the flag above, outgoing HTTP requests would have the header `x-engflow-resource-tag` set to `cost_center=moonshot`.

Use case: EngFlow supports a few custom headers. `x-engflow-resource-tag` lets clients add tags to their requests for cost analysis. For example, this could be used to distinguish costs of two teams that share the service, or a new toolchain that might use more execution resources.

jayconrod:
Added the RemoteHeaders field to DialParams, allowing the client to provide a map[string]string of headers. These get attached to outgoing RPCs with client interceptors.

Added the -remote_headers flag to let users configure this on the command line.

Fixes #599


jayconrod:
Happy to make a change, but what do you have in mind?

My goal with #599 is to be able to pass gRPC metadata headers from the reclient command line in the same format as Bazel's `--remote_headers`. Our customers use that with Bazel to tag requests for resource usage accounting, and they'd like to do the same with reclient. For example, that lets them measure the usage of a particular team, or check whether upgrading to a new toolchain lowers costs. For example, they might set something like:

```
--remote_header=x-engflow-resource-tag=project=fusion_reactor
--remote_header=x-engflow-resource-tag=cost_center=moonshot
```

As I understand, the `contextmd` package marshals info about the client into the `RequestMetadata` proto, then sends the proto with the `build.bazel.remote.execution.v2.requestmetadata-bin` metadata header.

mrahs:
That's fair. The new flag provides a simple way to pass headers through to the server and I cannot think of a reason for the client to want to inspect those headers.

jayconrod:
I apologize, one more change coming to this PR: I realized we may need to set the same header multiple times. I'll change the flag to `map[string][]string` (matching gRPC's `metadata.MD`) and add some flag parsing code for that.

mrahs:
> I apologize, one more change coming to this PR: I realized we may need to set the same header multiple times. I'll change the flag to `map[string][]string` (matching gRPC's `metadata.MD`) and add some flag parsing code for that.

No worries. I was actually going to ask if you can confirm this PR works as intended in test or staging if possible.

jayconrod:
> No worries. I was actually going to ask if you can confirm this PR works as intended in test or staging if possible.

I got this working against a local cluster, adding this line to my `reproxy.cfg` file:

```
remote_headers=x-engflow-resource-tag=a=b,x-engflow-resource-tag=c=d
```

I confirmed in a debugging that we get `{"a": "b", "c": "d"}` on the server.

I'm glad you mentioned this actually because I realized it was an error for a key-value pair to have multiple `=` characters, as above. I relaxed the flag parsing at bit to cut on the first `=` but allow `=` within the value. Let me know if that's not okay, we could probably allow `:` or something internally. It feels better to allow more characters in the value though.

PR #604, opened 2024-11-14
Add wrapper that can be used to plug bazel style credshelper into remote-apis-sdks

banikharbanda:


gkousik:
Can you add documentation to the CL on how it can be used with the Bazel creds helper? Any example usage would also help!

gkousik (approved):

PR #607, opened 2024-12-17
Prevent side effects in MergeMetadata by creating new instances

### LINKED ISSUE #606: Unexpected Side Effects in MergeMetadata Implementation
I have identified that the current implementation of the `MergeMetadata` function in remote-apis-sdk directly modifies input pointer arguments, which can cause unexpected side effects for SDK users.

## Problem Details
The `MergeMetadata` function currently modifies the first metadata pointer in its arguments. This behavior is problematic for several reasons:

**SDK Usability**:
- Users must create deep copies of inputs to avoid unintended side effects
- This creates unnecessary overhead for SDK users
**Immutability Principles**:
- Modern API design favors immutable arguments
- This is particularly important for shared data structures like metadata

meithon:
This PR modifies the `MergeMetadata` function to create new instances instead of modifying input arguments, preventing unexpected side effects for SDK users.

fixes #606

### DIFF
--- a/go/pkg/contextmd/contextmd.go
+++ b/go/pkg/contextmd/contextmd.go
@@ -125,15 +125,20 @@ func MergeMetadata(metas ...*Metadata) *Metadata {
 		return &Metadata{}
 	}
 
-	md := metas[0]
 	actionIds := make(map[string]struct{}, len(metas))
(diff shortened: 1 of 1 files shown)


mrahs (approved):
Thank you for the PR!

Issue #610, opened 2024-12-30
Unexpected OutputFiles Modification Due to Shallow Copy in Command#ToREProto()


alsu4r:
I am using ReClient (https://github.com/bazelbuild/reclient) as the client and an RBE Service that supports RE API v2.1 or higher for Android RBE builds. During the build process, I encountered an issue where OutputFiles were unexpectedly modified. While debugging, I identified that this issue was caused by a shallow copy resulting from the use of append() in the Command#ToREProto() function in the remote-apis-sdk. Since Go's append can result in a shallow copy depending on the slice's capacity, shouldn't it be explicitly modified to ensure a deep copy?

I believe modifying the remote-apis-sdk code is a more appropriate solution to this issue than addressing it in ReClient, so I made adjustments to the SDK code. Below is the modification I tested.

There are many ways to address this issue, but this is how I resolved it: https://github.com/bazelbuild/remote-apis-sdks/blob/master/go/pkg/command/command.go#L600
+        cmdPb.OutputPaths = append(append([]string{}, c.OutputFiles...), c.OutputDirs...)

mrahs:
Good catch. The conversion must produce a deep clone. I'd allocate a properly sized slice before filling it up from both sources to avoid potential resizes. Please feel free to post a PR for the remote-apis-sdks.

PR #611, opened 2025-01-07
Fix unexpected OutputFiles changes in Command.ToReProto due to shallow copy in Go

alsu4r:
Explicitly copy the data to prevent unintended changes due to shallow copy occurring based on the characteristics of slices in Golang. This is mentioned in Issue: [610](https://github.com/bazelbuild/remote-apis-sdks/issues/610)


alsu4r (commented):
I think the method @ywmei-brt1 suggested is better from a performance perspective. I have made the changes and updated the patch.

ywmei-brt1 (approved):
Thank you very much for catching this for us.

PR #612, opened 2025-01-10
Add option to replace ExecutionMedatadata.AuxiliaryMetadata for tests

bentekkie:


go/pkg/fakes/server.go
```diff
@@ -473,3 +473,12 @@ func (c ExecutionCacheHit) apply(ac *repb.ActionResult, s *Server, execRoot stri
 	s.Exec.Cached = bool(c)
 	return nil
 }
+
+// AuxiliaryMetadata replaces ExecutionMetadata.AuxiliaryMetadata
+type AuxiliaryMetadata []*anypb.Any
+
+// Apply puts the action auxiliary metadata in the given ActionResult the beginning of the list.
+func (o AuxiliaryMetadata) apply(ac *repb.ActionResult, s *Server, execRoot string) error {
+	ac.ExecutionMetadata.AuxiliaryMetadata = o
```
ywmei-brt1, line 482: I am not sure if this is possible or does it worth it, but do you think we can have a deep copy here?

like: 

```
ac.ExecutionMetadata.AuxiliaryMetadata = a_deep_copy_of(o)
```

ywmei-brt1 (approved):

PR #628, opened 2025-08-15
Upload channel of blobs

fangism:
Re-use the Client.Upload implementation for batching
uploads of files with direct support for uploading blobs.

Change-Id: I0d79ccfc77feafeed92058e65121b31fbaf68055

### DIFF
--- a/go/pkg/cas/upload.go
+++ b/go/pkg/cas/upload.go
@@ -90,6 +90,15 @@ type UploadInput struct {
 	u                   *uploader
 }
 
+// UploadBlob specifies an in-memory blob to upload.
+type UploadBlob struct {
+	// Title is only used for diagnostic messages.
+	Title string
+
+	// Contents is the blob to upload.
+	Contents []byte
+}
+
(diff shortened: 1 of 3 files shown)


go/pkg/cas/upload.go
```diff
@@ -375,21 +384,55 @@ func (c *Client) Upload(ctx context.Context, opt UploadOptions, inputC <-chan *U
...
 		}()
 
-		for {
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case in, ok := <-inputC:
-				if !ok {
-					return nil
+		ueg, ctx := errgroup.WithContext(ctx)
```
jayzhuang, line 387: I think here you want to create a `gCtx` and use it instead of `ctx` inside the goroutines you launch with `ueg.Go`.

fangism: as discussed, using a non-cancelling `errgroup.Group` fixes it.  Uploaded.

fangism:
This is now ready, but requires #629 to land first to fix tests.

PR #629, opened 2025-08-15
I think we don't have any test to protect this repo

ywmei-brt1:


fangism (approved):
I can confirm, `bazel test ...` passes with this change.

PR #630, opened 2025-09-24
credshelper accepts bazel credhelper compat GetCredentialsResponse json.

ukai:
It would be nice if we can use the same credential helper with bazel's
https://github.com/EngFlow/credential-helper-spec/blob/main/spec.md

This CL allows credshelper package parses GetCredentialsResponse json message of bazel credential helper.


gkousik:
Looks good. Given the fact that none of the current unit tests break, I'm assuming reclient creds helper also works.

PR #655, opened 2026-05-19
client: validate OutputDirectory.Path before os.RemoveAll in DownloadActionOutputs

adilburaksen:
## Summary

`DownloadActionOutputs()` deletes existing output directories before downloading
action outputs. The deletion path was not validated — `dir.Path` from an
untrusted `ActionResult` was passed directly to `os.RemoveAll(filepath.Join(outDir, dir.Path))`.

A malicious or compromised remote execution/cache service can return a path like
`../victim`, causing deletion of files or directories outside `outDir`. This
bypasses the fix in #649 because the unsafe `RemoveAll()` call happens before
`DownloadOutputs()` validates the flattened output paths.

## Fix

Use the existing `getAbsPath()` helper (introduced in #649) to apply the same
boundary check to the deletion path before calling `os.RemoveAll()`.

### DIFF
--- a/go/pkg/client/cas_download.go
+++ b/go/pkg/client/cas_download.go
@@ -522,7 +522,11 @@ func (c *Client) DownloadActionOutputs(ctx context.Context, resPb *repb.ActionRe
 	}
 	// Remove the existing output directories before downloading.
 	for _, dir := range resPb.OutputDirectories {
-		if err := os.RemoveAll(filepath.Join(outDir, dir.Path)); err != nil {
+		absPath, err := getAbsPath(outDir, dir.Path)
+		if err != nil {
+			return nil, fmt.Errorf("invalid output directory path %q: %v", dir.Path, err)
+		}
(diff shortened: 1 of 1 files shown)


adilburaksen:
Gentle ping after ~2 weeks — validates OutputDirectory.Path before os.Read to prevent a path-traversal write. CI is green. Happy to adjust the validation or add a test if you'd like.

adilburaksen:
@tjgq could you take a look when you get a chance? It validates OutputDirectory.Path before the os.RemoveAll in DownloadActionOutputs, so a server-supplied output path can't resolve outside the intended root and delete something it shouldn't. It's a small, contained change — happy to move the check somewhere you'd prefer if this spot isn't ideal.

ywmei-brt1:
Thanks for reporting this issue, we fixed this with  https://github.com/bazelbuild/remote-apis-sdks/pull/665 and https://github.com/bazelbuild/remote-apis-sdks/pull/667

PR #656, opened 2026-05-30
Verify content digest on the batch CAS download path

evilgensec:
## What

`BatchDownloadBlobsWithStats` stores each blob returned by `BatchReadBlobs` keyed by the digest the server asserts in the response, without re-hashing the returned bytes:

```go
res[digest.NewFromProtoUnvalidated(r.Digest)] = bi
```

There is no length check and no `NewFromBlob(r.Data) == requested` check, so a CAS that returns bytes which do not match the requested digest — a buggy or compromised server, a caching/forwarding proxy, or an on-path party on a non-TLS connection — is taken at face value, and the bytes are later written to disk by `downloadBatch` / `downloadNonUnified`.

The streamed path already guards against this: `readBlobStreamed` tees the stream through `digest.NewFromReader` and returns `calculated digest %s != expected digest %s` on mismatch. Because `useBatchOps` defaults to true and `makeBatches` groups every sub-`MaxBatchSize` blob into a batch, the unverified batch path is the common route for most blobs.

## Change

Verify the returned (post-decompression) bytes hash to the requested digest before storing them, mirroring the streamed path, and key the result by the verified digest. Adds `TestBatchDownloadBlobsDigestMismatch`.

### DIFF
--- a/go/pkg/client/batch_retries_test.go
+++ b/go/pkg/client/batch_retries_test.go
@@ -47,9 +47,9 @@ func (f *flakyBatchServer) BatchReadBlobs(ctx context.Context, req *repb.BatchRe
 		f.numErrors++
 		resp := &repb.BatchReadBlobsResponse{
 			Responses: []*repb.BatchReadBlobsResponse_Response{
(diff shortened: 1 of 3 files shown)


evilgensec:
Gentle bump on this one. Has anyone had a chance to look yet? Happy to rebase or adjust if that helps.

PR #657, opened 2026-06-01
Validate download symlink targets stay within the output directory

adilburaksen:
`DownloadOutputs` (`go/pkg/client/cas_download.go`) creates symlinks from output trees using the `SymlinkTarget` value returned by the server, used verbatim. A server could therefore return a symlink whose target is an absolute path or uses `..` to point outside the output directory, creating links to arbitrary locations on the client filesystem.

This rejects symlink targets that resolve outside `outDir` before the link is created, mirroring the containment already applied to output paths. Targets that stay within the output directory are unaffected.

### DIFF
--- a/go/pkg/client/cas_download.go
+++ b/go/pkg/client/cas_download.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


adilburaksen:
Closing in favor of #650, which addresses the same unvalidated download symlink target and additionally covers the `os.RemoveAll` output-directory traversal in `DownloadActionOutputs`. Deferring to that earlier, more complete fix.

PR #662, opened 2026-07-25
Stop embedding raw response body in actas parse-error messages

herdiyana256:
getToken() and getSignedJWT() in go/pkg/actas/actas.go both included the full raw HTTP response body in the error returned when json.Unmarshal failed on an otherwise-successful (HTTP 200) response.

Since these responses carry an access token and a signed JWT respectively, a response that fails to unmarshal while still containing a live credential -- an unexpected field type such as expires_in coming back as a string instead of a number, or a body truncated by a network intermediary -- put that credential verbatim into the returned error, and from there into whatever logs the caller writes errors to.

Both call sites now format only the underlying json.Unmarshal error, dropping the raw response body entirely.

Added two regression tests reproducing both trigger conditions (a malformed-but-200 token response, a truncated signJWT response), confirming the credential is present in the error before the fix and absent after it.

### DIFF
--- a/go/pkg/actas/actas.go
+++ b/go/pkg/actas/actas.go
@@ -180,7 +180,10 @@ func (s *TokenSource) getSignedJWT(headers map[string]string) (*signaturePayload
 
 	payload := &signaturePayload{}
 	if err := json.Unmarshal(signatureBody, payload); err != nil {
-		return nil, fmt.Errorf("failed to parse sign jwt payload %q: %v", string(signatureBody), err)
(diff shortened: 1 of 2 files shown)


herdiyana256:
Bump on this, open since late July with no review yet. It stops getToken and getSignedJWT from putting the raw 200-response body, which can hold a live access_token or signedJwt, into the returned error. Small change plus a regression test. @gkousik could you take a look when you get a chance?

PR #663, opened 2026-07-25
Validate symlink targets stay within outDir in DownloadOutputs

herdiyana256:
DownloadOutputs in go/pkg/client/cas_download.go validates TreeOutput.Path (where a symlink gets created) via getAbsPath, but never validated TreeOutput.SymlinkTarget (what the symlink points to) before calling os.Symlink. That value comes directly from server-supplied proto fields (ActionResult.OutputFileSymlinks/OutputDirectorySymlinks, or a Directory's Symlinks), so a symlink target could resolve to any path outside outDir with no error at all.

This is asymmetric with the upload side: ComputeMerkleTree already enforces target containment for local symlinks via getTargetRelPath/TreeSymlinkOpts, rejecting a target outside the exec root by default. The download side had no equivalent check, not even behind an opt-in flag.

Added checkSymlinkTargetContained, applied the same way getAbsPath is applied to Path: resolves the target (relative to the symlink's own directory, or as-is if absolute) and rejects it if it falls outside outDir.

Added two regression tests: one confirming an escaping target is rejected and no symlink is created, one confirming a normal within-tree relative target still works.

### DIFF
--- a/go/pkg/client/cas_download.go
+++ b/go/pkg/client/cas_download.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


herdiyana256:
Gentle bump, this has been sitting since late July. It adds the missing containment check so a server-supplied symlink target can't escape outDir on download, matching what the upload side (ComputeMerkleTree) already enforces. @tjgq mind a quick review?

